PR #417, opened 2015-04-09
Implement support for localloc.

AndyAyersMS:
Getting the LLVM IR right here is not too hard. Note we're zeroing things via the helper since
using LLVM's memset might lead to calls into the CRT.

Most of the challenge here is in getting this IR properly lowered to machine code. I have some
corequisite changes to LLVM for this.

### DIFF
--- a/lib/Jit/CMakeLists.txt
+++ b/lib/Jit/CMakeLists.txt
@@ -38,13 +38,22 @@ if (WIN32)
     ${LLILCJIT_EXPORTS_DEF_TEMP} ${LLILCJIT_EXPORTS_DEF})
 
   set(SHARED_LIB_SOURCES ${SOURCES} ${LLILCJIT_EXPORTS_DEF})
+
+  # For windows we need to specify the CoreCLR environment
+  set(LLILC_TARGET_TRIPLE "${LLVM_DEFAULT_TARGET_TRIPLE}-coreclr")
 else()
   if (UNIX)
     set(LLILCJIT_LINK_LIBRARIES ${LLILCJIT_LINK_LIBRARIES} coreclrpal)
   endif()
 
   set(SHARED_LIB_SOURCES ${SOURCES})
+
+  # For non-windows we can use the default triple for now
+  set(LLILC_TARGET_TRIPLE "${LLVM_DEFAULT_TARGET_TRIPLE}")
 endif()
+
+message(STATUS "LLILC_TARGET_TRIPLE is ${LLILC_TARGET_TRIPLE}")
+add_definitions(-DLLILC_TARGET_TRIPLE="${LLILC_TARGET_TRIPLE}")
 
 set(LLVM_EXPORTED_SYMBOL_FILE ${LLILCJIT_EXPORTS_DEF})
 
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -368,9 +368,7 @@ public:
                     ReaderAlignType Alignment, bool IsVolatile) override;
 
   IRNode *loadNull() override;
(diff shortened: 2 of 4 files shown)


AndyAyersMS:
@pgavlin Unless I'm confused there's no place in source control where we specify the target triple, other than docs.

pgavlin:
The corresponding change may be in LLVM, but surely the CoreCLR environment isn't the default when building LLVM?

AndyAyersMS:
I don't think the LLVM build is particularly sensitive to the value of the default triple when it's built. So modifying LLVM's build seems like the wrong way to go about this.

LLILC refers to LLVM_DEFAULT_TARGET_TRIPLE when setting up the module, this seems like the thing we should change. Instead of using the default we should have some other  value we use here that is set up by LLILC's build scripts.

pgavlin:
Hence my question--I thought that you had made (or intended to make) such a change, and was confused by its absence :)

AndyAyersMS:
Ok, good.

AndyAyersMS:
Seems to work but could not figure out how to pass a string as a define value through CMake. So I'm currently stringizing via preprocessor magic.... Seem ok to you? I suspect there are macros like these in headers we include too so perhaps I can use something that is already there.

```
#define STRINGIZE(X) #X
#define ID(X) STRINGIZE(X)
  const char * TargetTriple = ID(LLILC_TARGET_TRIPLE);
  Context.CurrentModule->setTargetTriple(TargetTriple);
```

AndyAyersMS:
@pgavlin PTAL at the updated changes. LLVM part is in already.

lib/Jit/CMakeLists.txt
```diff
@@ -38,14 +38,23 @@ if (WIN32)
...
   endif()
 
   set(SHARED_LIB_SOURCES ${SOURCES})
+
+  # For non-windows we can use the default triple for now
+  set(LLILC_TARGET_TRIPLE "${LLVM_DEFAULT_TARGET_TRIPLE}")
 endif()
 
+message(STATUS "LLILC_TARGET_TRIPLE is ${LLILC_TARGET_TRIPLE}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DLLILC_TARGET_TRIPLE=${LLILC_TARGET_TRIPLE}")
```
pgavlin: I think you can escape the quotes in order to get the value passed as a string: `"-DLLILC_TARGET_TRIPLE=\"${LLILC_TARGET_TRIPLE}\""`

pgavlin:
IR looks good, but I'd definitely like to avoid the extra macros if possible.

lib/Jit/CMakeLists.txt
AndyAyersMS: Tried that already and the quotes didn't make it through. I can experiment some more.

pgavlin: It looks like you might need to escape the escapes. Not sure if there is a good cross-platform way to do that. For the Linux build, I expect that `"-DFOO=\\\"${FOO}\\\""` would work, but CMD might need `"-DFOO=^\"${FOO}^\""` or similar.

pgavlin: If worst comes to worst and there's no good way to do this, your current workaround is fine.

AndyAyersMS: CMake's add_definitions seems a bit better about preserving quotes. So will go with that.

pgavlin: Excellent, thanks.

AndyAyersMS:
retest this please

AndyAyersMS:
retest this please

AndyAyersMS:
CI failures look spurious, so am going to merge this.

PR #426, opened 2015-04-13
Update badges in README

adiaaida:
We updated the names of the windows jobs in the lab, so we need to
update the link in the README.


adiaaida:
@russellhadley PTAL

russellhadley:
Green

PR #427, opened 2015-04-14
Improve documentation, as described below.

mewmew:
* Fix broken fragment identifiers.
* Use relative links when possible.
* Remove unused {#mainpage} from index.md
* Use consistent formatting for subheaders; e.g.

Use:

    ## Subheader

Instead of:

    Subheader
    ----------


pgavlin:
LGTM; thanks for cleaning this up.

PR #433, opened 2015-04-15
Fix the Linux build and some style issues.

pgavlin:
- The PAL definitions we need from CoreCLR moved out of libcoreclr and
  into libcoreclrpal. Update CMakeLists accordingly.
- The JIT options sources were inconsistent with the rest of the code
  w.r.t. the use of the `this` pointer and were missing copyright
  headers.


pgavlin:
@adiaaida @russellhadley PTAL

mjsabby:
LGTM

adiaaida:
LGTM

lib/Jit/options.cpp
```diff
@@ -17,19 +33,19 @@ Options::Options(LLILCJitContext *Context) : Context(Context) {}
 
 void Options::initialize() {
   // Set 'IsAltJit' based on environment information.
-  this->setIsAltJit();
+  setIsAltJit();
```
russellhadley: Is it *really* in the style guide to remove the "this"?  Not prepending members with 'this->' is like my a #1 pet peeve.

pgavlin: Yes, it is. We are only using `this` when necessary for disambiguation.

PR #435, opened 2015-04-16
Fix some wide string issues in JIT options.

pgavlin:
On *Nix platforms, a `wchar_t` is actually a UTF-32 codepoint; on Windows,
it is a UTF-16 codepoint. Thus, when the CLR headers say `wchar_t`, what
they actually mean is `char16_t`. Fix the JIT options code to deal in
UTF-16 strings on all platforms.

### DIFF
--- a/lib/Jit/utility.cpp
+++ b/lib/Jit/utility.cpp
@@ -43,8 +43,13 @@ bool MethodSet::contains(const char *Name, const char *ClassName,
   return false;
 }
 
-std::unique_ptr<std::string> Convert::wideToUtf8(const wchar_t *WideStr) {
-  ArrayRef<char> SrcBytes((const char *)WideStr, (2 * wcslen(WideStr)));
+std::unique_ptr<std::string> Convert::utf16ToUtf8(const char16_t *WideStr) {
+  // Get the length of the input
+  size_t SrcLen = 0;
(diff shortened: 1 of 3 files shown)


pgavlin:
@russellhadley PTAL

lib/Jit/utility.cpp
```diff
@@ -43,8 +43,13 @@ bool MethodSet::contains(const char *Name, const char *ClassName,
   return false;
 }
 
-std::unique_ptr<std::string> Convert::wideToUtf8(const wchar_t *WideStr) {
-  ArrayRef<char> SrcBytes((const char *)WideStr, (2 * wcslen(WideStr)));
+std::unique_ptr<std::string> Convert::utf16ToUtf8(const char16_t *WideStr) {
+  // Get the length of the input
+  size_t SrcLen = 0;
+  for (; WideStr[SrcLen] != (char16_t)0; SrcLen++)
+    ;
```
russellhadley: I can see the push back on wcslen, but is there something in the stdlib that does the thing above.  Something in me is rebelling at having this raw loop here.

russellhadley:
Aside from not liking reimplementing wcslen it looks fine.

lib/Jit/utility.cpp
pgavlin: Unfortunately there is not.

PR #437, opened 2015-04-16
Some doxygen updates.

AndyAyersMS:
Documented reader.h's flow graph interface. Trimmed out some of the options that we'll never use, as well as a couple of useless helpers. Fix some typos.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -136,21 +136,8 @@ extern "C"
     jitFilter(PEXCEPTION_POINTERS ExceptionPointersPtr, void *Param);
 extern void _cdecl fatal(int Errnum, ...);
 
-// Global environment config variables (set by GetConfigString).
-// These are defined/set in jit.cpp.
+// Environment config variables
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
-extern uint32_t EnvConfigCseOn;
-#ifndef NDEBUG
-extern uint32_t EnvConfigCseBinarySearch;
-extern uint32_t EnvConfigCseMax;
-extern uint32_t EnvConfigCopyPropMax;
-extern uint32_t EnvConfigDeadCodeMax;
-extern uint32_t EnvConfigCseStats;
-#endif // !NDEBUG
 #if !defined(CC_PEVERIFY)
 extern uint32_t EnvConfigTailCallOpt;
 #if !defined(NODEBUG)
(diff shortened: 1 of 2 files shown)


include/Reader/reader.h
```diff
@@ -136,21 +136,8 @@ extern "C"
...
-extern uint32_t EnvConfigCseOn;
-#ifndef NDEBUG
-extern uint32_t EnvConfigCseBinarySearch;
-extern uint32_t EnvConfigCseMax;
-extern uint32_t EnvConfigCopyPropMax;
-extern uint32_t EnvConfigDeadCodeMax;
-extern uint32_t EnvConfigCseStats;
-#endif // !NDEBUG
 #if !defined(CC_PEVERIFY)
 extern uint32_t EnvConfigTailCallOpt;
```
russellhadley: These options that remain, do we want to move them to the Options object?  (leaving that for a next step is fine - just we should open an issue.)

russellhadley:
Green from me.

include/Reader/reader.h
AndyAyersMS: Yes, but we'll need to refactor the options object so it can be visible in the base reader (which is client agnostic). I'll open an issue.

pgavlin:
LGTM as well. Thanks for fixing those typos I left behind :)

AndyAyersMS:
Odd that there's no CI status. Anyways this one should be safe.

include/Reader/reader.h
```diff
@@ -1033,39 +1006,147 @@ void rgnSetCatchClassToken(EHRegion *CatchRegion, mdToken Token);
...
 
-// Interface to GenIR defined Flow Graph structures.
-// Implementation Supplied by Jit Client
+/// \name Client Flow Graph interface
+///
+///@{
+
+/// \brief Determine the EH region for this flow graph node
+///
+/// One the EH regions have been created, each flow graph node should have an
```
erozenfeld: Typo One --> Once

include/Reader/reader.h
```diff
@@ -1033,39 +1006,147 @@ void rgnSetCatchClassToken(EHRegion *CatchRegion, mdToken Token);
...
+/// innermost containing EH region. This method returns that region.
+///
+/// \param FgNode   The FlowGraphNode of interest.
+/// \returns        Associated EH region.
 EHRegion *fgNodeGetRegion(FlowGraphNode *FgNode);
+
+/// \brief Establish the EH region for this flow graph node
+///
+/// Sets up the association between a FlowGraphNode and and EH Region.
```
erozenfeld: Typo: and and --> and an

PR #439, opened 2015-04-16
Remove global dumpLevel routine.

russellhadley:
The global dumpLevel routine was left in the code by a mistaken merge.  This change just removes the dead code.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -45,24 +45,6 @@
 
 using namespace llvm;
 
-// Get the LLVM IR dump level. For now this is done by directly
-// accessing environment variable. When CLR config support is
-// included, update it here.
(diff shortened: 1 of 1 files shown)


pgavlin:
Should this also be moved into `Options`?

```
bool shouldUseConservativeGC() {
   const char *LevelCStr = getenv("COMPLUS_GCCONSERVATIVE");
   return (LevelCStr != nullptr) && (strcmp(LevelCStr, "1") == 0);
 }
```

russellhadley:
@pgavlin Yep.  I just wanted to do the merge clean up as it's own step.  Moving this (as well as doing the partition for the reader options) is a individual change.  I'll open another issue for this.  (I've been opening issues for each environment variable and capturing what extra work should go with it.  @jimhogg  is working on the AltJit parsing right now for instance.

pgavlin:
SGTM.

PR #442, opened 2015-04-16
Move initial capacity calculation into createStack

JosephTremoulet:
We were using the same calculation at all callsites, so just remove the
parameter and do the calculation in the method itself; this simplifies
adding new callsites.

Also remove the ReaderBase parameter, since it's always the same as the
instance pointer.


AndyAyersMS:
LGTM

PR #443, opened 2015-04-16
Allow explicit offset argument to createPointBlock

JosephTremoulet:
This allows creating point blocks for points other than CurrInstrOffset.


AndyAyersMS:
LGTM. You may want to squash your commits.

PR #444, opened 2015-04-16
Allow splitting point blocks in splitCurrentBlock

JosephTremoulet:
Check for the case that the current block is a point block for point
CurrInstrOffset; in that case, using NextInstrOffset for the split would
extend the block's range, so preserve the smaller end offset.


pgavlin:
LGTM. Thanks for the helpful comments.

PR #445, opened 2015-04-16
Preserve stack-propagating-ness in fgSplitBlock

JosephTremoulet:
If we're splitting a block that we've already marked as not propagating
stack (i.e. a point block), preserve that annotation on the newly-created
tail block (since it represents the same flow point as exit from the
original block).  The bit is left set on the original block (which becomes
the head); it's up to the caller to know why it's splitting the block here
and whether stack is propagated at this split point.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -1888,6 +1888,7 @@ void GenIR::fgNodeSetEndMSILOffset(FlowGraphNode *Fg, uint32_t Offset) {
 FlowGraphNode *GenIR::fgSplitBlock(FlowGraphNode *Block, IRNode *Node) {
   Instruction *Inst = (Instruction *)Node;
   BasicBlock *TheBasicBlock = (BasicBlock *)Block;
+  bool PropagatesStack = fgNodePropagatesOperandStack(Block);
   BasicBlock *NewBlock;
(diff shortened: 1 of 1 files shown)


pgavlin:
LGTM. Are there any other annotations that need to be similarly propagated?

JosephTremoulet:
These are the annotations we currently have:

```c++
  /// Byte Offset in the MSIL instruction stream to the first instruction
  /// in this basic block.
  uint32_t StartMSILOffset;

  /// Byte offset that is just past the last MSIL instruction of the Basic
  /// Block.
  uint32_t EndMSILOffset;

  /// In algorithms traversing the flow graph, used to track which basic blocks
  /// have been visited.
  bool IsVisited;

  /// Used to track what is on the operand stack on entry to the basic block.
  ReaderStack *TheReaderStack;

  /// true iff this basic block uses an operand stack and propagates it to the
  /// block's successors when it's not empty on exit.
  bool PropagatesOperandStack;
```

`splitBlock` is already updating start/end offsets.  `IsVisited` should be false for the new block.  `TheReaderStack` will get updated when we move to the new block.  `PropagatesOperandStack` is what this change covers.  So no, I don't think there are other properties that need updating.

pgavlin:
Great.

PR #446, opened 2015-04-16
Annotate root region as covering entire method

JosephTremoulet:
Without this, we were leaving its start and end offsets both set to zero,
which makes it appear not to contain its children and confuses things.


pgavlin:
LGTM

PR #447, opened 2015-04-16
Remove `beginfinally` intrinsic from EH plans

JosephTremoulet:
The LLVM plan is actually that filter invocations must be included in the
possible side-effects of calls that may throw exceptions.  That plan
should work for LLILC as well.


AndyAyersMS:
LGTM.

PR #448, opened 2015-04-16
Add __pycache__ to .gitignore

JosephTremoulet:
Python creates test/__pycache__ as a side-effect of running
test/llilc_runtests.py.


adiaaida:
LGTM.

PR #449, opened 2015-04-16
Fix a number of warnings from clang.

pgavlin:
- Remove unused __cdecl attributes where necessary
- Change a zero-sized array to a one-sized array; 0-sized arrays are not
  standard C++
- Give names to two anonymous structs in ntimage.h

### DIFF
--- a/include/Pal/LLILCPal.h
+++ b/include/Pal/LLILCPal.h
@@ -157,8 +157,10 @@ extern "C" {
 
 #define UNALIGNED
 
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(BIT64)
 #define __cdecl __attribute__((cdecl))
+#else
+#define __cdecl
 #endif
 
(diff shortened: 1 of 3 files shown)


pgavlin:
@russellhadley @joeduffy PTAL

AndyAyersMS:
Supposedly there are are declarations for PE headers in LLVM somewhere, if you decide you want to get rid of ntimage.h entirely. Not sure how much work this would be.

russellhadley:
Green.  But getting rid of ntimage.h would be awesome.

pgavlin:
I agree. Unfortunately, cor.h makes use of definitions from ntimage.h. Were we to get rd of ntimage.h, we'd have to find a reasonably maintainable way to define these ourselves. Perhaps this would be a mapping from whatever PE information is present in LLVM, but it would still be something that we'd need to maintain.

PR #450, opened 2015-04-16
Allow blocks to have zero stack-propagating preds

JosephTremoulet:
Update `maintainOperandStacks` and new block processing in `msilToIR` to
accomodate this.  Some EH-related blocks will have only exception dispatch
point blocks as predecessors.


erozenfeld:
LGTM.

Issue #32, opened 2015-02-17
GC: Implement safepoint insertion


AndyAyersMS:
Write an LLVM phase to insert safepoints into the IR.

swaroop-sridhar:
Testing with statepoints inserted succeeded: http://dotnet-ci.cloudapp.net/job/dotnet_llilc_prtest/650/

Issue #440, opened 2015-04-16
Move GCCONSERVATIVE environment varaible into Options class


russellhadley:
Rework GCCONSERVATIVE as an Option.  This should result in a bool field on the Options class, as well as a setter routine that does the query of the CoreCLR config value.

russellhadley:
Added during option partitioning work.

PR #451, opened 2015-04-16
Separate options for ConservativeGC and Inserting Safepoints

swaroop-sridhar:
CoreCLR's COMPLUS_GCCONSERVATIVE flag tells the runtime to use
conservative when set. LLILC also used this flag to determine whether
the insertion and lowering of statepoint intrinsics are done.

With this change, steatpoints are inserted if the COMPLUS_INSERTSTATEPOINTS
switch is set, regardless of the COMPLUS_GCCONSERVATIVE flag.
This allows testing of LLILC Jit with statepoints inserted. That is, to
check that the code itself is still correct, regardless of the GC information.

Issue #32

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -71,6 +71,12 @@ bool shouldUseConservativeGC() {
   return (LevelCStr != nullptr) && (strcmp(LevelCStr, "1") == 0);
 }
 
+// Determine if GC statepoints should be inserted.
+bool shouldInsertStatepoints() {
+  const char *LevelCStr = getenv("COMPLUS_INSERTSTATEPOINTS");
+  return (LevelCStr != nullptr) && (strcmp(LevelCStr, "1") == 0);
+}
+
 // The one and only Jit Object.
 LLILCJit *LLILCJit::TheJit = nullptr;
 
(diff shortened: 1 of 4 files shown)


swaroop-sridhar:
@AndyAyersMS  please review

AndyAyersMS:
Can you coordinate this with @russellhadley who is moving the existing conservative GC check to the new options object?

russellhadley:
@swaroop-sridhar there's an issue for this already and I've assigned it to myself.  I'll move it at the same time I refactor the Options class.

lib/Jit/LLILCJit.cpp
```diff
@@ -102,7 +109,9 @@ LLILCJit::LLILCJit() {
   InitializeNativeTargetAsmParser();
   llvm::linkStatepointExampleGC();
 
-  ShouldUseConservativeGC = shouldUseConservativeGC();
+  ShouldInsertStatepoints = shouldInsertStatepoints();
+  assert(ShouldInsertStatepoints || shouldUseConservativeGC() &&
+    "Statepoints not inserted in precise GC mode");
```
pgavlin: I would change the wording here to "Statepoints are required for precise GC" or similar. The current message leaves me with the impression that statepoints should not be inserted for precise GC.

swaroop-sridhar:
@russellhadley , you mean that we should take this change, and you'll move both the flags as part of  #440, right?

russellhadley:
Yes.  But add a note about the other flag to #440 so it's clear (and I don't forget).

AndyAyersMS:
LGTM in case you're waiting for that. Good to see that you and Russ are in sync.

PR #452, opened 2015-04-16
Add llilc_checkpass Python Script.

libengu:
llilc_checkpass compares two test results, no matter summary or verbose,
and checks if any function was successfully jitted by LLILC in base result
but failed jitting in diff result.

If the check passed, the script return 0. In case of 0, it might have
jitted more functions in diff result than in base result. The two results
are not necessary the same in this case. Counting result is reported.

If the check failed due to newly failed function, return 1.

If the check failed due to unexpected reason, return negative numbers.

usage: llilc_checkpass.py [-h] -b BASE_RESULT_PATH -d DIFF_RESULT_PATH

optional arguments:
  -h, --help            show this help message and exit

required arguments:
  -b BASE_RESULT_PATH, --base-result-path BASE_RESULT_PATH
                        full path to base result
  -d DIFF_RESULT_PATH, --diff-result-path DIFF_RESULT_PATH
                        full path to diff result

Also this change includes some supporting changes in llilc_runtest
and applyfilter:
1. In llilc_runtest result normalization, perform a summary extraction on
   verbose result and a renaming on summary result.
2. applyfilter is enhanced with summary extraction so that filtering
   and summary extraction can be done at the same time without the
   need to scan through result file twice.
3. Summary result renaming added in applyfilter
4. Introduce const module in for symbolic constant return code
5. Default to single quote style. Python is a little tricky when double
   quotes and single quotes are mixed.
6. Create required argument section in help for llilc_runtest and llilc_checkpass
7. Add more error and exception handling
8. Remove llvm verification failure from summary


libengu:
@AndyAyersMS @adiaaida PTAL

test/llilc_checkpass.py
```diff
@@ -0,0 +1,101 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-b", "--base-result-path", type=str, required=True,
+                        help="full path to base result")
+    parser.add_argument("-d", "--diff-result-path", type=str, required=True,
+                        help="full path to diff result")
+    args, unknown = parser.parse_known_args(argv)
+
+    if unknown:
+        print("Unknown argument(s): ", ", ".join(unknown))
+        return -2
```
AndyAyersMS: Might want to use symbolic constants for the various return codes, so they're a bit more meaningful.

test/llilc_checkpass.py
```diff
@@ -0,0 +1,101 @@
...
+ 
+    # Collect a list of summary result file relative path in diff result    
+    diff_files = []
+    for root, dirs, files in os.walk(args.diff_result_path):
+        for file in files:
+            if file.endswith('sum.txt'):
+                relative_path = os.path.relpath(root, args.diff_result_path)
+                relative_path = os.path.join(relative_path, file)
+                diff_files.append(relative_path)
+
```
AndyAyersMS: Seems like you might want to warn here if no files were found in base or diff.

AndyAyersMS:
I only sort of understand why renaming is needed here. Seems like we could instead pass the desired name down when running the tests. Maybe the issue is you'd need to pass it down when running checking too? If so maybe we could agree on a sensible default and if a full set of matching files wasn't found, the script could complain with alternate usage.

I'd also prefer if the various operations done to analyze results were non-destructive, so I could have the "raw" verbose files, filtered verbose files, summary files all sitting there side by side.

libengu:
@AndyAyersMS I agree with most of your points. As to result file name,  it may be good for LLILC if we could pass the name down into CoreCLR test script so that CoreCLR test harness creates the file name LLILC wants.  However the summary file is LLILC specific and it is not general to CoreCLR. For CoreCLR it is just error output in general. If there were similar requirements from other JITs, I think it would be reasonable to make the name customizable in CoreCLR test scripts. If so, the change should be in CoreCLR test harness not in LLILC. At this point, we probably will stay with current solution for now.

llilc_checkpass only cares about summary file. In current design, any result will always have summary files. It may or may not have verbose files. Sure, it is always safer to check whether summary file exist or not.

About non-destructive, first of all, the script is actually not desctrutive. The very original "raw" result still reside in CoreCLR bin directory. Second, if we had "raw" verbose files, filter verbose files, summary files sitting there side by side, it would be convenient from one perspective. But it will defeat the goal of normalization. With normalization, we want to convert the result directory into a form that any reasonable diff tool could do an easy side-by-side directory comparison. With "raw" verbose result staying in the directory, we would expect a lot of difference basically every time. The real LLVM IR diffs that developer wants to check would be buried in those noise difference. Plus I only see very little value in having "raw" verbose files staying there. How often do we need to see the real value of normalized address or type ID? I think it is very rare. If we were really in this rare case, developer could go to the "raw" result directory in CoreCLR bin directly. Probably we should put a comment in the script about this.

test/applyfilter.py
```diff
@@ -52,31 +55,64 @@
...
+        with open(src, 'r') as ins, open(dest, 'w') as outs:
+            for line in ins:
+                line = re_addr.sub(r'i64 NORMALIZED_ADDRESS', line)
+                line = re_type.sub(r'%\1\2.NORMALIZED_TYPEID\1 addrspace', line)
+                line = re_phi.sub(r'[ NORMALIZED_ADDRESS, %', line)
+                outs.write(line)
+    else:
+        re_read_failure = re.compile(r'Failed to read ')
+        re_read_success = re.compile(r'Successfully read ')
+        re_vfcn_failure = re.compile(r' but failed verification')
```
erozenfeld: This is no longer needed. Verification failures are now asserts (or normal failures in release builds) so they don't need special treatment here.

russellhadley:
@AndyAyersMS, @libengu, on the passing the name down and the output directories, it was exactly this case kind of issue that lead me to think that we should completely replace the logic of 'runtest.cmd' with python and extend it to allow testing for both LLILC and RyuJITs as well as enabling basic CoreCLR testing cross platform.

AndyAyersMS:
@russellhadley we're headed that way, but the normalization/vetting process is inherently stateful so we need to have a clear picture of where we're headed.

I looked through the logic again and it's actually not that bad. I just had to draw a picture. When you run the tests (either in summary or verbose mode) you get the raw output. In verbose mode this is normalized and summarized; in summary mode the raw output is the summary. The logic there makes sure that either way the summary files have the same names.

So the one step we could save is that if we know we're in summary mode we could pass the desired name down, but we'd still have the discovery/copy step to move it over into the results tree, so it's not that big a savings.

What I would find helpful is if the script emits a high-level description of what it is doing to my local file system. Eg something like this during runtest verbose:
```
Verbose-mode post-processing
found 201 raw test outputs (error.txt)  under bin-dir
creating normalized outputs (error.txt) under results-dir
creating summary outputs (sum.txt) under results-dir
```
and likewise for summary mode
```
Summary-mode post-processing
processing 201 raw test outputs (error.txt) found under bin-dir
creating summary outputs (sum.txt) under results-dir
```
and for (none mode)
```
no post-processing needed
```

@libengu please address this and the other feedback but the file handling part seems ok to me.

test/llilc_checkpass.py
```diff
@@ -0,0 +1,101 @@
...
+# 
+# If the check failed due to newly faily routine, return 1.
+#     
+# If the check failed due to unexpected reason, return negative numbers.
+#
+# usage: llilc_checkpass.py [-h] -b BASE_RESULT_PATH -d DIFF_RESULT_PATH
+#
+# optional arguments:
+#   -h, --help            show this help message and exit
+#   -b BASE_RESULT_PATH, --base-result-path BASE_RESULT_PATH
```
adiaaida: You've listed this under "optional arguments" here, but below it is marked as required. Same with diff.

libengu:
@russellhadley We will get there. Rewriting coreclr runtest in Python will be following. For LLILC, we need to have llilc_runtest and llilc_checkpass in first with current coreclr runtest facilty to get things going. Then we proceed to improving CoreCLR test harness and make corresponding changes in LLILC.

test/llilc_checkpass.py
libengu: @AndyAyersMS I like your suggestion about adding high level description of what is going on. I will add it.

libengu:
@AndyAyersMS I like your suggestion about adding high level description of what is going on. I will add it.

russellhadley:
Thanks.  I understand you're staging.  I just want to make sure you're not doing extra work to handle the partition between wrapper and runtest when they will ultimately be subsumed into a single tool in CoreCLR.

test/llilc_checkpass.py
libengu: @adiaaida This is still an open Python issue. See http://bugs.python.org/issue9694. The issue is that required arguments are listed under optional arguments section. Not sure if it is fixed in Python 3.5. But in 3.4.2 the problem is still there. But we do not assume developer all use latest Python. After some searching, I think the best solution would be using add_argument_group with title "required arguments". Actually the required argument style was from you in other PR. I used to use positional argument in these cases.

The required arguments style was actually from you in other PR. I used to use positional argument for this kind of arguments.

adiaaida: All I was pointing out is that there was a mismatch with the comment. If that's a known python issue, then it's ok.

test/applyfilter.py
libengu: Will remove.

libengu:
4883209 address all issue and pass my verification.

AndyAyersMS:
LGTM. Thanks.

libengu:
a42e9ab added newline to the end of const.py to avoid any potential issue.

libengu:
Just merged this directly as it won't affect CI.

PR #454, opened 2015-04-17
Fix and workarounds for Roslyn.

erozenfeld:
Fix a check in GenIR::simpleFieldAddress that was allowing bad field indices for StructGEP.

Add NYIs for struct type mismatches.

Treat verification errors as compilation failures.


pgavlin:
LGTM.

PR #455, opened 2015-04-17
Fix for using namespace llvm inside header file

kyulee1:
When including the header file for other purpose, this caused ambiguous name conflicts for other types.
By convention, I moved it to cpp file where it is used.


AndyAyersMS:
LGTM

PR #457, opened 2015-04-17
Add makeCall wrapper

JosephTremoulet:
This wraps LLVMBuilder->CreateCall, and will be where the logic goes that
instead creates an invoke with an appropriate exception edge if we're in a
protected region.

I decided to go ahead and split this out from my EH changes and PR it to master since it touches many lines and I expect this aspect of the changes (having a makeCall helper with this signature) to stick.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -24,6 +24,7 @@
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/Verifier.h"
 #include "llvm/IR/CFG.h"
+#include "llvm/IR/CallSite.h"
 #include "reader.h"
 #include "abi.h"
 #include "abisignature.h"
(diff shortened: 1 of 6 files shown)


include/Reader/readerir.h
```diff
@@ -735,12 +736,13 @@ class GenIR : public ReaderBase {
                      bool CanMoveUp = false) override;
 
   // Generate call to helper
-  IRNode *callHelperImpl(CorInfoHelpFunc HelperID, llvm::Type *ReturnType,
-                         IRNode *Arg1 = nullptr, IRNode *Arg2 = nullptr,
-                         IRNode *Arg3 = nullptr, IRNode *Arg4 = nullptr,
-                         ReaderAlignType Alignment = Reader_AlignUnknown,
-                         bool IsVolatile = false, bool NoCtor = false,
-                         bool CanMoveUp = false);
+  llvm::CallSite callHelperImpl(CorInfoHelpFunc HelperID, bool MayThrow,
```
pgavlin: Not something that you have to change now, but it might be nice if this method were changed to take an `ArrayRef<Value*>` instead of the optional `IRNode` parameters.

include/Reader/readerir.h
```diff
@@ -1139,6 +1143,18 @@ class GenIR : public ReaderBase {
...
+  ///
+  /// The call is inserted at the LLVMBuilder's current insertion point.
+  ///
+  /// \param Callee    Target of the call
+  /// \param Args      Arguments to pass to the callee
+  /// \param MayThrow  True if the callee may raise an exception
+  /// \returns         A \p CallSite wrapping the CallInst or InvokeInst
+  llvm::CallSite makeCall(llvm::Value *Callee,
+                          llvm::ArrayRef<llvm::Value *> Args,
+                          bool MayThrow = true);
```
pgavlin: Having a default value for `MayThrow` seems like the wrong thing to me. I'd much rather require that MayThrow be explicitly specified at all callsites for the sake of clarity (and to encourage some thought about whether or not the target might actually throw).

JosephTremoulet: Good catch, thanks.  I'd actually started with defaults and then decided it was a bad idea and switched the helper call stuff to no default, but forgot to go the next level down and get rid of the default on makeCall.  I've updated to remove the default on makeCall and push the parameterization upstream to where we can reasonably make the decision (though in the case of ReaderBase::rdrCall I actually didn't spot any intrinsics that I was sure are safe...).

pgavlin:
LGTM.

PR #458, opened 2015-04-17
Support for synchronized methods.

erozenfeld:
Add calls to MONITOR_ENTER on entry to the method and MONITOR_EXIT on returns.

We can't catch exceptions yet; when we can we will need to add a try/fault for the entire method
so that we can call MONITOR_EXIT on unhandled exceptions.

We have a synchronized method in HelloWorld (System.IO.TextWriter+SyncTextWriter.WriteLine).
I verified that the generated IR looks correct. I also tested with a simple two-thread race that the
behavior of synchronized instance and static methods is as expected.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -436,7 +436,7 @@ public:
   };
 
   void rethrow() override { throw NotYetImplementedException("rethrow"); };
-  void returnOpcode(IRNode *Opr, bool SynchronousMethod) override;
+  void returnOpcode(IRNode *Opr, bool IsSynchronizedMethod) override;
   IRNode *shift(ReaderBaseNS::ShiftOpcode Opcode, IRNode *ShiftAmount,
                 IRNode *ShiftOperand) override;
   IRNode *sizeofOpcode(CORINFO_RESOLVED_TOKEN *ResolvedToken) override;
@@ -754,6 +754,13 @@ public:
   /// PHI of NullCheckArg and the generated call instruction.
   IRNode *callRuntimeHandleHelper(CorInfoHelpFunc Helper, IRNode *Arg1,
                                   IRNode *Arg2, IRNode *NullCheckArg);
+
+  /// Generate a helper call to enter or exit a monitor used by synchronized
+  /// methods.
+  ///
+  /// \param IsEnter true if the monitor should be entered; false if the monitor
+  /// should be exited.
+  void callMonitorHelper(bool IsEnter);
 
   IRNode *convertToBoxHelperArgumentType(IRNode *Opr,
                                          CorInfoType CorType) override;
@@ -1274,6 +1281,13 @@ private:
   bool NeedsSecurityObject;
   llvm::BasicBlock *EntryBlock;
   llvm::Instruction *TempInsertionPoint;
+  IRNode *MethodSyncHandle; ///< If the method is synchronized, this is
+                            ///< the handle used for entering and exiting
+                            ///< the monitor.
+  llvm::Value *SyncFlag;    ///< For synchronized methods this flag
+                            ///< indicates whether the monitor has been
+                            ///< entered. It is set and checked by monitor
+                            ///< helpers.
   uint32_t TargetPointerSizeInBits;
   const uint32_t UnmanagedAddressSpace = 0;
   const uint32_t ManagedAddressSpace = 1;
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -303,9 +303,10 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
       ABIMethodSignature(MethodSignature, *this, *JitContext->TheABIInfo);
   Function = ABIMethodSig.createFunction(*this, *JitContext->CurrentModule);
 
-  EntryBlock = BasicBlock::Create(*JitContext->LLVMContext, "entry", Function);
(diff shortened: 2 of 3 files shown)


lib/Reader/readerir.cpp
```diff
@@ -303,9 +303,10 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
       ABIMethodSignature(MethodSignature, *this, *JitContext->TheABIInfo);
   Function = ABIMethodSig.createFunction(*this, *JitContext->CurrentModule);
 
-  EntryBlock = BasicBlock::Create(*JitContext->LLVMContext, "entry", Function);
+  llvm::LLVMContext *LLVMContext = JitContext->LLVMContext;
```
pgavlin: Nit: this could be `LLVMContext &LLVMContext = *JitContext->LLVMContext`

lib/Reader/readerir.cpp
```diff
@@ -4236,44 +4267,55 @@ bool GenIR::fgOptRecurse(mdToken Token) {
...
 }
 
-void GenIR::returnOpcode(IRNode *Opr, bool IsSynchronousMethod) {
+void GenIR::returnOpcode(IRNode *Opr, bool IsSynchronizedMethod) {
   const ABIArgInfo &ResultInfo = ABIMethodSig.getResultInfo();
   const CallArgType &ResultArgType = MethodSignature.getResultType();
   CorInfoType ResultCorType = ResultArgType.CorType;
+  Value *ResultValue = nullptr;
 
   if (ResultCorType == CORINFO_TYPE_VOID) {
```
pgavlin: Nit: the condition could be lifted and cached in a local (`bool IsVoidReturn = ResultCorType == CORINFO_TYPE_VOID;`)

pgavlin:
LGTM modulo two nits.

lib/Reader/readerir.cpp
```diff
@@ -367,9 +368,26 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
   // Check for special cases where the Jit needs to do extra work.
   const uint32_t MethodFlags = getCurrentMethodAttribs();
 
-  // TODO: support for synchronized methods
+  // Check for syncronized method. If a method is syncronized the JIT is
```
AndyAyersMS: typo (2x): synchronized

include/Reader/readerir.h
```diff
@@ -436,7 +436,7 @@ class GenIR : public ReaderBase {
   };
 
   void rethrow() override { throw NotYetImplementedException("rethrow"); };
-  void returnOpcode(IRNode *Opr, bool SynchronousMethod) override;
+  void returnOpcode(IRNode *Opr, bool SynchronizedMethod) override;
```
AndyAyersMS: Nit: Can we make the 2nd param name here consistent with the base class name (since that one will be doxygen'd someday).

AndyAyersMS:
Ditto, LGTM with just a few other nits.

lib/Reader/readerir.cpp
```diff
@@ -539,6 +557,19 @@ void GenIR::insertIRForSecurityObject() {
...
+void GenIR::callMonitorHelper(bool IsEnter) {
+  CorInfoHelpFunc HelperId;
+  const uint32_t MethodFlags = getCurrentMethodAttribs();
+  if (MethodFlags & CORINFO_FLG_STATIC) {
+    HelperId =
+        IsEnter ? CORINFO_HELP_MON_ENTER_STATIC : CORINFO_HELP_MON_EXIT_STATIC;
+  } else {
+    HelperId = IsEnter ? CORINFO_HELP_MON_ENTER : CORINFO_HELP_MON_EXIT;
+  }
+  callHelperImpl(HelperId, Type::getVoidTy(*JitContext->LLVMContext),
```
JosephTremoulet: This will conflict with my #457.  How about you go in first, and I'll fix up this callsite before submitting #457.  When I do, should MayThrow be true, or are these calls guaranteed not to throw (and the monitor object guaranteed not to be null)?

lib/Reader/readerir.cpp
erozenfeld: Changed.

lib/Reader/readerir.cpp
erozenfeld: Done.

lib/Reader/readerir.cpp
erozenfeld: Fixed.

include/Reader/readerir.h
erozenfeld: Done.

lib/Reader/readerir.cpp
erozenfeld: Sounds good. I'll merge once CI build is done.
These calls are guaranteed not to throw and the monitor object is guaranteed not to be null.

pgavlin: > These calls are guaranteed not to throw and the monitor object is guaranteed not to be null.

Is this true for an arbitrary mscorlib?

russellhadley: What does arbitrary mean in this context?  Is this possibly a non-CoreCLR/modern surface mscorlib?  (we do not support desktop)

erozenfeld: These are calls to jit helpers, they don't call into mscorlib.

pgavlin: Yeah, I just looked. Disregard my comment.

PR #460, opened 2015-04-17
Correct main CMakeLists.txt file for standalone executables.

### LINKED ISSUE #459: Standalone configuration missing settings needed for driver
When configuring standalone LLILC (i.e. not contained in LLVM/tools)
and also adding a driver executable the cmake configuration failed
because LLVM_RUNTIME_OUTPUT_INTDIR and CMAKE_ARCHIVE_OUTPUT_DIRECTORY
were not set. Solution is to set them appropriately.

richardlford:
When configuring standalone LLILC (i.e. when the LLILC repo
is not placed in LLVM/tools), we were not setting
LLVM_RUNTIME_OUTPUT_INTDIR and LLVM_LIBRARY_OUTPUT_INTDIR.
These must be set if an executable is added.
Also added a dummy driver to verify that it is
compiled and executable created for it.

Closes #459.


richardlford:
@pgavlin  PTAL.

pgavlin:
CMake changes LGTM. @russellhadley for the rest.

russellhadley:
Is this just to hold the build scripts together?  Seems odd to just add an empty main() that does nothing.

richardlford:
As @russellhadley and I discussed, I added comments explaining that the llilc_aot.cpp file is a placeholder for what will eventually be the LLILC ahead-of-time compiler driver.

kyulee1:
@richardlford Can you remove this empty main? I don't see much value except causing merge problem with prototyping.

richardlford:
@kyulee1 I removed the placeholder driver and reverted the driver CMakeLists.txt to an empty file.

PR #461, opened 2015-04-18
Implement loadNonPrimitiveObj.

### LINKED ISSUE #189: Reader: implement ldobj msil instruction

erozenfeld:
Since we allow structs in SSA, the code is similar to that for loading
primitive types.

Change getPrimitiveAddress to getTypedAddress so that it can be used for
both primitive and non-primitive types.

Closes #189.


AndyAyersMS:
LGTM

Issue #462, opened 2015-04-19
Investigate LLILC CSE test case failures 


libengu:
25 JIT CSE test cases were added in CoreCLR repo in https://github.com/dotnet/coreclr/pull/738.
12 of them have issues if run with LLILC. Investigate the causes.

AndyAyersMS:
Started looking at some of the failures.

These tests require implementing stack probes in the prolog (see microsoft/llvm#23)

arrayexpr2
fieldexpr2

This test requires catching an overflow exception (which we throw)

fieldExprUnchecked1

This test triggers huge (~25GB) amounts of allocation in the reader. Might be a good one to work on when we are looking at possible memory leaks (see #233).

HugeArray

AndyAyersMS:
I'll take another look at these.

AndyAyersMS:
Current excludes vs 462:
```
       JIT\jit64\opt\cse\fieldExprUnchecked1.cmd
       JIT\jit64\opt\cse\HugeArray.cmd
       JIT\jit64\opt\cse\HugeArray1.cmd
       JIT\jit64\opt\cse\hugeexpr1.cmd
       JIT\jit64\opt\cse\HugeField1.cmd
       JIT\jit64\opt\cse\HugeField2.cmd
       JIT\jit64\opt\cse\hugeSimpleExpr1.cmd
       JIT\jit64\opt\cse\staticFieldExprUnchecked1.cmd
```

AndyAyersMS:
JIT\jit64\opt\cse\fieldExprUnchecked1.cmd -- EH
       JIT\jit64\opt\cse\HugeArray.cmd -- allocates > 60GB, then dies
       JIT\jit64\opt\cse\HugeArray1.cmd -- allocates > 90GB, then dies
       JIT\jit64\opt\cse\hugeexpr1.cmd -- passes (needs: 100sec, 715MB, with RelWithDbg)
       JIT\jit64\opt\cse\HugeField1.cmd -- allocates > 90GB, then dies
       JIT\jit64\opt\cse\HugeField2.cmd -- allocates > 90GB, then dies
       JIT\jit64\opt\cse\hugeSimpleExpr1.cmd -- passes
       JIT\jit64\opt\cse\staticFieldExprUnchecked1.cmd -- does not exist

I will open a new issue for excessive memory consumption / bad TP. Suspect we have some pathological memory leak because the huge allocation happens very quickly.

PR #463, opened 2015-04-19
Exclude JIT CSE test cases that has issue with LLILC

libengu:
25 JIT CSE test cases were added in CoreCLR repo in https://github.com/dotnet/coreclr/pull/738.
12 of them have issues if run with LLILC. Exclude them
in both powershell and python scripts. Issue https://github.com/dotnet/llilc/issues/462
is created for investigation of the causes.


libengu:
@AndyAyersMS PTAL

AndyAyersMS:
LGTM

PR #464, opened 2015-04-20
Implement the sqrt intrinsic

AndyAyersMS:


include/Reader/readerir.h
```diff
@@ -459,9 +459,12 @@ class GenIR : public ReaderBase {
...
   IRNode *stringGetChar(IRNode *Arg1, IRNode *Arg2) override;
 
-  bool sqrt(IRNode *Arg1, IRNode **RetVal) override {
-    throw NotYetImplementedException("sqrt");
-  };
+  /// Optionally generate inline code for the \p sqrt operation
+  ///
+  /// \param Argument      input value for sqrt
+  /// \param Result [out]  resulting sqrt value, if reader decided to expand
+  /// \returns             true if Result represents the sqrt
```
JosephTremoulet: iff?

lib/Reader/readerir.cpp
```diff
@@ -5479,6 +5479,22 @@ bool GenIR::abs(IRNode *Argument, IRNode **Result) {
   return false;
 }
 
+bool GenIR::sqrt(IRNode *Argument, IRNode **Result) {
+  Type *Ty = Argument->getType();
+
+  if (Ty->isFloatingPointTy()) {
+    Type *Types[] = {Ty};
+    Value *FSqrt = Intrinsic::getDeclaration(JitContext->CurrentModule,
+                                             Intrinsic::sqrt, Types);
```
JosephTremoulet: Can we lower this or does it expect a helper we don't have?

JosephTremoulet:
LGTM aside from one nit and one question.

lib/Reader/readerir.cpp
AndyAyersMS: It lowers fine on x64. Not sure if we'll hit issues with other platforms.

include/Reader/readerir.h
AndyAyersMS: Sure. Also this should be in reader.h, not readerir.h, will move.

lib/Reader/readerir.cpp
JosephTremoulet: Ok.  Thanks for verifying.

I assume that if we do hit issues with other platforms we'll want to use this intrinsic and add the right lowering under our environment anyway.

AndyAyersMS:
Latest commit was just a squash so am going to merge.

PR #475, opened 2015-04-21
Handle merging of nativeint and unmanaged pointers as PHI operand types.

### LINKED ISSUE #468: Reader: allow merge of nativeint and unmanaged pointer values
Some IL stubs have nativeint and unmanaged pointer values when merging operand stacks. Handle that case in GenIR::getStackMergeType.

erozenfeld:
We hit these cases in IL stubs in Roslyn bring-up.

Closes #468.


JosephTremoulet:
LGTM

Issue #477, opened 2015-04-21
Split Options class into a base and client class


russellhadley:
Separate the JIT specific options from a base set that can be accessed by the common MSIL reader.

russellhadley:
dup.  Closed

PR #482, opened 2015-04-21
Add Reader required options

russellhadley:
Adds options DoTailCallOpt, TailCallMax as well as others mentioned in the
MSIL reader.  Also includes the UseConservativeGC/DoInsertStatepoint
options used by the GC work.

This change implements a base Options class (#477) to allow access by the MSIL reader while hiding the client implementation.

@AndyAyersMS, @pgavlin take a look.


include/Jit/options.h
```diff
@@ -31,28 +31,56 @@ enum OptLevel {
...
+
 /// \brief The JIT options provided via CoreCLR configurations.
 ///
-/// This class implements the JIT options flags.  The CoreCLR interface
-/// is queried and results are cached in this object.
+/// This class exposes the JIT options flags. This interface is passed
+/// around to Options consumers to give them access to the results of
+/// the ConfigValue queries but no query logic is exposed here.
 ///
 class Options {
```
pgavlin: Nit: s/class/struct/ and remove the `public:` access specifier below.

include/Jit/options.h
```diff
@@ -31,28 +31,56 @@ enum OptLevel {
...
 public:
-  /// Construct an Options object based on the passed JitContext.
-  Options(LLILCJitContext *Context);
+  DumpLevel DumpLevel;      ///< Dump level for this JIT invocation.
+  OptLevel OptLevel;        ///< Optimization level for this JIT invocation.
+  bool UseConservativeGC;   ///< True if the environment is set to use CGC.
+  bool DoInsertStatepoints; ///< True if the environment calls for statepoints.
+  bool DoTailCallOpt;       //< Tail call optimization.
```
pgavlin: Typo: `//< ` should be `///< `

include/Jit/options.h
```diff
@@ -67,15 +95,31 @@ class Options {
...
+  /// If COMPLUS_TAILCALLOPT is set in the environment set
+  /// DoTailCallOpt to be true;
+  void setDoTailCallOpt();
+
 public:
-  LLILCJitContext *Context; ///< Invocation CLR Execution Engine flags.
-  LLVMDumpLevel DumpLevel;  ///< Dump level for this JIT invocation.
-  ::OptLevel OptLevel;      ///< Optimization level for this JIT invocation.
-  bool IsAltJit;            ///< True if compiling as the alternative JIT.
+  bool IsAltJit; ///< True if compiling as the alternative JIT.
```
pgavlin: Nit: s/compiling/running/

lib/Jit/options.cpp
```diff
@@ -48,9 +48,7 @@ void freeStringConfigValue(ICorJitInfo *CorInfo, UTF16CharT *Value) {
   return CorInfo->freeStringConfigValue((wchar_t *)Value);
 }
 
-Options::Options(LLILCJitContext *Context) : Context(Context) {}
-
-void Options::initialize() {
+JitOptions::JitOptions(LLILCJitContext &Context) : Context(Context) {
```
pgavlin: Is the context used after the constructor?

russellhadley: Today no since all the setters are run within the Constructor.  This wasn't true in earlier implementations, and since we can imagine cases where some option could only be set sometime after the constructor is finished I left the reference as a member.  I'd be fine pulling it but it's a coin toss as to whether I wouldn't be adding it back next week.

pgavlin: Personally, I'd remove it unless it's actually necessary.

lib/Reader/readerir.cpp
```diff
@@ -770,6 +770,24 @@ void GenIR::createSafepointPoll() {
...
+bool GenIR::doTailCallOpt() { return JitContext->Options->DoTailCallOpt; }
+
+// Declare static memory for TailCallCount.
+uint32_t GenIR::TailCallCount = 0;
+
+bool GenIR::checkTailCallMax() {
+  // TODO - maybe cache this on GenIR.  Called once per call.
+  uint32_t TailCallMax = JitContext->Options->TailCallMax;
+
+  if (TailCallMax) {
```
pgavlin: Nit: `if (TailCallMax != 0)`

AndyAyersMS:
I was kind of expecting options.h to be logically associated with the reader and so appear over in the reader includes, and some new header jitoptions.h to be the jit derived part over on the jit side. Having them both in one header seems odd.

lib/Jit/options.cpp
AndyAyersMS: Agreed, it shouldn't be there unless it is needed.

include/Reader/readerir.h
```diff
@@ -1317,6 +1325,7 @@ class GenIR : public ReaderBase {
   uint32_t TargetPointerSizeInBits;
   const uint32_t UnmanagedAddressSpace = 0;
   const uint32_t ManagedAddressSpace = 1;
+  static uint32_t TailCallCount;
 };
```
AndyAyersMS: How are you going to reconcile a static field like `TaillCallCount` with threading? I agree we need these limit-counter type things but we should figure out how they are supposed to work.

russellhadley: In this I just modeled the current behavior from Jit64.  My hand wavier was that we'd run in "single threaded mode".  That being said I don't know how to tell CoreCLR to do that.  But I agree we should figure this out.

russellhadley:
@AndyAyersMS I thought about doing the header split too but didn't want to end up messing with cmake.  But I'll take a run at it now.

lib/Jit/options.cpp
russellhadley: Maybe I was a bit unclear.  The calls that do the queries to fill out the options need Context.  And by design they're not particularly tied to the constructor.  Today they're all called in the constructor but that's not required.  I could change it so that all getters need to be passed a Context but that would complicate any refactoring where they were called outside the constructor.  So It's needed.

pgavlin: I think the suggestion is to avoid stashing it away.

lib/Reader/readerir.cpp
russellhadley: Fixed (in next push)

include/Reader/readerir.h
jkotas: CoreCLR always assumes that the JIT is multithreading safe. It would be up to LLILC to add a giant lock around the compile method if this assumption does not hold.

lib/Jit/options.cpp
russellhadley: I've refactored the code to avoid stashing it away.  I'm still pretty convinced that I'm going to have to undo this in my next round of changes but it's undone.  (it's also now static)

include/Reader/readerir.h
russellhadley: I can't understand how JIT64 preserved the thread safety.  They do have the code #ifdefd out for Release but that seems to be all I can see.

AndyAyersMS:
You have one more update coming, right? Let me know when you've pushed it up.

russellhadley:
@AndyAyersMS hopefully the last push.  Take a look.

PR #483, opened 2015-04-21
Fix MD array creation.

### LINKED ISSUE #469: Reader: fix verification error when building Roslyn
We hit two instances of a verification error when building Roslyn:
Operand is null
  %7 = call %"System.Byte[,]" addrspace(1)* (i64, i32, ...)* inttoptr (i64 140707705165804 to %"System.Byte[,]" addrspace(1)* (i64, i32, ...)*)(i64 140706120665714, i32 2, i32 9, i32 11, <null operand!>)
Failed to read Microsoft.CodeAnalysis.CSharp.Syntax.InternalSyntax.Lexer..cctor[verification error]

Operand is null
  %5 = call %"System.Byte[,]" addrspace(1)* (i64, i32, ...)* inttoptr (i64 140707705165804 to %"System.Byte[,]" addrspace(1)* (i64, i32, ...)*)(i64 140706120665714, i32 2, i32 28, i32 28, <null operand!>)
Failed to read ConversionEasyOut..cctor[verification error]

erozenfeld:
We had an off-by-1 error in the allocation of the argument array.
Because of that we had an extra null argument in the helper call which
resulted in llvm verification failure when building Roslyn.

Closes #469.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -3628,8 +3628,11 @@ IRNode *GenIR::genNewMDArrayCall(ReaderCallTargetData *CallTargetData,
 
   // The helper is variadic; we only want the types of the two fixed arguments
   // but need the values of all the arguments.
-  Type *ArgumentTypes[2];
-  SmallVector<Value *, 16> Arguments(ArgCount + 2);
+  const uint32_t FixedArgCount = 2;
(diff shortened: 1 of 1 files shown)


pgavlin:
LGTM. Thanks for fixing this.

PR #485, opened 2015-04-22
Fix llvm assert when calling StructType::setBody from getClassType.

### LINKED ISSUE #467: Reader: fix assert when calling StructType::setBody from getClassType
When trying to build BabyRoslyn we an assert when calling [this](https://github.com/dotnet/llilc/blob/2ad3639ae1dd4e7c70542504f7d255f1ffc0e7c4/lib/Reader/readerir.cpp#L1408-L1410). The struct should be opaque at that point but it's not.

erozenfeld:
It's possible that while adding the fields to a struct we may complete the struct
due to circular type definitions. Detect that case and return the result right away.

Closes #467.


AndyAyersMS:
LGTM

PR #486, opened 2015-04-22
Reposition first-pass point blocks

JosephTremoulet:
Point blocks created in the first pass might not yet have a spot for their
point available to insert them into the block list.  This change:
 1. adds tracking of whether the first pass is completed or not,
 2. initializes the NodeOffsetListArray earlier so it will be available
    when point blocks are craeted in readerPrePass,
 2. during the first pass (and pre-pass), updates point block creation to call
    fgAddNodeMSILOffset to ensure that an appropriate spot will be created
    for it in fgReplaceBranchTargets, and
 3. in replaceFlowGraphNodeUses (which gets called from
    fgReplaceBranchTargets), moves the point node to the appropriate spot.

This gets first-pass created point blocks in correct MSIL offset order
before we process them in the second pass.

Also update endfinally block creation to use createPointBlock, and fix
some stale comments around fgReplaceBranchTargets.

### DIFF
--- a/lib/Reader/reader.cpp
+++ b/lib/Reader/reader.cpp
@@ -1919,7 +1919,7 @@ FlowGraphNode *ReaderBase::buildFlowGraph(FlowGraphNode **FgTail) {
 // fgAddNodeMSILOffset
 //
 //  The FlowGraphNodeOffsetList acts as a work list. Each time a
-//  branch is added to the the IR stream a temporary target nodeis
+//  branch is added to the the IR stream a temporary target node is
 //  added to the FlowGraphNodeOffsetList.  After all the branches have
 //  been added the worklist is traversed and each temporary node is
 //  replaced with a real one.
@@ -1969,7 +1969,7 @@ FlowGraphNodeOffsetList *ReaderBase::fgAddNodeMSILOffset(
   NewElement->setOffset(TargetOffset);
 
   if (*Node == nullptr) {
-    *Node = makeFlowGraphNode(TargetOffset,
+    *Node = makeFlowGraphNode(TargetOffset, nullptr,
                               fgGetRegionFromMSILOffset(TargetOffset));
   }
   NewElement->setNode(*Node);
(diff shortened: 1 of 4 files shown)


JosephTremoulet:
@erozenfeld PTAL.  This moves endfinally blocks to the logical spot (and more importantly will allow entercatch point blocks to be in a legal spot from which to push the caught exception onto the evaluation stack).

lib/Reader/reader.cpp
```diff
@@ -2292,17 +2292,11 @@ ReaderBase::fgReplaceBranchTarget(uint32_t Offset,
...
-    // Note that this routine is used to insert labels *before* EH regions
-    // are applied, and that branch targets can only be valid
-    // MSIL offsets, which each have distinct instructions and hence
-    // have non-zero block sizes.  Another special case is the final
-    // block, which may have a label but contain no MSIL due, for example,
-    // to an Endfinally instruction.
-    if ((Offset >= Start && Offset < End) ||
-        (Offset == Start && Offset == End && !NextBlock)) {
+    // Find the MSIL block corresponding to the branch target.  Note that the
+    // test used here preculdes selecting a point block as the branch target;
```
erozenfeld: Typo: preculdes --> precludes

erozenfeld:
LGTM with one comment typo. Should we now strengthen checking in msilToIR to not allow out-of-order blocks when computing FlowGraphMSILOffsetOrder?

JosephTremoulet:
@erozenfeld yes, strengthening the check makes sense.  I had to make a couple more changes to make it succeed, though:
 1. Create the UnreachableContinuationBlock as a point block (at end-of-function)
 2. Change the signature on `makeFlowGraphNode` to take an insertion point and use that when we create a new block for the code that follows `ret`.  Without this, end-of-function point blocks were ending up immediately after the first `ret` rather than all the way at the end-of-function.

PTAL.  e188ed79f4 has the delta from previous.

erozenfeld:
Looks good. Thanks for making the change.

Issue #466, opened 2015-04-20
Find better way to exclude failing tests


AndyAyersMS:
Our current technique for excluding failing tests is to delete the tests after building them. This is a bit fragile since the tests are also built when building the CoreCLR -- it also makes it painful to investigate the failing tests.

We need to find a better method.

libengu:
CoreCLR PR https://github.com/dotnet/coreclr/pull/802 and LLILC PR https://github.com/dotnet/llilc/pull/505 should handle the issue.

libengu:
Both PR merged. Close the issue.

PR #488, opened 2015-04-22
Exclude Inline optimization test cases that have issue with LLILC

libengu:
39 inlining optimization test cases were added in CoreCLR repo in dotnet/coreclr#776.
3 of them have issue if run with LLILC, Inline_Handler, Inline_Vars, InlineThrow.
Exclude them in both powershell and python scripts. Issue dotnet#487
is created for investigation of the causes.

### DIFF
--- a/test/LLILCEnv.ps1
+++ b/test/LLILCEnv.ps1
@@ -898,6 +898,14 @@ function Global:ExcludeTest([string]$Arch="x64", [string]$Build="Release")
   del Int_Xor_Op*
   del Double_Xor_Op*
   popd
+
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
LGTM. 

As we start to accumulate more and more of this it would be nice to see some progress on #466.

libengu:
retest this please

libengu:
@AndyAyersMS Yes. I include this as part of rewriting CoreCLR runtest to Python. It needs to control xunit for this purpose. Details under looking.

PR #491, opened 2015-04-22
Implement reader support for catch handlers and may-throw calls

JosephTremoulet:
Map each `BasicBlock` to its containing `EHRegion`.
Generate code for catch handlers, and landing pads with appropriate exception dispatch for catch-protected try regions.
Generate `invoke` with appropriate unwind label rather than `call` for calls that may throw.
Other handler types are still discarded; the plan is to push catch regions through the full phase list before moving on to other handler types (to get an earlier handle on unknowns).
When a function is compiled that contains a catch handler, it goes through reading, verification, and dumping (if enabled), but then is discarded (since otherwise the downstream processing would crash on the first method with a catch handler).

This PR is for a merge into the EH branch; doing a PR now to get the work-in-progress reviewed, but keeping it into the branch to avoid regressing our ability to compile methods that contain catch handlers (but don't dynamically catch exceptions at run-time during test execution).

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -104,8 +104,11 @@ public:
   EHRegionList *Children;
   uint32_t StartMsilOffset;
   uint32_t EndMsilOffset;
+  union {
+    llvm::SwitchInst *EndFinallySwitch; // Only used for Finally regions
+    mdToken CatchClassToken;            // Only used for Catch regions
+  };
   ReaderBaseNS::RegionKind Kind;
-  llvm::SwitchInst *EndFinallySwitch;
 };
 
 struct EHRegionList {
@@ -242,8 +245,12 @@ EHRegion *rgnGetCatchTryRegion(EHRegion *CatchRegion) { return nullptr; }
 void rgnSetCatchTryRegion(EHRegion *CatchRegion, EHRegion *TryRegion) {
   return;
 }
-mdToken rgnGetCatchClassToken(EHRegion *CatchRegion) { return 0; }
-void rgnSetCatchClassToken(EHRegion *CatchRegion, mdToken Token) { return; }
+mdToken rgnGetCatchClassToken(EHRegion *CatchRegion) {
+  return CatchRegion->CatchClassToken;
+}
+void rgnSetCatchClassToken(EHRegion *CatchRegion, mdToken Token) {
+  CatchRegion->CatchClassToken = Token;
+}
 
 #pragma endregion
 
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -3374,8 +3374,35 @@ LoadInst *GenIR::makeLoad(Value *Address, bool IsVolatile,
 
 CallSite GenIR::makeCall(Value *Callee, bool MayThrow, ArrayRef<Value *> Args) {
   if (MayThrow) {
-    // TODO: Generate an invoke with an appropriate unwind label.
+    LandingPadInst *LandingPad =
+        (CurrentRegion == nullptr ? nullptr : CurrentRegion->HandlerLandingPad);
+    if (LandingPad == nullptr) {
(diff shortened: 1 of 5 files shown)


lib/Reader/readerir.cpp
```diff
@@ -2073,6 +2097,214 @@ void GenIR::beginFlowGraphNode(FlowGraphNode *Fg, uint32_t CurrOffset,
...
+    // (ProcessCLRException).  For now, just pretend to be Windows MSVCRT C++
+    // since it triggers the most appropriate paths in WinEHPrepare.
+    // Use a dummy function type.
+    FunctionType *PersonalityTy =
+        FunctionType::get(Type::getVoidTy(LLVMContext), false);
+    PersonalityFunction =
+        Function::Create(PersonalityTy, Function::ExternalLinkage,
+                         "__CxxFrameHandler3", JitContext->CurrentModule);
+  }
+  // Landingpad defiens an i8* (in addrspace 0) which is intended to be a
```
AndyAyersMS: typo: defines

lib/Reader/readerir.cpp
```diff
@@ -1879,11 +1902,13 @@ bool GenIR::isUnmanagedPointerType(llvm::Type *Type) {
...
-EHRegion *fgNodeGetRegion(FlowGraphNode *Node) { return nullptr; }
-
-EHRegion *fgNodeGetRegion(llvm::Function *Function) { return nullptr; }
+EHRegion *GenIR::fgNodeGetRegion(FlowGraphNode *Node) {
+  return FlowGraphInfoMap[Node].Region;
+}
 
-void fgNodeSetRegion(FlowGraphNode *Node, EHRegion *Region) { return; }
+void GenIR::fgNodeSetRegion(FlowGraphNode *Node, EHRegion *Region) {
+  FlowGraphInfoMap[Node].Region = Region;
```
AndyAyersMS: If we never change our minds about block->region association then seems like we could assert that fgNodeGetRegion is nullptr here.

Not sure if that always holds though, given we may need to do things like wrap a try/finally around a synchronized method body. But maybe we could have an Update variant for that.

lib/Reader/readerir.cpp
JosephTremoulet: Fixed

lib/Reader/readerir.cpp
JosephTremoulet: I tried your suggestion but it failed [here](https://github.com/JosephTremoulet/llilc/commit/62fd29482a3e9ade54b98b54dcb022a8df95c5d1#diff-f3d514eb35910c5b4b5e3c8b5af72317R2986):
```c++
      if (fgNodeGetStartMSILOffset(Block) != CurrentOffset) {
        // Split the block so we can maintain a many-to-one block-to-region
        // mapping
        Block = fgSplitBlock(Block, CurrentOffset, nullptr);
      }
      // Set the region on the current block
      fgNodeSetRegion(Block, CurrentRegion);
```
since I changed splitBlock to copy the region to the newly-created tail but then here I patch it up because we're splitting for the sake of starting a new region.  I'm inclined to just leave things alone, but I could add an explicit region parameter to splitBlock if you think it's important.

JosephTremoulet: Sure.  I added the assert, but I already had to change [this caller](https://github.com/JosephTremoulet/llilc/commit/033ce255f4bd8dd129441d6514f1cd103b44e8b8#diff-f3d514eb35910c5b4b5e3c8b5af72317R2987) to use the Update variant -- we create these blocks by successively splitting one big block, and splitBlock annotates each newly-created tail with the splittee's region, so when we first get to each region we're actually overriding the previous region on that block.  (you can see just what I changed [here](https://github.com/dotnet/llilc/compare/62fd29482...JosephTremoulet:b86c9df)).

AndyAyersMS:
LGTM

PR #493, opened 2015-04-22
Support indirect unmanaged calls.

pgavlin:
Unmanaged calls require some extra machinery to setup the necessary data
structures and inform the runtime that control is entering unmanaged code.

This code is comprised of two pieces:
- Unmanaged call frame setup, which occurs in the method prolog, and
- GC mode transition, which occurs at an unmanaged call site.

Unmanaged call frame setup is relatively simple and can be accomplished
entirely in LLVM IR.

The GC transition, however, is more complicated due to the constraints
that must be satisfied for the GC to operate correctly when unmanaged code
is running. In particular, managed code must not run if the GC has been
informed that unmanaged code is running. The approach taken by other
managed code generators (e.g. RyuJIT) in order to meet this constraint is
to inform the GC of the transitions to and from unmanaged code immediately
before and after the machine call instruction, respectively. The same approach
is taken here through the use of the @llvm.experimental.gc_transition intrinic,
which informs the backend to surround the call with the appropriate transition
code. Any code that is not part of the transition itself is represented
directly in IR.

### DIFF
--- a/include/Reader/abisignature.h
+++ b/include/Reader/abisignature.h
@@ -57,13 +57,33 @@ class ABICallSignature : public ABISignature {
 private:
   const ReaderCallSignature &Signature; ///< The target function signature.
 
+  /// \brief Emits a call to an unmanaged function.
+  ///
+  /// This method is called by \p emitCall when emitting a call that targets an
+  /// unmanaged function. It is responsible for emitting the IR required to
+  /// perform any necessary bookkeeping for the GC as well as the call itself.
+  /// The arguments must already have been arranged as per the calling
+  /// convention and target ABI.
+  ///
+  /// \param Reader        The \p GenIR instance that will be used to emit IR.
+  /// \param Target        The call target.
+  /// \oaram MayThrow      True iff the callee may raise an exception.
+  /// \param Args          The arguments to the call, arranged as per the
+  ///                      calling convention and target ABI.
+  /// \param Result [out]  The result of the call, if any.
+  ///
+  /// \returns The call site corresponding to the unmanaged call.
+  llvm::CallSite emitUnmanagedCall(GenIR &Reader, llvm::Value *Target,
+                                   bool MayThrow,
+                                   llvm::ArrayRef<llvm::Value *> Args,
+                                   llvm::Value *&Result) const;
+
 public:
   ABICallSignature(const ReaderCallSignature &Signature, GenIR &Reader,
                    const ABIInfo &TheABIInfo);
 
-  /// \brief Emits a call to a function with using the argument and result
-  ///        passing information for the signature provided when this value
-  ///        was created.
+  /// \brief Emits a call to a function using the argument and result passing
+  ///        information for the signature provided when this value was created.
   ///
   /// \param Reader           The \p GenIR instance that will be used to emit
   ///                         IR.
@@ -75,7 +95,7 @@ public:
   /// \param CallNode [out]   The call instruction.
   ///
   /// \returns The result of the call to the target.
-  llvm::Value *emitCall(GenIR &Reader, llvm::Value *Target, bool mayThrow,
+  llvm::Value *emitCall(GenIR &Reader, llvm::Value *Target, bool MayThrow,
                         llvm::ArrayRef<llvm::Value *> Args,
                         llvm::Value *IndirectionCell,
                         llvm::Value **CallNode) const;
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -256,7 +256,8 @@ public:
                  llvm::Type *> *ArrayTypeMap,
         std::map<CORINFO_FIELD_HANDLE, uint32_t> *FieldIndexMap)
       : ReaderBase(JitContext->JitInfo, JitContext->MethodInfo,
-                   JitContext->Flags) {
+                   JitContext->Flags),
+        UnmanagedCallFrame(nullptr), ThreadPointer(nullptr) {
     this->JitContext = JitContext;
     this->ClassTypeMap = ClassTypeMap;
     this->ReverseClassTypeMap = ReverseClassTypeMap;
@@ -1280,6 +1281,10 @@ private:
   /// \brief Insert IR to setup the security object
   void insertIRForSecurityObject();
 
+  /// \brief Insert IR to setup the unmanaged call frame (PInvoke frame) and
+  ///        the thread pointer.
+  void insertIRForUnmanagedCallFrame();
+
   /// \brief Create the @gc.safepoint_poll() method
   /// Creates the @gc.safepoint_poll() method and insertes it into the
   /// current module. This helper is required by the LLVM GC-Statepoint
@@ -1314,6 +1319,12 @@ private:
   std::map<CORINFO_FIELD_HANDLE, uint32_t> *FieldIndexMap;
   std::map<llvm::BasicBlock *, FlowGraphNodeInfo> FlowGraphInfoMap;
   std::vector<llvm::Value *> LocalVars;
+  llvm::Value *UnmanagedCallFrame; ///< If the method contains unmanaged calls,
+                                   ///< this is the address of the unmanaged
+                                   ///< call frame.
+  llvm::Value *ThreadPointer;      ///< If the method contains unmanaged calls,
+                                   ///< this is the address of the pointer to
+                                   ///< the runtime thread.
   std::vector<CorInfoType> LocalVarCorTypes;
   std::vector<llvm::Value *> Arguments;
   llvm::Value *IndirectResult;
--- a/lib/Reader/abisignature.cpp
+++ b/lib/Reader/abisignature.cpp
@@ -19,6 +19,7 @@
 #include "llvm/IR/CallingConv.h"
 #include "llvm/IR/DataLayout.h"
 #include "llvm/IR/DerivedTypes.h"
+#include "llvm/IR/Intrinsics.h"
 #include "llvm/IR/Module.h"
 #include "reader.h"
 #include "readerir.h"
@@ -42,6 +43,21 @@ static CallingConv::ID getLLVMCallingConv(CorInfoCallConv CC) {
   }
 }
 
+static CorInfoCallConv
+getNormalizedCallingConvention(const ReaderCallSignature &Signature) {
+  // NOTE: this is only correct for X86-64
+
+  CorInfoCallConv CC = Signature.getCallingConvention();
+  switch (CC) {
+  case CORINFO_CALLCONV_STDCALL:
+  case CORINFO_CALLCONV_THISCALL:
+  case CORINFO_CALLCONV_FASTCALL:
+    return CORINFO_CALLCONV_C;
+  default:
+    return CC;
+  }
+}
+
 ABISignature::ABISignature(const ReaderCallSignature &Signature, GenIR &Reader,
                            const ABIInfo &TheABIInfo) {
   const CallArgType &ResultType = Signature.getResultType();
@@ -56,7 +72,8 @@ ABISignature::ABISignature(const ReaderCallSignature &Signature, GenIR &Reader,
     LLVMArgTypes[I++] = Reader.getType(Arg.CorType, Arg.Class);
   }
 
-  CallingConv::ID CC = getLLVMCallingConv(Signature.getCallingConvention());
+  CallingConv::ID CC =
+      getLLVMCallingConv(getNormalizedCallingConvention(Signature));
   TheABIInfo.computeSignatureInfo(CC, LLVMResultType, LLVMArgTypes, Result,
                                   Args);
 
@@ -68,6 +85,8 @@ ABISignature::ABISignature(const ReaderCallSignature &Signature, GenIR &Reader,
 }
 
 Value *ABISignature::coerce(GenIR &Reader, Type *TheType, Value *TheValue) {
+  assert(!TheType->isVoidTy());
+
   Type *ValueType = TheValue->getType();
 
   if (TheType == ValueType) {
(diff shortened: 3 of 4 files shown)


pgavlin:
Note: this code requires [PR 25](https://github.com/Microsoft/llvm/pull/25) in LLVM before it will work.

pgavlin:
@AndyAyersMS @swaroop-sridhar @erozenfeld PTAL

lib/Reader/readerir.cpp
```diff
@@ -575,6 +576,92 @@ void GenIR::callMonitorHelper(bool IsEnter) {
...
+      JitContext->EEInfo.inlinedCallFrameInfo;
+  LLVMContext &LLVMContext = *JitContext->LLVMContext;
+  Type *Int8Ty = Type::getInt8Ty(LLVMContext);
+  Type *Int32Ty = Type::getInt32Ty(LLVMContext);
+  Type *Int8PtrTy = getUnmanagedPointerType(Int8Ty);
+  IRBuilder<>::InsertPoint SavedInsertPoint = LLVMBuilder->saveIP();
+
+  // Mark this function as requiring a frame pointer and as using GC.
+  Function->addFnAttr("no-frame-pointer-elim-non-leaf");
+  Function->setGC("statepoint-example");
```
AndyAyersMS: Does setting GC like this cause issues when we're using conservative GC?

lib/Reader/readerir.cpp
```diff
@@ -575,6 +576,92 @@ void GenIR::callMonitorHelper(bool IsEnter) {
...
+  const struct CORINFO_EE_INFO::InlinedCallFrameInfo &CallFrameInfo =
+      JitContext->EEInfo.inlinedCallFrameInfo;
+  LLVMContext &LLVMContext = *JitContext->LLVMContext;
+  Type *Int8Ty = Type::getInt8Ty(LLVMContext);
+  Type *Int32Ty = Type::getInt32Ty(LLVMContext);
+  Type *Int8PtrTy = getUnmanagedPointerType(Int8Ty);
+  IRBuilder<>::InsertPoint SavedInsertPoint = LLVMBuilder->saveIP();
+
+  // Mark this function as requiring a frame pointer and as using GC.
+  Function->addFnAttr("no-frame-pointer-elim-non-leaf");
```
AndyAyersMS: In `GenIR::insertIRToKeepGenericContextAlive` I did this by modifying the target machine options.
```
  // This method now requires a frame pointer.
  TargetMachine *TM = JitContext->EE->getTargetMachine();
  TM->Options.NoFramePointerElim = true;
```
We should agree on which is the right mechanism and make it a helper and use it in both places.

lib/Reader/readerir.cpp
swaroop-sridhar: This means that statepoints are inserted into this function -- it'll work OK with Conservative GC.
This extra step will not be necessary once we enable statepoint insertion for all functions.

lib/Reader/readerir.cpp
```diff
@@ -575,6 +576,92 @@ void GenIR::callMonitorHelper(bool IsEnter) {
...
+  Type *CallFrameTy = ArrayType::get(Int8Ty, CallFrameInfo.size);
+  Instruction *CallFrameAddress =
+      createTemporary(CallFrameTy, "InlinedCallFrame");
+
+  Type *ThreadPointerTy = getUnmanagedPointerType(ArrayType::get(Int8Ty, 0));
+  Instruction *ThreadPointerAddress =
+      createTemporary(ThreadPointerTy, "ThreadPointer");
+
+  // Intialize the call frame
+  LLVMBuilder->SetInsertPoint(ThreadPointerAddress->getNextNode());
```
AndyAyersMS: Is there any ordering constraint between this and other similar things that happen "on entry" to the method?

lib/Reader/readerir.cpp
```diff
@@ -575,6 +576,92 @@ void GenIR::callMonitorHelper(bool IsEnter) {
...
+  MDString *SPName = MDString::get(LLVMContext, "rsp");
+  Metadata *SPNameNodeMDs[]{SPName};
+  MDNode *SPNameNode = MDNode::get(LLVMContext, SPNameNodeMDs);
+  Value *SPNameValue = MetadataAsValue::get(LLVMContext, SPNameNode);
+  Value *SPValue = LLVMBuilder->CreateCall(ReadRegister, SPNameValue);
+  LLVMBuilder->CreateStore(SPValue, CallSiteSPAddress);
+
+  llvm::Function *FrameAddress = Intrinsic::getDeclaration(
+      JitContext->CurrentModule, Intrinsic::frameaddress);
+  Value *CalleeSavedFPIndices[] = {
```
AndyAyersMS: Do you know what the frame address gets used for by the runtime? I wonder if there is some expectation about frame layout we need to honor.

lib/Reader/readerir.cpp
```diff
@@ -575,6 +576,92 @@ void GenIR::callMonitorHelper(bool IsEnter) {
...
+      JitContext->CurrentModule, Intrinsic::read_register, ReadRegisterTypes);
+  Value *CallSiteSPIndices[] = {
+      ConstantInt::get(Int32Ty, 0),
+      ConstantInt::get(Int32Ty, CallFrameInfo.offsetOfCallSiteSP)};
+  Value *CallSiteSPAddress =
+      LLVMBuilder->CreateInBoundsGEP(CallFrameAddress, CallSiteSPIndices);
+  CallSiteSPAddress = LLVMBuilder->CreatePointerCast(
+      CallSiteSPAddress, getUnmanagedPointerType(SPType));
+  MDString *SPName = MDString::get(LLVMContext, "rsp");
+  Metadata *SPNameNodeMDs[]{SPName};
```
AndyAyersMS: All this will be problematic if we ever try and inline a method with pinvokes.

lib/Reader/readerir.cpp
pgavlin: I believe that it is used during stack unwinding for EH.

lib/Reader/readerir.cpp
pgavlin: Yes, it will--I was just talking to Russ about this. We will need to recognize this case and inline appropriately.

lib/Reader/readerir.cpp
pgavlin: Setting the target options offers module-level control; using the attribute offers function-level control. I would lean towards the attribute, assuming that it also does the correct thing in the case that we encounter leaf functions which require a frame pointer.

lib/Reader/readerir.cpp
pgavlin: The only constraint that I am aware of is that the stack and frame pointers must be final modulo stack allocations later on in the method.

lib/Reader/abisignature.cpp
```diff
@@ -150,14 +346,24 @@ Value *ABICallSignature::emitCall(GenIR &Reader, llvm::Value *Target,
   Type *FunctionPtrTy = Reader.getUnmanagedPointerType(FunctionTy);
 
   Target = Builder.CreateIntToPtr(Target, FunctionPtrTy);
-  CallSite Call = Reader.makeCall(Target, MayThrow, Arguments);
+
+  // Unmanaged calls require an intrinsic in order to ensure that any
+  // callee-save registers that might contain GC pointers are spilled.
```
AndyAyersMS: That's not the only reason for an intrinsic here, right? You are handling the GC mode switch and such there now too.

lib/Reader/abisignature.cpp
```diff
@@ -86,43 +105,220 @@ ABICallSignature::ABICallSignature(const ReaderCallSignature &TheSignature,
...
+      JitContext.EEInfo.inlinedCallFrameInfo;
+  LLVMContext &LLVMContext = *JitContext.LLVMContext;
+  Type *Int8Ty = Type::getInt8Ty(LLVMContext);
+  Type *Int32Ty = Type::getInt32Ty(LLVMContext);
+  Type *Int8PtrTy = Reader.getUnmanagedPointerType(Int8Ty);
+  IRBuilder<> &Builder = *Reader.LLVMBuilder;
+
+  if (Reader.UnmanagedCallFrame == nullptr) {
+    Reader.insertIRForUnmanagedCallFrame();
+    assert(Reader.UnmanagedCallFrame != nullptr);
```
AndyAyersMS: I would just as soon see you call `insertIRForUnmanagedCallFrame` uncondtionally and do the "I've already done this" check there instead.

AndyAyersMS:
Overall this looks fine. It's a shame we don't have a good doc-level description of what is going on but maybe we can fix that down the road.

lib/Reader/readerir.cpp
swaroop-sridhar: Here, we'll setup the function to insert statepoints. However, statepoints  will actually not be inserted unless we set the COMPLUS_InsertStatepoints option, because the insertion and rewrite phases are only run under this switch.

swaroop-sridhar: This is the expected behavior, as long as we're running using conservative GC.

swaroop-sridhar: I meant, fur the gc_transition intrinsic, pointer liveness computation and inserting relocations will not be done if the advice phases do not run. So, the intrinsic will simply report no pointers as live.

lib/Reader/abisignature.cpp
pgavlin: Good catch; I'll update the comment.

lib/Reader/abisignature.cpp
pgavlin: That would be better. I'll make the switch.

include/Reader/abisignature.h
```diff
@@ -57,6 +57,11 @@ class ABICallSignature : public ABISignature {
 private:
   const ReaderCallSignature &Signature; ///< The target function signature.
```
erozenfeld: Please add doxygen comments for this method.

include/Reader/readerir.h
```diff
@@ -1310,6 +1315,8 @@ class GenIR : public ReaderBase {
   std::map<CORINFO_FIELD_HANDLE, uint32_t> *FieldIndexMap;
   std::map<llvm::BasicBlock *, FlowGraphNodeInfo> FlowGraphInfoMap;
   std::vector<llvm::Value *> LocalVars;
+  llvm::Value *UnmanagedCallFrame;
```
erozenfeld: Please add doxygen comments for these two fields.

lib/Reader/abisignature.cpp
```diff
@@ -86,43 +105,219 @@ ABICallSignature::ABICallSignature(const ReaderCallSignature &TheSignature,
                                    GenIR &Reader, const ABIInfo &TheABIInfo)
     : ABISignature(TheSignature, Reader, TheABIInfo), Signature(TheSignature) {}
 
-Value *ABICallSignature::emitCall(GenIR &Reader, llvm::Value *Target,
-                                  bool MayThrow, llvm::ArrayRef<Value *> Args,
-                                  llvm::Value *IndirectionCell,
-                                  llvm::Value **CallNode) const {
+static Value *getFieldAddress(IRBuilder<> &Builder, Value *Base,
+                              uint32_t Offset, Type *FieldTy) {
+  // The base value should be an i8*.
```
erozenfeld: or i8[]*

russellhadley:
Good point @AndyAyersMS, @pgavlin can you work up a bit of markdown about calling convention handling and add it to our Documentation?

lib/Reader/abisignature.cpp
```diff
@@ -86,43 +105,219 @@ ABICallSignature::ABICallSignature(const ReaderCallSignature &TheSignature,
...
+  LLVMContext &LLVMContext = *JitContext.LLVMContext;
+  Type *Int8Ty = Type::getInt8Ty(LLVMContext);
+  Type *Int32Ty = Type::getInt32Ty(LLVMContext);
+  Type *Int8PtrTy = Reader.getUnmanagedPointerType(Int8Ty);
+  IRBuilder<> &Builder = *Reader.LLVMBuilder;
+
+  Reader.insertIRForUnmanagedCallFrame();
+  assert(Reader.UnmanagedCallFrame != nullptr);
+  assert(Reader.ThreadPointer != nullptr);
+
```
erozenfeld: Nit: I would move the two lines below to above the asserts  and use CallFrame and Thread in the asserts.

lib/Reader/abisignature.cpp
```diff
@@ -86,43 +105,219 @@ ABICallSignature::ABICallSignature(const ReaderCallSignature &TheSignature,
...
+  Value *CallFrame = Reader.UnmanagedCallFrame;
+  Value *Thread = Reader.ThreadPointer;
+
+  // Set m_pDatum if necessary
+  if (Reader.MethodSignature.hasSecretParameter()) {
+    // TODO: this needs to be updated for direct unmanaged calls.
```
erozenfeld: I'm not sure what this comment refers to. Can you elaborate?

lib/Reader/abisignature.cpp
```diff
@@ -86,43 +105,219 @@ ABICallSignature::ABICallSignature(const ReaderCallSignature &TheSignature,
...
+  // 1) Address of the GC mode field
+  // 2) Address of the thread trap global
+  // 3) Address of CORINFO_HELP_STOP_FOR_GC
+  Module *M = Reader.Function->getParent();
+  Type *CallTypeArgs[] = {Target->getType()};
+  Function *CallIntrinsic = Intrinsic::getDeclaration(
+      M, Intrinsic::experimental_gc_transition, CallTypeArgs);
+
+  const uint32_t TargetArgCount = Arguments.size();
+  SmallVector<Value *, 16> IntrinsicArgs(3 + TargetArgCount + 6);
```
erozenfeld: Please use symbolic constants instead of 3 and 6.

lib/Reader/abisignature.cpp
```diff
@@ -86,43 +105,219 @@ ABICallSignature::ABICallSignature(const ReaderCallSignature &TheSignature,
...
   Value *ResultNode = nullptr;
   SmallVector<Type *, 16> ArgumentTypes(Args.size() + NumExtraArgs);
   SmallVector<Value *, 16> Arguments(Args.size() + NumExtraArgs);
   IRBuilder<> &Builder = *Reader.LLVMBuilder;
 
-  // Check for VSD calls.
+  // Check for calls with special args.
+  //
+  // Any special arguments are passed immediately preceeding the normal
+  // arguments. The backend will place these arguments in the appropriates
```
erozenfeld: Typo: appropriates --> appropriate

lib/Reader/readerir.cpp
```diff
@@ -575,6 +576,98 @@ void GenIR::callMonitorHelper(bool IsEnter) {
                  MethodSyncHandle, (IRNode *)SyncFlag);
 }
 
+void GenIR::insertIRForUnmanagedCallFrame() {
+  // There's nothing more to do if we've already inserted the necessary IR.
+  if (Reader.unmanagedCallFrame != nullptr) {
```
erozenfeld: Should it be Reader.UnmanagedCallFrame (with a capital U)? How does it compile with this typo?

erozenfeld:
LGTM too modulo a few small comments. I agree with @AndyAyersMS . This really needs a doc describing what's going on.

lib/Reader/abisignature.cpp
pgavlin: Direct unmanaged calls require the method handle instead of the stub secret parameter. This will require some additional plumbing (probably nothing more than a new field on ReaderCallSignature).

erozenfeld: Ok, please update the comment with that info.

lib/Reader/readerir.cpp
pgavlin: Created #500 to track.

pgavlin:
All feedback addressed; PTAL.

I have created #499 to track support for direct unmanaged calls and #501 to track documenting our support for unmanaged calls.

include/Reader/abisignature.h
```diff
@@ -57,13 +57,33 @@ class ABICallSignature : public ABISignature {
...
 
+  /// \brief Emits a call to an unmanaged function.
+  ///
+  /// This method is called by \p emitCall when emitting the call target is an
```
erozenfeld: the call target is an unmanaged function --> a call whose target is an unmanaged function

include/Reader/abisignature.h
```diff
@@ -57,13 +57,33 @@ class ABICallSignature : public ABISignature {
...
+  /// unmanaged function. It is responsible for emitting the IR required to
+  /// perform any necessary bookeeping for the GC as well as the call itself.
```
erozenfeld: bookeeping --> bookkeeping

include/Reader/readerir.h
```diff
@@ -1310,6 +1315,12 @@ class GenIR : public ReaderBase {
   std::map<CORINFO_FIELD_HANDLE, uint32_t> *FieldIndexMap;
   std::map<llvm::BasicBlock *, FlowGraphNodeInfo> FlowGraphInfoMap;
   std::vector<llvm::Value *> LocalVars;
+  llvm::Value *UnmanagedCallFrame; ///< If the method contains unmanaged calls,
+                                   ///< this is the address of the unmanaged
+                                   ///< call frame.
+  llvm::Value *ThreadPointer;      ///< If the method contains unamanged calls,
```
erozenfeld: unamanged --> unmanaged

lib/Reader/abisignature.cpp
```diff
@@ -86,43 +105,224 @@ ABICallSignature::ABICallSignature(const ReaderCallSignature &TheSignature,
...
+  // 1) Address of the GC mode field
+  // 2) Address of the thread trap global
+  // 3) Address of CORINFO_HELP_STOP_FOR_GC
+  Module *M = Reader.Function->getParent();
+  Type *CallTypeArgs[] = {Target->getType()};
+  Function *CallIntrinsic = Intrinsic::getDeclaration(
+      M, Intrinsic::experimental_gc_transition, CallTypeArgs);
+
+  const uint32_t PrefixArgCount = 3;
+  const uint32_t PostfixArgCount = 6;
```
erozenfeld: I would also add a symbolic const TransitionArgCount and use it here (+2) and down on line 248.

erozenfeld:
Looks good, just a few more typos and a minor suggestion.

AndyAyersMS:
Your latest version LGTM. Thanks for the updates.

PR #494, opened 2015-04-23
Test: Import GcInfo Encoder Library to LLILCGC branch for lab testing

swaroop-sridhar:
**This change is still a test**, creating a PR into GC branch for lab testing.

This change brings in the GCInfo encoding library from the CoreCLR sources.
It does not import the old dbg-encoder (used for cross verification) and GCDump library.


jkotas:
You should not need all the header files. Have you tried trimming it down to more reasonable set?

jkotas:
If there are some changes required - e.g. adding a new #define for standalone compilation with minimum set of dependencies - it should be fine to send them as PR to CoreCLR repo.

swaroop-sridhar:
I trimmed down the set of required headers, but not the exact definitions within the headers yet.

I've extracted out only the required definitions from the palclr headers in this change. Need to prune more headers, given that the linux build failed.

russellhadley:
@swaroop-sridhar I think @jkotas is suggesting to do the refactor, assuming the changes are tractable, inside CoreCLR so we don't need to make a copy of the lib sources and the built library and headers describing the interface can be dropped into the build output director for use.  Was there something in particular that requires a separate copy? Today we already have a dependency on a built CoreCLR so my expectation is that getting a lib from there shouldn't be too much of an issue.

jkotas:
I think it is a good idea to get the source. Just gcinfoencoder.*, without all the random header files that have nothing to do with it. These random header files should not be required to compile gcinfoencoder.*. They are included accidentally. The fixes required to allow gcinfoencoder.* to be build without them can be done in CoreCLR, so that the two copies can stay exactly the same.

jkotas:
We may also drop gcinfoencoder.* into the package - same scheme as what is used for JIT-EE interface.

jkotas:
The hand-crafter container classes (ArrayList, etc.) can be replace with STL equivalents, under ifdef.

pgavlin:
I think that ideally we would link against the library produced by the CLR build, but enabling that seems like it would require either a) linking the gcinfo library's dependencies into the library itself, or b) changing the interface that gcinfo expects to link against. I think that what Jan is suggesting above--#ifdef'ing the sources--is a reasonable enough approach at the moment, but IMHO it'd be great if we didn't stop there and continued driving towards option (b) in the long term.

jkotas:
I think passing C++ implementation objs or libs between projects is broken. It is prone to mismatches between compiler version, or OS/CRT/STL header definitions.

If you would like to avoid compiling the encoder implementation in llilc, I think that the right approach would be to put the encoder behind interface - it would work exactly like the rest of the JIT-EE interface.

pgavlin:
Part of the work involved would be preparing gcinfo to be used as a static library. In this sense, it would become a project in its own right, rather than an implementation detail of CoreCLR. There are many examples of successful projects that ship as static libraries.

swaroop-sridhar:
Thanks for the feedback @jkotas. So, if I understand correctly, we agree that the next steps are: 
(1) Reduce the dependencies of the GCInfo library in the CoreCLR tree
(2) Get the source code into LLILC tree for now.

In the long term, make GCInfo a part of the JIT-EE interface, right? Thanks.

jkotas:
Right

adiaaida:
@dotnet-bot test this please

PR #497, opened 2015-04-23
Support for interlockedCmpXchg

kyulee1:
I use llvm intrinsic to implement this.
I verified code sequence and ran it with a unit test,
which I will add to our test asset later.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -2891,9 +2891,18 @@ public:
   virtual bool interlockedIntrinsicBinOp(IRNode *Arg1, IRNode *Arg2,
                                          IRNode **RetVal,
                                          CorInfoIntrinsics IntrinsicID) = 0;
-  virtual bool interlockedCmpXchg(IRNode *Arg1, IRNode *Arg2, IRNode *Arg3,
-                                  IRNode **RetVal,
+
+  /// Generate inline code for the \p interlockedCmpXchg operation
+  ///
+  /// \param Destination    A pointer to the destination
+  /// \param Exchange       The exchange value
+  /// \param Comparand      The value to compare to the destination
+  /// \param Result [out]   The result is the initial destination
+  /// \returns              true iff the client expanded the interlockedCmpXchg
+  virtual bool interlockedCmpXchg(IRNode *Destination, IRNode *Exchange,
+                                  IRNode *Comparand, IRNode **Result,
                                   CorInfoIntrinsics IntrinsicID) = 0;
+
   virtual bool memoryBarrier() = 0;
   virtual void switchOpcode(IRNode *Opr) = 0;
   virtual void throwOpcode(IRNode *Arg1) = 0;
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -4457,6 +4457,44 @@ void GenIR::dup(IRNode *Opr, IRNode **Result1, IRNode **Result2) {
   *Result2 = Opr;
 }
 
+bool GenIR::interlockedCmpXchg(IRNode *Destination, IRNode *Exchange,
+                               IRNode *Comparand, IRNode **Result,
+                               CorInfoIntrinsics IntrinsicID) {
+  ASSERT(Exchange->getType() == Comparand->getType());
+  switch (IntrinsicID) {
+  case CORINFO_INTRINSIC_InterlockedCmpXchg32:
+    ASSERT(Exchange->getType() == Type::getInt32Ty(*JitContext->LLVMContext));
+    break;
+  case CORINFO_INTRINSIC_InterlockedCmpXchg64:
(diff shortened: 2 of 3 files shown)


include/Reader/reader.h
```diff
@@ -2891,9 +2891,18 @@ class ReaderBase {
...
-  virtual bool interlockedCmpXchg(IRNode *Arg1, IRNode *Arg2, IRNode *Arg3,
-                                  IRNode **RetVal,
+
+  /// Generate inline code for the \p interlockedCmpXchg operation
+  ///
+  /// \param Destination    A pointer to the destination
+  /// \param Exchange       The exchange value
+  /// \param Comparand      The value to compare to the destination
+  /// \param Result [out]   The result is the initial destination
+  /// \returns              true iff Result represents the interlockedCmpXchg
```
pgavlin: A better way to express this might be "true if the client expanded the interlockedCmpXchg operation"

lib/Reader/readerir.cpp
```diff
@@ -4457,6 +4457,42 @@ void GenIR::dup(IRNode *Opr, IRNode **Result1, IRNode **Result2) {
...
+  ASSERT(Exchange->getType() == Comparand->getType());
+  switch (IntrinsicID) {
+  case CORINFO_INTRINSIC_InterlockedCmpXchg32:
+    ASSERT(Exchange->getType() == Type::getInt32Ty(*JitContext->LLVMContext));
+    break;
+  case CORINFO_INTRINSIC_InterlockedCmpXchg64:
+    ASSERT(Exchange->getType() == Type::getInt64Ty(*JitContext->LLVMContext));
+    break;
+  default:
+    throw NotYetImplementedException("interlockedCmpXchg");
```
pgavlin: Did you use `throw NYI` here instead of `llvm_unreachable` just in case more CMPXCHG intrinsics are added in the future?

include/Reader/reader.h
kyulee1: Okay. Will reword it before merge.

lib/Reader/readerir.cpp
```diff
@@ -4457,6 +4457,42 @@ void GenIR::dup(IRNode *Opr, IRNode **Result1, IRNode **Result2) {
...
+        (IRNode *)LLVMBuilder->CreateIntToPtr(Destination, ComparandPtrTy);
+  } else {
+    ASSERT(DestinationTy->isPointerTy());
+    Destination =
+        (IRNode *)LLVMBuilder->CreatePointerCast(Destination, ComparandPtrTy);
+  }
+
+  Value *Pair = LLVMBuilder->CreateAtomicCmpXchg(
+      Destination, Comparand, Exchange, SequentiallyConsistent,
+      SequentiallyConsistent);
```
pgavlin: Nit: maybe s/SequentiallyConsistent/llvm::SequentiallyConsistent/ for clarity.

lib/Reader/readerir.cpp
kyulee1: Yep..There are more cases, especially for AOT side.

lib/Reader/readerir.cpp
```diff
@@ -4457,6 +4457,42 @@ void GenIR::dup(IRNode *Opr, IRNode **Result1, IRNode **Result2) {
...
+  *Result = (IRNode *)LLVMBuilder->CreateExtractValue(Pair, 0, "loaded");
```
pgavlin: Also, I would probably replace "loaded" with "cmpxchg_result" or similar.

lib/Reader/readerir.cpp
pgavlin: Okay, cool.

lib/Reader/readerir.cpp
```diff
@@ -4457,6 +4457,42 @@ void GenIR::dup(IRNode *Opr, IRNode **Result1, IRNode **Result2) {
...
+  Type *ComparandPtrTy = getUnmanagedPointerType(ComparandTy);
+  Type *DestinationTy = Destination->getType();
+
+  if (DestinationTy->isIntegerTy()) {
+    Destination =
+        (IRNode *)LLVMBuilder->CreateIntToPtr(Destination, ComparandPtrTy);
+  } else {
+    ASSERT(DestinationTy->isPointerTy());
+    Destination =
+        (IRNode *)LLVMBuilder->CreatePointerCast(Destination, ComparandPtrTy);
```
pgavlin: I don't think this is correct--if the destination is a managed pointer (i.e. in `addrspace(1)`), this will end up casting to `addrspace(0)`. All we want to do is cast the destination's element type while leaving the address space alone.

Also, in the case above that where destination type is an integral type, I think that we have to be conservative and treat it as a managed pointer.

kyulee1:
@pgavlin Updated the change based on your feedbacks. PTAL

pgavlin:
LGTM.

PR #502, opened 2015-04-24
Fix unqualified (and hence ambiguous) references.

pgavlin:


pgavlin:
@russellhadley PTAL

russellhadley:
Strange.  Passed the Linux build when I checked in.  Green.

pgavlin:
Out-of-tree builds using a newer clang complain about this.

PR #505, opened 2015-04-24
Update individual tests exclusion mechanism in LLILC side

libengu:
CoreCLR PR https://github.com/dotnet/coreclr/pull/802 implemented an exclusion
mechanism for individual test cases which works generally for different JITs.
This is a corresponding test case exclusion mechanism update in LLILC side
to use that. Individual tests to exclude are specified in ExcluseList ItemGroup
in exclusion.targets file. With this change, we no longer hard delete built
test cases which should reduce a lot of issues and inconvinience for developers.
exclusion file removed. Both Powershell and Python script are updated.

This PR is supposed to be retested after https://github.com/dotnet/coreclr/pull/802 is merged.


libengu:
@AndyAyersMS PTAL

russellhadley:
I think Andy is oof today.  Looks good to me.  Thanks.

libengu:
@russellhadley Thanks.

libengu:
retest this please

PR #506, opened 2015-04-25
Zero initialize locals.

### LINKED ISSUE #495: Reader: zero initialize locals
We need to zero initialize all locals when CORINFO_OPT_INIT_LOCALS is set. We also probably need to zero initialize GC pointer locals and GC pointer fields of local structs even if CORINFO_OPT_INIT_LOCALS is not set.

erozenfeld:
If CORINFO_OPT_INIT_LOCALS is set, zero initialize all locals;
otherwise, zero initialize gc pointers and structs with gc pointers.

This bug was causing bad codegen and runtime failures in 4
methods in baby Roslyn.

Closes #495.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -367,6 +367,8 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
     J++;
   }
 
+  zeroInitLocals();
+
   // Check for special cases where the Jit needs to do extra work.
   const uint32_t MethodFlags = getCurrentMethodAttribs();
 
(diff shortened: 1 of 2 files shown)


lib/Reader/readerir.cpp
```diff
@@ -742,6 +744,51 @@ void GenIR::createSym(uint32_t Num, bool IsAuto, CorInfoType CorType,
...
+}
+
+void GenIR::zeroInitBlock(Value *Address, Value *Size) {
+  // TODO: For small structs we may want to generate an integer StoreInst
+  // instead of calling a helper.
+  LLVMContext &LLVMContext = *JitContext->LLVMContext;
+  const bool MayThrow = false;
+  Type *VoidTy = Type::getVoidTy(LLVMContext);
+  Value *ZeroByte = ConstantInt::get(LLVMContext, APInt(8, 0, true));
+  callHelperImpl(CORINFO_HELP_MEMSET, MayThrow, VoidTy, (IRNode *)Address,
```
pgavlin: Out of curiosity, why do we not use the `@llvm.memset` intrinsic? Is it because in certain cases it lowers to a call to the CRT's `memset`, which is not available? If this is the case, we should file an issue to track getting this to work (similar to what we should eventually be doing for other operations that lower to target library calls).

erozenfeld: Yes. From #417: "Getting the LLVM IR right here is not too hard. Note we're zeroing things via the helper since using LLVM's memset might lead to calls into the CRT." @AndyAyersMS Is there an issue on this already?

AndyAyersMS:
No issue open for memset yet, so I created #508.

erozenfeld:
@AndyAyersMS PTAL

pgavlin:
LGTM.

AndyAyersMS:
LGTM too.

PR #507, opened 2015-04-25
Remove addressOfLeaf.

pgavlin:
All uses of addressOfLeaf can instead be replaced with uses of
addressOfValue.

With these changes, LLILC is able to compile all methods
necessary for the Hello World sample at https://github.com/dotnet/corefxlab/tree/master/demos/CoreClrConsoleApplications/HelloWorld.


pgavlin:
@erozenfeld @AndyAyersMS PTAL

erozenfeld:
LGTM.

PR #509, opened 2015-04-26
Support for storing value type to static field

kyulee1:
This is the case where we don't need writer barrier which was already filtered out.
I use cpBlk which falls back to memcpy helper call to implement this.
I've verified this with a unit test which will commit to our test asset in separate.


russellhadley:
LGTM.

AndyAyersMS:
LGTM too.

PR #510, opened 2015-04-27
Override deregisterEHFrames.

pgavlin:
This is a no-op for now, but is required for correctness given that we
also overload registerEHFrames.

### DIFF
--- a/include/Jit/EEMemoryManager.h
+++ b/include/Jit/EEMemoryManager.h
@@ -108,12 +108,24 @@ public:
   ///
   /// This is currently invoked once per .xdata section. The EE uses this info
   /// to build and register the appropriate .pdata with the OS.
-  /// \param Addr           The address of the data in the pre-loaded image.
-  /// \param LoadAddr       The address the data will have once loaded.
-  /// \param Size           Size of the unwind data in bytes.
+  ///
+  /// \param Addr      The address of the data in the pre-loaded image.
+  /// \param LoadAddr  The address the data will have once loaded.
+  /// \param Size      Size of the unwind data in bytes.
+  ///
   /// \note Because we're not relocating data during loading, \p Addr and
   /// \p LoadAddr are currently identical.
   void registerEHFrames(uint8_t *Addr, uint64_t LoadAddr, size_t Size) override;
+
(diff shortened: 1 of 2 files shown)


pgavlin:
@AndyAyersMS @russellhadley PTAL.

pgavlin:
I should note that this also weakens an overzealous assert in `reserveAllocationSpace`. These changes are necessary to unblock the JIT on Linux.

russellhadley:
@pgavlin I'm curious, how did the problem express it self on Linux?  Code itself looks fine.

AndyAyersMS:
The assert is possibly hinting at a bug in the dyld support. Seems odd for it to anticipate loading a RW section and then actually never get around to loading it. So more details here would be good.

AndyAyersMS:
Also we should really strive to have issues open for any new lines of development, so if you're working through a set of issues to unblock Linux then it would be good to have more detail here too.

pgavlin:
@AndyAyersMS there are a few interesting bits here:
- We didn't hit this before because the zero-sized .data section was not being reported. After commit https://github.com/Microsoft/llvm/commit/78906c655566249da0f14c4e23653ea52bd1f571, all sections are treated as containing at least one byte, whether or not they actually do. The commit message refers to a Radar issue (Radar is Apple's internal bug tracking system), which isn't very helpful.
- We will also hit this (and perhaps the assert in `AllocateDataSection` as well) when running on OS X, because there are RO sections for unwind data in the MachO object file that IIRC aren't tracked as such until linking. I might be off a bit here--it's been a while since I tried running the JIT on OS X--but I did try to fix the issue in LLVM and found that there wasn't a clear path forwards.

I am not sure why there is a zero-sized .data section in the image in the first place. Tracking this down is a bit trickier, and may not be very enlightening (viz. the issue encountered when attempting to run on OS X). I'd be happy to dig in further, however, if you'd like.

AndyAyersMS:
I'm fine with taking this now to unblock things since we still have the assert to catch if we actually try and allocate a RW section. When we get a clearer picture on the LLVM side we can come back and tighten up the checks as appropriate.

PR #515, opened 2015-04-28
Exclude Methodical\MDArray tests. 

erozenfeld:
Exclude Methodical\MDArray tests since llilc fails on them when ngen is disabled.


AndyAyersMS:
LGTM. Might as well merge this straight away.

PR #518, opened 2015-04-29
Fix for GenIR::cmp

### LINKED ISSUE #517: Reader: Zero-extend the result of ICmp and FCmp from i1 to i32 instead of sign-extending it.
We are sign-extending the result of ICmp and FCmp from i1 to i32 instead of zero-extending it. Because of that we generate -1 instead of 1 for true results. That causes bad codegen in a couple of methods in Roslyn.

erozenfeld:
Zero-extend the result of ICmp and FCmp from i1 to i32 instead of sign-extending it.

Now the result of ICmp and FCmp is 0 or 1 instead of 0 or -1.
This was causing bad codegen for a couple of Roslyn methods.

Closes #517.


russellhadley:
LGTM

PR #519, opened 2015-04-29
Disable fast ISel.

pgavlin:
The current implementation of the GC statepoint design is incompatible
with fast ISel, which may visit a call to @llvm.experimental.gc_result
before the corresponding call to @llvm.experimental.gc_statepoint.
Unfortunately, disabling fast ISel turns out not to be as simple as
setting TargetOptions.EnableFastISel to false; instead, the command-line
option must be set to 'false'.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -32,6 +32,7 @@
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Verifier.h"
 #include "llvm/IR/PassManager.h"
+#include "llvm/Support/CommandLine.h"
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Support/Timer.h"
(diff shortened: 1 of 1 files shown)


pgavlin:
@swaroop-sridhar @AndyAyersMS PTAL

swaroop-sridhar:
Looks like this behavior changed recently?

I had watched the Jit take the the normal Isel path after setting TargetOptions.EnableFastISel= false.

swaroop-sridhar:
The change is green for me.

pgavlin:
@swaroop-sridhar I don't think that anything changed in LLVM. The relevant logic is here: https://github.com/Microsoft/llvm/blob/master/lib/CodeGen/LLVMTargetMachine.cpp#L126-L130

What I believe happened is that we started triggering the RHS of the condition.

AndyAyersMS:
It seems odd that the only way to enforce this is via command line options parsing, but that appears to be the case.

LGTM too.

pgavlin:
Build failure is due to #524; merging.

PR #520, opened 2015-04-29
Infer the type of 'this' in indirect calls.

pgavlin:
Indirect call signatures may not specify the type of the 'this' parameter
of the target method. To handle this case, infer the type of the 'this'
parameter from the type of the 'this' argument. This is consistent with
the behavior prescribed by ECMA-335, which states that the type of the
'this' argument should be assumed to be correct in this case (see section
III.3.20).

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -3209,6 +3209,13 @@ public:
 
   virtual IRNode *makeDirectCallTargetNode(void *CodeAddress) = 0;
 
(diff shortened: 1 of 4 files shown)


pgavlin:
@erozenfeld @AndyAyersMS PTAL

erozenfeld:
LGTM

AndyAyersMS:
LGTM too.

pgavlin:
Build failure is due to #524; merging.

PR #521, opened 2015-04-29
Handle classes with no parent class.

pgavlin:
This arises in the case of types that are not present in metadata,
e.g. the type that contains unboxing stubs.


pgavlin:
@erozenfeld @AndyAyersMS PTAL

erozenfeld:
LGTM assuming these classes aren't supposed to have a vtable field.

AndyAyersMS:
It would be nice to verify when there is no parent that the type is one of these special cases.

Do these types have any non-static fields? If not, then describing the layout won't really matter anyways.

pgavlin:
These types do not have any fields, no.

The build failure is due to #524; merging.

PR #536, opened 2015-05-02
Enable tests that were failing because they needed prolog stack probes.

AndyAyersMS:
Now that Microsoft/llvm#35 is in, enable some dependent tests.


pgavlin:
LGTM

PR #537, opened 2015-05-04
Implement loadAndBox

### LINKED ISSUE #300: Reader: implement constrained virtual calls

AndyAyersMS:
This is a simple-minded implementation of `loadAndBox` that loads the value and then passes it off to the box handler.

Closes #300.

For primitives and value types the box handler will immediately turn around and store the value so it can pass the value's address to the box helper, which then loads the value and stores it into the box payload. I've left optimizing this path as a TODO and will open a linked issue.


pgavlin:
LGTM

PR #538, opened 2015-05-04
Update EH document

JosephTremoulet:
- Update section on implicit null checks wrt the null check folding RFC.

- Mark reader catch support in EH branch implemented.  This was added in
  e2bba9a


AndyAyersMS:
LGTM

PR #542, opened 2015-05-05
Implement EndFault stub support

### LINKED ISSUE #393: Investigate cases where endfinally can't find the enclosing EH region
See this [NYI](https://github.com/dotnet/llilc/blob/c0f8f27a03b68c3170df36e34eab0c2399d41425/lib/Reader/readerir.cpp#L1842-L1844).

This came up in Roslyn.  I can track down the method names if we have trouble finding simpler repro cases.

JosephTremoulet:
`endfinally` and `endfault` are the same MSIL instruction (distinguished by
whether the enclosing region's type is `fault` or `finally`).  Accomodate
both by changing the `getInnermostFinallyRegion` helper to also find
`fault`s (and rename it to `getInnermostFaultOrFinallyRegion`), adding
`fgMakeEndFault` parallel to `fgMakeEndFinally`, and invoking one or the
other according to the discovered region type.  Also get rid of the
superfluous `fgMakeEndFinallyHelper` routine.

`fgMakeEndFault` just generates an UnreachableInst for now (since handlers
are stubbed out).

Closes #393

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -1661,12 +1661,12 @@ protected:
   FlowGraphNodeOffsetList *fgAddNodeMSILOffset(FlowGraphNode **Node,
                                                uint32_t TargetOffset);
 
-  /// Get the innermost finally region enclosing the given \p Offset.
+  /// Get the innermost fault or finally region enclosing the given \p Offset.
   ///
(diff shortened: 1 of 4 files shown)


AndyAyersMS:
LGTM.

pgavlin:
LGTM as well.

PR #544, opened 2015-05-05
Support for array and string operations on unexpected types.

### LINKED ISSUE #392: Handle ldlen/strlen cases where the underlying type doesn't look like an array/string
See this [NYI](https://github.com/dotnet/llilc/blob/c0f8f27a03b68c3170df36e34eab0c2399d41425/lib/Reader/readerir.cpp#L2095-L2097) and this [NYI](https://github.com/dotnet/llilc/blob/c0f8f27a03b68c3170df36e34eab0c2399d41425/lib/Reader/readerir.cpp#L2136-L2138).

One case I know of is from `SZArrayHelper::get_Count`, where we see the following IL

```
.method assembly hidebysig instance int32 
        get_Count<T>() cil managed
{
  .custom instance void System.Security.SecuritySafeCriticalAttribute::.ctor() = ( 01 00 00 00 ) 
  // Code size       9 (0x9)
  .maxstack  8
  IL_0000:  ldarg.0
  IL_0001:  call       !!0 System.Runtime.CompilerServices.JitHelpers::UnsafeCast<!!0[]>(object)
  IL_0006:  ldlen
  IL_0007:  conv.i4
  IL_0008:  ret
} // end of method SZArrayHelper::get_Count
```

AndyAyersMS:
This change inserts casts when the incoming (and in one case) the outgoing types are not as expected.

In our current tests these unexpected types are references to System.__Canon. This can represent arrays and strings as well as other reference types.

Closes #392.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -257,7 +257,8 @@ public:
         std::map<CORINFO_FIELD_HANDLE, uint32_t> *FieldIndexMap)
       : ReaderBase(JitContext->JitInfo, JitContext->MethodInfo,
                    JitContext->Flags),
-        UnmanagedCallFrame(nullptr), ThreadPointer(nullptr) {
+        UnmanagedCallFrame(nullptr), ThreadPointer(nullptr),
+        ArrayOfReferenceType(nullptr) {
     this->JitContext = JitContext;
     this->ClassTypeMap = ClassTypeMap;
     this->ReverseClassTypeMap = ReverseClassTypeMap;
@@ -1312,6 +1313,41 @@ private:
   /// \param Size Size of the block.
   void zeroInitBlock(llvm::Value *Address, llvm::Value *Size);
 
+  /// Check if this LLVM type appears to be a CLR array type.
+  ///
+  /// \param Type   Type to examine.
+  /// \returns      True if this type looks like a CLR array type.
+  bool isArrayType(llvm::Type *Type);
+
+  /// Get the LLVM type for the built-in string type.
+  ///
+  /// \returns    LLVM type that models the built-in string type.
+  llvm::Type *getBuiltInStringType();
+
+  /// Get the LLVM type for an array of references.
+  ///
+  /// Used when we know that some type must be an array but our local
+  /// type information thinks otherwise.
+  ///
+  /// \returns    LLVM type that models array of references.
+  llvm::PointerType *getArrayOfReferenceType();
+
+  /// Create the LLVM type for an array of references.
+  void createArrayOfReferenceType();
+
+  /// Create the length, padding, and elements fields for an array type.
+  ///
+  /// \param Fields             Field collection for the array. On input,
+  ///                           should contain only the vtable pointer.
+  /// \param ElementCorType     CorType for the array elements.
+  /// \param ElementClassHandle Class handle for the array elements.
+  /// \returns                  Byte size of the fields added. Fields updated
+  ///                           with length, padding (if needed), and the array
+  ///                           itself.
+  uint32_t addArrayFields(std::vector<llvm::Type *> &Fields,
+                          CorInfoType ElementCorType,
+                          CORINFO_CLASS_HANDLE ElementClassHandle);
+
 private:
   LLILCJitContext *JitContext;
   ABIInfo *TheABIInfo;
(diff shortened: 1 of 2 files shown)


include/Reader/readerir.h
```diff
@@ -1312,6 +1313,28 @@ class GenIR : public ReaderBase {
   /// \param Size Size of the block.
   void zeroInitBlock(llvm::Value *Address, llvm::Value *Size);
 
+  /// Check if this LLVM type appears to be a CLR array type.
+  ///
+  /// \param Type   type to examine.
```
erozenfeld: Capitalize type

lib/Reader/readerir.cpp
```diff
@@ -2020,6 +2020,91 @@ bool GenIR::isUnmanagedPointerType(llvm::Type *Type) {
   return Type->isPointerTy() && !isManagedPointerType(Type);
 }
 
+bool GenIR::isArrayType(llvm::Type *ArrayTy) {
```
pgavlin: Should this also validate the presence of something that looks like a length field?

pgavlin:
LGTM

lib/Reader/readerir.cpp
```diff
@@ -2020,6 +2020,91 @@ bool GenIR::isUnmanagedPointerType(llvm::Type *Type) {
...
+
+PointerType *GenIR::getArrayOfReferenceType() {
+  if (ArrayOfReferenceType == nullptr) {
+    createArrayOfReferenceType();
+  }
+  assert((ArrayOfReferenceType != nullptr) && "should be initialized");
+  return ArrayOfReferenceType;
+}
+
+void GenIR::createArrayOfReferenceType() {
```
erozenfeld: It would be nice to factor at least some of this code with code in getClassType that creates array structs.

erozenfeld:
Looks good with one nit and one suggestion.

lib/Reader/readerir.cpp
AndyAyersMS: I thought about it but it seemed like it would be messy to try and common the code. Do you have some ideas about how this could be done?

lib/Reader/readerir.cpp
AndyAyersMS: I was thinking trailing zero-length array fields were rare enough that they were a pretty good indication that this was an array. If we wanted to do more careful checking, I'd probably track the llvm types that were built for arrays explicitly somehow.

lib/Reader/readerir.cpp
erozenfeld: Something like this:
``` c++
uint32_t addArrayFields(const std::vector<Type *> &Fields,
                        CorInfoType ElementCorType,
                        CORINFO_CLASS_HANDLE ElementHandle) {
  LLVMContext &LLVMContext = *JitContext->LLVMContext;
  const DataLayout *DataLayout = JitContext->EE->getDataLayout();
  // Maybe you can find a better name for this var...
  uint32_t NonPayloadFieldsByteSize = 0;
  // Array length is (u)int32 ....
  Type *ArrayLengthTy = Type::getInt32Ty(LLVMContext);
  Fields.push_back(ArrayLengthTy);
  NonPayloadFieldsByteSize += DataLayout->getTypeSizeInBits(ArrayLengthTy) / 8;

  // For 64 bit targets there's then a 32 bit pad.
  const uint32_t PointerSize = DataLayout->getPointerSizeInBits();
  if (PointerSize == 64) {
    Type *ArrayPadTy = ArrayType::get(Type::getInt8Ty(LLVMContext), 4);
    Fields.push_back(ArrayPadTy);
    NonPayloadFieldsByteSize += DataLayout->getTypeSizeInBits(ArrayPadTy) / 8;
  }

  Type *ElementTy = getType(ElementCorType, ElementHandle);

  Type *ArrayOfElementTy = ArrayType::get(ElementTy, 0);
  Fields.push_back(ArrayOfElementTy);
  return NonPayloadFieldsByteSize;
}
```

AndyAyersMS: Ok, thanks. I'll push an update in a bit.

AndyAyersMS:
Updated wrt review feedback, @erozenfeld PTAL.

erozenfeld:
LGTM, thanks for making the change.

PR #546, opened 2015-05-06
Add GetManagedThreadId to the list of intrinsics handled as normal calls...

AndyAyersMS:
....

As with the other similar cases the jit can simply decline to optimize, and that's what we do here for GetManagedThreadId.


JosephTremoulet:
LGTM

PR #549, opened 2015-05-06
Add Issue ItemMetaData to ItemGroup ExcludeList in exclusion.targets 

libengu:
Add Issue ItemMetaData to ItemGroup ExcludeList in excusion.targets
for each test case so that when a known issue is fixed we know which
tests to re-enable. Some of the issue numbers are relatively general
and more like a placeholder. As investigation goes on and we become
clearer about the causes, more specific issue number will replace the
placeholder.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -1,65 +1,185 @@
 <Project DefaultTargets = "GetListOfTestCmds"
     xmlns="http://schemas.microsoft.com/developer/msbuild/2003" >
     <ItemGroup Condition="'$(XunitTestBinBase)' != ''">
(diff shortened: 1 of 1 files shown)


libengu:
@AndyAyersMS PTAL

AndyAyersMS:
Thanks, LGTM.

libengu:
1aafadf changes Issue from attribute to ItemMetaData. Item has no built-in attribute called Issue.

AndyAyersMS:
Still looks good. 

Are you going to squash your two commits before you merge?

libengu:
6b2cd83 squashed two commits into one.

PR #550, opened 2015-05-06
Stub out getValueFromRuntimeHandle.

pgavlin:
Other JITs either
a) do not optimize this path, or
b) only optimize here if the incoming argument is the result of lowering
   a ldtoken instruction.

We don't yet have the ability do detect (b); stick with (a) in the meantime.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -3882,6 +3882,18 @@ IRNode *GenIR::getTypeFromHandle(IRNode *Arg1) {
   return (IRNode *)LLVMBuilder->CreateLoad(FieldAddress, IsVolatile);
 }
 
+IRNode *GenIR::getValueFromRuntimeHandle(IRNode *Arg1) {
+  // TODO: other JITs either
+  // a) do not optimize this path, or
+  // b) only optimize here if the incoming argument is the result of lowering
+  //    a ldtoken instruction.
+  //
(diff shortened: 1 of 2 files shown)


pgavlin:
This closes #550; #551 tracks further optimizations.

AndyAyersMS:
LGTM.

lib/Reader/readerir.cpp
```diff
@@ -3718,7 +3718,8 @@ IRNode *GenIR::callHelper(CorInfoHelpFunc HelperID, bool MayThrow, IRNode *Dst,
       (Dst == nullptr) ? Type::getVoidTy(LLVMContext) : Dst->getType();
   return (IRNode *)callHelperImpl(HelperID, MayThrow, ReturnType, Arg1, Arg2,
                                   Arg3, Arg4, Alignment, IsVolatile, NoCtor,
-                                  CanMoveUp).getInstruction();
+                                  CanMoveUp)
+      .getInstruction();
```
russellhadley: What's with these weird dangling .<blah>?

pgavlin: When clang-format says jump, I jump.

russellhadley: Shrug.  I figured.  But it looks strange to me.

pgavlin: I agree. We'll see how the formatting job feels about it, since it's likely that I'm running a slightly newer version of clang-format locally.

PR #553, opened 2015-05-07
Implement support for union types.

### LINKED ISSUE #275: Reader: support union types
Union types aren't common but can be created with explicit layout. Right now we bail out on any method that refers to a union.

See this [TODO](https://github.com/dotnet/llilc/blob/73ce7a1cee4a971ea7bb5875c53e7f8c83ae6cb6/lib/Reader/readerir.cpp#L1057-L1061).

AndyAyersMS:
Closes #275.

Types with explicit layout may have fields that overlap one another (aka union types). Before this change we'd fail to compile any method that referenced one of these union types. We now model them as best we can using LLVM types.

LLVM doesn't provide a strong way to describe unions. Instead we generally provide a byte array that covers the extent of the union as a representative placeholder. That means for some fields there is no exact LLMV type counterpart. Downstream consumers cope with this as follows: for any field within the extent of a set of overlapping fields, we omit that field handle from the `FieldIndexMap`, so that when a ldfld or similar goes to find out what LLVM type index to use, it comes up empty-handed. We already had a fall-back path here to simply use the EE provided field offset when this happens, and so now that path kicks in for accesses to overlapping fields, and subsequent pointer casts then fix up the types properly.

However things are not quite that simple. We also want the LLVM type to fully describe the location of any GC references within the type, and it is valid for GC references to appear in one of these overlap sets. GC references must fully overlap one another and can't be safely overlapped with non-GC data (see Ecma-355, II.10.7). This means the extents of the GC references in an union partition the union into ranges of either non-GC data or GC references. We report the former using the byte arrays, and the latter as object references. Thus for instance a type `C` like
```
[StructLayout(LayoutKind.Explicit)]
public struct A
{
    [FieldOffset(0)]public string Name;
    [FieldOffset(8)]public int Size;
}

[StructLayout(LayoutKind.Explicit)]
public struct C
{
    [FieldOffset(0)]public A X;
    [FieldOffset(0)]public string Name;
    [FieldOffset(8)]public int Size;
}

```
would on x64 be described as
```
type <{ %System.Object addrspace(1)*, [8 x i8] }>
```
This special handling for GC references is strictly only necessary for value types (since GC reference locations for value types on the stack must be reported at GC safe points) but for uniformity we do it for all types. We also continue to double-check value type GC locations against the GC pointer info provided by the EE.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -1418,16 +1418,74 @@ Type *GenIR::getClassType(CORINFO_CLASS_HANDLE ClassHandle, bool IsRefClass,
     // default comparator here.
     std::sort(DerivedFields.begin(), DerivedFields.end());
 
+    // If we find overlapping fields, we'll stash them here so we can look
+    // at them collectively.
+    std::vector<std::pair<uint32_t, Type *>> OverlappingFields;
+
     // Now walk the fields in increasing offset order, adding
     // them and padding to the struct as we go.
     for (const auto &FieldPair : DerivedFields) {
       const uint32_t FieldOffset = FieldPair.first;
       CORINFO_FIELD_HANDLE FieldHandle = FieldPair.second;
 
-      // Bail out for now if we see a union type.
+      // Prepare to add this field to the collection.
+      //
+      // If this field is a ref class reference, we don't need the full
+      // details on the referred-to class, and asking for the details here
+      // causes trouble with certain recursive type graphs, for example:
+      //
+      // class A { B b; }
+      // class B : extends A { int c; }
+      //
+      // We need to know the size of A before we can finish B. So we can't
+      // ask for B's details while filling out A.
+      CORINFO_CLASS_HANDLE FieldClassHandle;
+      CorInfoType CorInfoType = getFieldType(FieldHandle, &FieldClassHandle);
+      bool GetFieldDetails = (CorInfoType != CORINFO_TYPE_CLASS);
+      Type *FieldTy = getType(CorInfoType, FieldClassHandle, GetFieldDetails);
+      // If we don't get the details now, make sure to ask
+      // for them later.
+      if (!GetFieldDetails) {
+        DeferredDetailClasses.push_back(FieldClassHandle);
+      }
+
+      // If we see an overlapping field, we need to handle it specially.
       if (FieldOffset < ByteOffset) {
-        ASSERT(IsUnion);
-        throw NotYetImplementedException("union types");
+        assert(IsUnion && "unexpected overlapping fields");
+        // TypedByref and String get special treatement which we will skip
+        // in processing overlaps.
+        assert(!IsTypedByref && "No overlap expected in this type");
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
Hmm, CI failure doesn't give me much to go on. Am going to rerun.

AndyAyersMS:

lib/Reader/readerir.cpp
```diff
@@ -1418,16 +1418,75 @@ Type *GenIR::getClassType(CORINFO_CLASS_HANDLE ClassHandle, bool IsRefClass,
...
+          uint32_t PreviousSize =
+              DataLayout->getTypeSizeInBits(PreviousFieldTy) / 8;
+          uint32_t PreviousOffset = ByteOffset - PreviousSize;
+          addFieldsRecursively(OverlappingFields, PreviousOffset,
+                               PreviousFieldTy);
+        }
+
+        // Add the current field to the overlap set.
+        uint32_t FieldSize = DataLayout->getTypeSizeInBits(FieldTy) / 8;
+        OverlappingFields.push_back(std::make_pair(FieldOffset, FieldTy));
```
JosephTremoulet: Should this be here?  If the field is a scalar, it looks like it's redundant with the subsequent call to `addFieldsRecursively`.  If the field is a struct, it seems odd that we'd want top-level struct fields on the list but not struct-typed sub-fields (which `addFieldsRecursively` will not include in the list)..

lib/Reader/readerir.cpp
```diff
@@ -1634,6 +1675,84 @@ Type *GenIR::getClassType(CORINFO_CLASS_HANDLE ClassHandle, bool IsRefClass,
...
+    Fields.push_back(getBuiltInObjectType());
+    CurrentOffset += getPointerByteSize();
+  }
+
+  // Create a trailing non-gc field if needed.
+  uint32_t OverlapSize = OverlapEndOffset - FirstOffset;
+  uint32_t CurrentOverlapSize = CurrentOffset - FirstOffset;
+  assert((CurrentOverlapSize <= OverlapSize) && "overlap size overflow");
+  if (CurrentOverlapSize < OverlapSize) {
+    uint32_t RemainingSize = OverlapSize - CurrentOverlapSize;
```
JosephTremoulet: Seems like the five lines above could be simplified to:
```c++
  if (CurrentOffset < OverlapEndOffset) {
    uint32_t RemainingSize = OverlapEndOffset - CurrentOffset;
```
no?

(Fine as-is if you find it more readable your way)

JosephTremoulet:
LGTM aside from the two points above.

lib/Reader/readerir.cpp
AndyAyersMS: Yeah, this should not be there, it's left over from an earlier iteration. Thanks for spotting it.

lib/Reader/readerir.cpp
AndyAyersMS: Seems simpler your way... I'll update.

PR #562, opened 2015-05-09
Implement multidimensional array access intrinsics.

### LINKED ISSUE #530: Implement GenIR::arrayGet

### LINKED ISSUE #540: Implement GenIR::arraySet

### LINKED ISSUE #560: Implement GenIR::arrayAddress

erozenfeld:
Implement arrayGet, arraySet, and arrayAddress to expand 2D and 3D
array accesses inline; allow runtime helper to do the expansions
for arrays of other dimensions.

Fix getTypedAddress to deal with ldind.ref with a nativeint arg.
(This was a new test failure after unblocking methods with md array
accesses).

Remove NYI for intrinsics in GenIR::genCall since only a fixed set
of intrinsics is expected to be expanded there; the rest should be
turned into runtime helper calls. Note that ReaderBase::rdrCall also
looks at intrinsics.

Add printMSIL() to dump msil for the current method.

Closes #530.
Closes #540.
Closes #560.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -252,7 +252,7 @@ public:
         std::map<CORINFO_CLASS_HANDLE, llvm::Type *> *ClassTypeMap,
         std::map<llvm::Type *, CORINFO_CLASS_HANDLE> *ReverseClassTypeMap,
         std::map<CORINFO_CLASS_HANDLE, llvm::Type *> *BoxedTypeMap,
-        std::map<std::tuple<CorInfoType, CORINFO_CLASS_HANDLE, uint32_t>,
+        std::map<std::tuple<CorInfoType, CORINFO_CLASS_HANDLE, uint32_t, bool>,
                  llvm::Type *> *ArrayTypeMap,
         std::map<CORINFO_FIELD_HANDLE, uint32_t> *FieldIndexMap)
       : ReaderBase(JitContext->JitInfo, JitContext->MethodInfo,
@@ -377,9 +377,7 @@ public:
 
   IRNode *loadLen(IRNode *Array, bool ArrayMayBeNull = true) override;
 
-  bool arrayAddress(CORINFO_SIG_INFO *Sig, IRNode **RetVal) override {
-    throw NotYetImplementedException("arrayAddress");
-  };
+  bool arrayAddress(CORINFO_SIG_INFO *Sig, IRNode **RetVal);
   IRNode *loadStringLen(IRNode *Arg1) override;
 
   IRNode *getTypeFromHandle(IRNode *HandleNode) override;
@@ -821,12 +819,8 @@ public:
 #endif
 
   // Used to expand multidimensional array access intrinsics
-  bool arrayGet(CORINFO_SIG_INFO *Sig, IRNode **RetVal) override {
-    throw NotYetImplementedException("arrayGet");
-  };
-  bool arraySet(CORINFO_SIG_INFO *Sig) override {
-    throw NotYetImplementedException("arraySet");
-  };
+  bool arrayGet(CORINFO_SIG_INFO *Sig, IRNode **RetVal) override;
+  bool arraySet(CORINFO_SIG_INFO *Sig) override;
 
 #if !defined(NDEBUG)
   void dbDumpFunction(void) override {
@@ -1078,10 +1072,9 @@ private:
 
   /// Generate array bounds check.
   ///
-  /// \param Array Array to be accessed.
+  /// \param ArrayLength Length of the array to be accessed.
   /// \param Index Index to be accessed.
-  /// \returns The input array.
-  IRNode *genBoundsCheck(IRNode *Array, IRNode *Index);
+  void genBoundsCheck(llvm::Value *ArrayLength, llvm::Value *Index);
 
   /// \brief Generate conditional throw for conv.ovf.
   ///
@@ -1228,6 +1221,32 @@ private:
                                   CorInfoType *CorType,
                                   ReaderAlignType *Alignment);
 
+  /// Check whether access to this multidimensional array can be expanded as an
+  /// intrinsic.
+  ///
+  /// \param Sig Intrinsic signature.
+  /// \param IsStore true iff this array access is a store.
+  /// \param IsLoadAddr true iff this array access is a load of an element
+  /// address.
+  /// \param Rank [OUT] Array rank.
+  /// \param ElemCorType [OUT] CorType of the array element.
+  /// \param ElemType [OUT] Type of the array element.
+  /// \returns true iff access to this multidimensional array can be expanded as
+  /// an intrinsic.
+  bool canExpandMDArrayRef(CORINFO_SIG_INFO *Sig, bool IsStore, bool IsLoadAddr,
+                           uint32_t *Rank, CorInfoType *ElemCorType,
+                           llvm::Type **ElemType);
+
+  /// Get address of an element of a multidimensional array.
+  ///
+  /// This method reads array address and indices for each dimensions from the
+  /// operand stack.
+  ///
+  /// \param Rank Array rank.
+  /// \param ElemType Type of the array element.
+  /// \returns Node representing the address of the array element.
+  IRNode *mdArrayRefAddr(uint32_t Rank, llvm::Type *ElemType);
+
   /// Create a PHI node in a block that may or may not have a terminator.
   ///
   /// \param Block Block to create the PHI node in.
@@ -1340,15 +1359,18 @@ private:
 
   /// Create the length, padding, and elements fields for an array type.
   ///
-  /// \param Fields             Field collection for the array. On input,
-  ///                           should contain only the vtable pointer.
-  /// \param ElementCorType     CorType for the array elements.
-  /// \param ElementClassHandle Class handle for the array elements.
-  /// \returns                  Byte size of the fields added. Fields updated
-  ///                           with length, padding (if needed), and the array
-  ///                           itself.
-  uint32_t addArrayFields(std::vector<llvm::Type *> &Fields,
-                          CorInfoType ElementCorType,
+  /// \param Fields                  Field collection for the array. On input,
+  ///                                should contain only the vtable pointer.
+  /// \param IsVector                true iff this is a zero-lower-bound
+  ///                                single-dimensional array.
+  /// \param ArrayRank               Rank of the array.
+  /// \param ElementCorType          CorType for the array elements.
+  /// \param ElementClassHandle      Class handle for the array elements.
+  /// \returns                       Byte size of the fields added. Fields
+  ///                                updated with length, padding (if needed),
+  ///                                and the array itself.
+  uint32_t addArrayFields(std::vector<llvm::Type *> &Fields, bool IsVector,
+                          uint32_t ArrayRank, CorInfoType ElementCorType,
                           CORINFO_CLASS_HANDLE ElementClassHandle);
 
   /// Add fields of a type to the field vector, expanding structures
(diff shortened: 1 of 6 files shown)


lib/Reader/readerir.cpp
```diff
@@ -5530,20 +5772,38 @@ IRNode *GenIR::getTypedAddress(IRNode *Addr, CorInfoType CorInfoType,
...
+      ASSERT(AddressTy == NativeIntTy);
+
+      CORINFO_CLASS_HANDLE ObjectClass =
+          getBuiltinClass(CorInfoClassId::CLASSID_SYSTEM_OBJECT);
+
+      const bool IsRefClass = true;
+      const bool GetRefClassFields = true;
+      Type *ObjectType =
+          getClassType(ObjectClass, IsRefClass, GetRefClassFields);
+      TypedAddr = (IRNode *)LLVMBuilder->CreateIntToPtr(
```
AndyAyersMS: I just added `getBuiltInObjectType` for this.

lib/Reader/readerir.cpp
```diff
@@ -5530,20 +5772,38 @@ IRNode *GenIR::getTypedAddress(IRNode *Addr, CorInfoType CorInfoType,
...
+        throw NotYetImplementedException(
+            "unexpected type in load/store primitive");
+      }
+      PointerType *ReferentPtrTy = cast<PointerType>(ReferentTy);
+      ASSERT(isManagedPointerType(ReferentPtrTy));
+      ASSERT(ReferentTy->getPointerElementType()->isStructTy());
+    } else {
+      // This must be a nativeint, in which case we cast the address
+      // to an address of Object.
+      Type *NativeIntTy =
```
AndyAyersMS: How do we reconcile this part with GC tracking?

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3802,248 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+        return false;
+      }
+    }
+  }
+
+  *ElemCorType = CorType;
+  *ElemType = getType(CorType, Class);
+
+  return true;
+};
```
AndyAyersMS: Nit, `};` -> `}`

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3802,248 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+    if (CorType == CORINFO_TYPE_BYREF) {
+      CORINFO_CLASS_HANDLE ClassTmp = nullptr;
+      GCCorType = getChildType(Class, &ClassTmp);
+      GCClass = ClassTmp;
+    }
+
+    // See if the element is an object.
+    if (GCCorType == CORINFO_TYPE_STRING || GCCorType == CORINFO_TYPE_CLASS) {
+      // If it's not final, we can't expand due to security reasons.
+      uint32_t ClassAttribs = getClassAttribs(GCClass);
```
AndyAyersMS: Presumably this is the store covariance check?

lib/Reader/readerir.cpp
```diff
@@ -2137,6 +2138,21 @@ uint32_t GenIR::addArrayFields(std::vector<llvm::Type *> &Fields,
     FieldByteSize += DataLayout->getTypeSizeInBits(ArrayPadTy) / 8;
   }
 
+  // For multidimensional arrays there are arrays of dimension lengths and
+  // lower bounds. Note that single-dimensional arrays with a non-zero lower
+  // bound have the same layout as single-dimensional arrays with a zero
+  // lower bound.
```
AndyAyersMS: Might be worth describing how the nonzero lower bound is handled for the rank 1 case.

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3802,248 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+  const bool IsLoadAddr = true;
+
+  if (!canExpandMDArrayRef(Sig, IsStore, IsLoadAddr, &Rank, &ElemCorType,
+                           &ElementTy)) {
+    return false;
+  }
+
+  *RetVal = mdArrayRefAddr(Rank, ElementTy);
+  return true;
+};
```
AndyAyersMS: Nit, `};` -> `}`

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3802,248 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+    rdrCallWriteBarrierHelper(ElementAddress, Value, Reader_AlignNatural,
+                              IsVolatile, ResolvedToken, IsNonValueClass,
+                              IsValueIsPointer, IsFieldToken, IsUnchecked);
+  } else {
+    assert(!ElementTy->isStructTy());
+    makeStoreNonNull(Value, ElementAddress, IsVolatile);
+  }
+
+  return true;
+};
```
AndyAyersMS: Nit, `};` -> `}`

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3802,248 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+  // Load the value
+  assert(!ElementTy->isStructTy());
+  const bool IsVolatile = false;
+  LoadInst *LoadInst = makeLoadNonNull(ElementAddress, IsVolatile);
+  uint32_t Align = convertReaderAlignment(Reader_AlignNatural);
+  LoadInst->setAlignment(Align);
+  *RetVal = convertToStackType((IRNode *)LoadInst, ElemCorType);
+
+  return true;
+};
```
AndyAyersMS: Nit, `};` -> `}`

lib/Reader/readerir.cpp
```diff
@@ -5015,14 +5264,9 @@ IRNode *GenIR::genNullCheck(IRNode *Node) {
   return Node;
 }
 
-// Generate array bounds check.
-IRNode *GenIR::genBoundsCheck(IRNode *Array, IRNode *Index) {
+void GenIR::genBoundsCheck(Value *ArrayLength, Value *Index) {
```
AndyAyersMS: The old model of returning a "new" array here was there to try and enforce dependence of subsequent array users on the check. Because we currently create explicit checks and control flow we have control dependence. If we go to an implicit check model we may lose that and have to fall back to data dependence.

AndyAyersMS:
Generally looks good but I have a few questions...

include/Reader/readerir.h
```diff
@@ -1228,6 +1221,32 @@ class GenIR : public ReaderBase {
...
+  /// \returns true iff access to this multidimensional array can be expanded as
+  /// an intrinsic.
+  bool canExpandMDArrayRef(CORINFO_SIG_INFO *Sig, bool IsStore, bool IsLoadAddr,
+                           uint32_t *Rank, CorInfoType *ElemCorType,
+                           llvm::Type **ElemType);
+
+  /// Get address of an element of a multidimensional array.
+  ///
+  /// This method reads array address and indices for each dimensions from the
+  /// operand stack.
```
JosephTremoulet: Probably worth mentioning here that it uses an expansion that assumes the array pointer is non-null (and that callers need to generate a null check before calling this if the array isn't known to be non-null).

lib/Reader/readerir.cpp
```diff
@@ -3746,7 +3763,9 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
                                    Type *ElementTy) {
```
JosephTremoulet: I'm surprised this method (`genArrayElemAddress`) didn't need a call to `mdArrayRefAddr` for the MD case.

lib/Reader/readerir.cpp
erozenfeld: The native int represents an unmanaged pointer to a GC pointer, not a GC pointer itself.That shouldn't be a problem for GC tracking, right? In any case, ldind.ref can legally take a nativeint and has to load a GC pointer so I don't see any alternative here.

lib/Reader/readerir.cpp
erozenfeld: Yes. ArrayTypeMismatchException should be thrown in a case like this:
class B {}
class C : B {}
class D : B {}
C[,] arrayC = new C[2, 2];
B[,] arrayB = arrayC;
arrayB[0, 0] = new D();

If B is final (i.e., sealed) this is not possible.

lib/Reader/readerir.cpp
erozenfeld: genArrayElemAddress is only called for the single-dimensional case.

lib/Reader/readerir.cpp
erozenfeld: genBoundsCheck doesn't do a null check anymore (doesn't call loadLen but gets the length as an ard). But I realized that I was missing a null check for md array accesses. I'm adding it to mdArrayRefAddr.

include/Reader/readerir.h
erozenfeld: This method should check for array pointer for null. I will add the check.

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3802,248 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+  Type *ElementTy = nullptr;
+  const bool IsStore = false;
+  const bool IsLoadAddr = true;
+
+  if (!canExpandMDArrayRef(Sig, IsStore, IsLoadAddr, &Rank, &ElemCorType,
+                           &ElementTy)) {
+    return false;
+  }
+
+  *RetVal = mdArrayRefAddr(Rank, ElementTy);
```
JosephTremoulet: Where's the null check on this path?

lib/Reader/readerir.cpp
erozenfeld: Changed to use that.

lib/Reader/readerir.cpp
erozenfeld: I checked again and it turns out I was wrong. Single-dimensional arrays with a non-zero lower
bound do have arrays (or length 1) for length and lower bound. I updated the code to reflect that.

erozenfeld:
I addressed all feedback. @AndyAyersMS @JosephTremoulet PTAL.

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3797,253 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+  // CORINFO_Array and CORINFO_RefArray.
+  //
+  // Valid indexes are
+  //     dimBounds[i] <= index[i] < dimBounds[i] + dimLengths[i]
+
+  assert(!ElemType->isStructTy());
+
+  if (UseExplicitNullChecks) {
+    Array = genNullCheck(Array);
+  }
```
JosephTremoulet: It would be good to have a comment saying that if UseExplicitNullChecks were false, we'd need to annotate the first load as exception-bearing.

lib/Reader/readerir.cpp
```diff
@@ -3783,6 +3797,253 @@ IRNode *GenIR::genArrayElemAddress(IRNode *Array, IRNode *Index,
...
+  const bool IsLoadAddr = false;
+
+  if (!canExpandMDArrayRef(Sig, IsStore, IsLoadAddr, &Rank, &ElemCorType,
+                           &ElementTy)) {
+    return false;
+  }
+
+  IRNode *Value = ReaderOperandStack->pop();
+
+  IRNode *ElementAddress = mdArrayRefAddr(Rank, ElementTy);
```
JosephTremoulet: Couldn't hurt to comment here that this will null-check the array and therefore we assume non-null at the store below.  Ditto for the mdArrayRefAddr call and load in arrayGet.

JosephTremoulet:
LGTM (two minor commenting nits), thanks.

AndyAyersMS:
LGTM too.

erozenfeld:
Comments added. Will merge once CI job is green.

PR #564, opened 2015-05-10
Remove two vestigial IRNode property setters

JosephTremoulet:
- irNodeSetRegion : we won't be mapping individual instructions to their
   regions, the call to do this for endfinally has already been removed
   (4bd6a1e), and the call to set this on `throw` nodes in the first pass
   is broken since we don't generate the throwing `call`/`invoke` until
   the second pass.
 - irNodeBranchSetMSILOffset : we won't be recording offsets on individual
   instructions, and the call for `throw` nodes is broken as for
   `irNodeSetRegion`


AndyAyersMS:
LGTM.

PR #566, opened 2015-05-12
Verification errors should result in compilation failure

### LINKED ISSUE #335: LLVM verifier errors should be treated as test failures
Or at the very least treated like other point method failures.

We're still silently letting this pass by.

erozenfeld:
Re-apply the change that was accidentally removed in 93b82adac634fe93c7b798f6d2a855b601ee9310. Verification errors
result in assert or compilation failure.

Closes #335.


JosephTremoulet:
LGTM

PR #569, opened 2015-05-12
TEST: Run tests with INSERTSTATEPOINTS

swaroop-sridhar:
I am **not** planning to checkin this change.
This PR is only to run the standard tests with statepoints inserted.
If this succeeds, we'll add test legs to run the standard tests with COMPLUS_InsertStatepoints set.


swaroop-sridhar:
Closing the PR since the test succeeded. Testing for issue https://github.com/dotnet/llilc/issues/32

PR #572, opened 2015-05-13
Adjust GC transition codegen to accomodate patchable statepoints.

pgavlin:
Patchable statepoint support added 2 additional arguments to the start of
the statepoint intrinsic. Pass zeroes for these arguments, as we don't need
either of them.

### DIFF
--- a/lib/Reader/abisignature.cpp
+++ b/lib/Reader/abisignature.cpp
@@ -139,6 +139,7 @@ CallSite ABICallSignature::emitUnmanagedCall(GenIR &Reader, Value *Target,
   LLVMContext &LLVMContext = *JitContext.LLVMContext;
   Type *Int8Ty = Type::getInt8Ty(LLVMContext);
   Type *Int32Ty = Type::getInt32Ty(LLVMContext);
+  Type *Int64Ty = Type::getInt64Ty(LLVMContext);
   Type *Int8PtrTy = Reader.getUnmanagedPointerType(Int8Ty);
   IRBuilder<> &Builder = *Reader.LLVMBuilder;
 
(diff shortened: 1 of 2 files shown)


pgavlin:
This is necessary to accommodate changes from https://github.com/Microsoft/llvm/pull/41.

pgavlin:
@AndyAyersMS @swaroop-sridhar PTAL

lib/Reader/abisignature.cpp
```diff
@@ -230,21 +231,23 @@ CallSite ABICallSignature::emitUnmanagedCall(GenIR &Reader, Value *Target,
...
+  IntrinsicArgs[0] = ConstantInt::get(Int64Ty, 0);
+  IntrinsicArgs[1] = ConstantInt::get(Int32Ty, 0);
+  IntrinsicArgs[2] = Target;
+  IntrinsicArgs[3] = ConstantInt::get(Int32Ty, TargetArgCount);
+  IntrinsicArgs[4] =
       ConstantInt::get(Int32Ty, (uint32_t)StatepointFlags::GCTransition);
 
   uint32_t I, J;
-  for (I = 0, J = 3; I < TargetArgCount; I++, J++) {
+  for (I = 0, J = 5; I < TargetArgCount; I++, J++) {
```
AndyAyersMS: Isn't `J`'s initial value `PrefixArgCount`? If so you should use it here.

AndyAyersMS:
One small nit, otherwise LTGM.

swaroop-sridhar:
Green for me too

lib/Reader/abisignature.cpp
pgavlin: Done.

PR #573, opened 2015-05-13
Update binaryOpType to handle more cases.

### LINKED ISSUE #558: Type conversion failure in binary op
In `System.Diagnostics.Tracing.DataCollector.ScalarsEnd`, we see:
```
  .locals init (valuetype System.Diagnostics.Tracing.EventSource/EventData* V_0)
  ...
  IL_0016:  ldloc.0
  IL_0017:  ldfld      int64 System.Diagnostics.Tracing.EventSource/EventData::m_Ptr
  IL_001c:  conv.ovf.u
  IL_001d:  sub
  IL_001e:  ldc.i4.1
  IL_001f:  div
```
we type the result of the `sub` as a pointer, and can't handle this when we hit the `div`.

AndyAyersMS:
`binaryOpType` now follows Ecma-355 table III.2 with some exceptions for internal use and for patterns seen in IL stubs.

No diffs in the existing test cases.

Closes #558.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -860,7 +860,18 @@ private:
   /// integral values are zero extended
   IRNode *convert(llvm::Type *Type, llvm::Value *Node, bool IsSigned);
 
-  llvm::Type *binaryOpType(llvm::Type *Type1, llvm::Type *Type2);
+  /// \brief Determine the result type of a binary op.
+  ///
+  /// Rougly follows table III.2 of Ecma-335, with some extra cases added
+  /// because LLILC and IL stubs use this in non-standard ways. Note the result
+  /// is opcode-dependent and that Type1 and Type2 are not symmetric.
+  ///
+  /// \param Opcode  The binary opcode.
+  /// \param Type1   Type of the first operand popped from the operand stack.
+  /// \param Type2   Type of the second operand popped from the operand stack.
+  /// \returns       Type of the result.
+  llvm::Type *binaryOpType(ReaderBaseNS::BinaryOpcode Opcode, llvm::Type *Type1,
+                           llvm::Type *Type2);
 
   IRNode *genPointerAdd(IRNode *Arg1, IRNode *Arg2);
   IRNode *genPointerSub(IRNode *Arg1, IRNode *Arg2);
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -2860,7 +2860,7 @@ IRNode *GenIR::binaryOp(ReaderBaseNS::BinaryOpcode Opcode, IRNode *Arg1,
 
   Type *Type1 = Arg1->getType();
   Type *Type2 = Arg2->getType();
-  Type *ResultType = binaryOpType(Type1, Type2);
+  Type *ResultType = binaryOpType(Opcode, Type1, Type2);
 
   // If the result is a pointer, see if we have simple
   // pointer + int op...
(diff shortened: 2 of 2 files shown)


AndyAyersMS:
@JosephTremoulet PTAL.

lib/Reader/readerir.cpp
```diff
@@ -2952,40 +2952,69 @@ IRNode *GenIR::binaryOp(ReaderBaseNS::BinaryOpcode Opcode, IRNode *Arg1,
...
-    return Type1;
-  } else if (Type2->isFloatingPointTy()) {
-    return Type2;
   }
 
-  if (Size1 > Size2) {
-    return Type1;
-  } else {
-    return Type2;
+  // For sub: If Type1 is a native pointer and Type2 is a native pointer, the
```
JosephTremoulet: I think we need to allow this for add as well.

lib/Reader/readerir.cpp
```diff
@@ -2952,40 +2952,69 @@ IRNode *GenIR::binaryOp(ReaderBaseNS::BinaryOpcode Opcode, IRNode *Arg1,
...
+  // result is a native int.
+  if (IsSub) {
+    if (isUnmanagedPointerType(Type1) && isUnmanagedPointerType(Type2)) {
+      return Type::getIntNTy(*JitContext->LLVMContext, TargetPointerSizeInBits);
+    }
+  }
+
+  // For add: if Type2 is a native pointer and Type1 is an integer, the result
```
JosephTremoulet: I think we need to allow this for sub as well.

include/Reader/readerir.h
```diff
@@ -860,7 +860,18 @@ class GenIR : public ReaderBase {
   /// integral values are zero extended
   IRNode *convert(llvm::Type *Type, llvm::Value *Node, bool IsSigned);
 
-  llvm::Type *binaryOpType(llvm::Type *Type1, llvm::Type *Type2);
+  /// \brief Determine the result type of a binary op.
+  ///
+  /// Largely follows table III.2 of Ecma-355, with some extra cases added
```
erozenfeld: Ecma-355 --> Ecma-335

lib/Reader/readerir.cpp
```diff
@@ -2952,40 +2952,69 @@ IRNode *GenIR::binaryOp(ReaderBaseNS::BinaryOpcode Opcode, IRNode *Arg1,
...
-                             TargetPointerSizeInBits);
-    }
-    ASSERTNR(!Type2->isFloatingPointTy());
-    return Type1;
-  } else if (Type2->isPointerTy()) {
-    ASSERTNR(!Type1->isFloatingPointTy());
-    return Type2;
+Type *GenIR::binaryOpType(ReaderBaseNS::BinaryOpcode Opcode, Type *Type1,
+                          Type *Type2) {
+  // Largely follows ECMA-355, Table III.2.
```
erozenfeld: ECMA-355 --> ECMA-335

JosephTremoulet:
We've convinced ourselves that it's valid IL that we need to handle if any combination of pointers and ints show up on any binary integer operation, right?  I think it's ok allow ptr+int=ptr, int+ptr=ptr, ptr-int=ptr, and ptr-ptr=int, but for all other cases on integer operations (including mul/div), I think we need to throw our hands up and convert both inputs to int.

JosephTremoulet:
(I think Table III.2 would justify us rejecting an `int64` paired with an `int32`, but on 64-bit platforms we're not tracking the difference between `int64` and `native int` in the stack operand types, so at least in that configuration we have to allow `int64` paired with `int32`, I think)

AndyAyersMS:
I'm happy to add more cases here, but I haven't seen these occur anywhere yet.

AndyAyersMS:
@JosephTremoulet @erozenfeld updated, PTAL.

lib/Reader/readerir.cpp
```diff
@@ -2952,40 +2952,64 @@ IRNode *GenIR::binaryOp(ReaderBaseNS::BinaryOpcode Opcode, IRNode *Arg1,
...
-    if (Type2->isFloatingPointTy()) {
-      return Size1 > Size2 ? Type1 : Type2;
+    // int + ptr = ptr
+    if (IsAdd && Type1IsInt && Type2IsUnmanagedPointer) {
+      return Type2;
     }
-    return Type1;
-  } else if (Type2->isFloatingPointTy()) {
-    return Type2;
+    // Otherwise type result as native int.
```
JosephTremoulet: On a 32-bit platform, if we see unmanaged pointer and int64, this would truncate the int64.  That seems a little odd, but since Table III.2 indicates that (`native int`, `int64`) and (`int32`, `int64`) are not valid combinations, I think it's ok (in that situation, the unmanaged pointer presumably should have stack type `int32` or `native int`, and the i64 should have stack type `int64).
I think that rejecting those cases would be ok too; whichever way you prefer.

JosephTremoulet:
LGTM, just one take-it-or-leave-it observation.

lib/Reader/readerir.cpp
erozenfeld: My preference is to reject invalid combinations.

JosephTremoulet: In that case, we should consider rejecting `int32`x`int64` when the target is 32-bits as well.

AndyAyersMS:
So it sounds like we can just insist that sizes match in the non-float cases, and that will be a bit permissive on 64 bit platforms, since native int and int64 should be invalid.

JosephTremoulet:
No, I think we have to allow `i32`x`i64` on x64 -- if we were tracking `native int` and `int64` separately, then yes I'd agree that per Table III.2 we could reject `int32`x`int64`, but since we track CIL types `native int` and `int64` both as LLVM type `i64`, we have to allow `i32`x`i64` since it might be standing in for `int32`x`native int`.

AndyAyersMS:
Right, just hit this case... so for int x int, either sizes match, or one is native int and the other is smaller.

JosephTremoulet:
> For int x int, either sizes match, or one is native int and the other is smaller
> For ptr x int, int x ptr, and ptr x ptr, we'll insist that no truncation happen

Both of those rules sound good to me.

AndyAyersMS:
Ok, posted a new update. @erozenfeld @JosephTremoulet PTAL.

JosephTremoulet:
LGTM, thanks

erozenfeld:
LGTM too.

AndyAyersMS:
Needed to fix formatting... will merge once the CI is green.

Issue #557, opened 2015-05-08
For some types the parent class hierarchy is incomplete.


AndyAyersMS:
One example is `System.Runtime.InteropServices.WindowsRuntime.RuntimeClass` from mscorlib, which is referenced from `System.Threading.Tasks.AsyncCausalityTracer.TracingStatusChangedHandler` and perhaps other places.

`RuntimeClass` extends `System.__ComObject` but when the EE is asked about the parent class it returns `System.Object`. That causes confusion in our `getClassType` code because we know the ultimate object has one field, but neither `RuntimeClass` nor any of its reported parents has any fields.

AndyAyersMS:
Jan Kotas suggested modifying the reported number of fields in the jit interface, but there are lots of consumers of this information and I'm not confident how many of them can tolerate being slightly misled.

Since this only happens for ref classes we can just ignore missing fields for now, since we don't need to know the GC layout, and LLILC is already set up to cope with missing field info from overlap and other similar cases.

AndyAyersMS:
We have a workaround in place that should be sufficient... will leave this issue open in case we want to go back into the jit interface and try and extract more accurate information.

PR #574, opened 2015-05-14
Fixes and workarounds for "expand all" mode when run on simple bring-…

### LINKED ISSUE #555: Integer convert overflow check must handle pointer typed inputs
Looks like the convert code handles this ok, but not the overflow check part.

An example from `System.UIntPtr.ToInt32`
```
  IL_0000:  ldarg.0
  IL_0001:  ldfld      void* System.UIntPtr::m_value
  IL_0006:  conv.ovf.u4.un
```

### LINKED ISSUE #559: Missing sign/zero extension or perhaps bad IL
In `System.IO.Directory.InternalCreateDirectory` we see
```
  int32 V_8,
  native int V_21,
  ...
  IL_00fe:  ldloc.s    V_8
  IL_0100:  dup
  IL_0101:  stloc.s    V_21
```
and get confused at the `stloc.s` in `convertFromStackType` because we have an i32 operand on the stack and want an i64, and don't expect to perform a lengthening convert there.

Not sure offhand if we should just do the extension here (sign? zero?) or if the IL is invalid.

### LINKED ISSUE #571: Unexpected type in getFieldAddress
In `System.Diagnostics.Tracing.EventProvider.EncodeObject`, we see
```
  IL_0000:  ldarg.1
  IL_0001:  ldind.i
  IL_0002:  ldc.i4.0
  IL_0003:  stfld      uint32 System.Diagnostics.Tracing.EventProvider/EventData::Reserved
```
where `arg1` is
```
valuetype System.Diagnostics.Tracing.EventProvider/EventData*& dataDescriptor
```
this leads us to an assert in `getFieldAddress` because the object type is not a pointer type.

AndyAyersMS:
…up tests.

These fixes allow us to run simple jit tests with complus_expandallonload=1.

Implicitly extend int32 to native int when popping the stack. Closes #559.

Reinterpret native int as unmanaged pointer in `getFieldAddress`. Closes #571.

Reinterpret unmanaged pointer as native int in `genConvertOverflowCheck`. Closes #555.

Bail out if we can't account for all the fields in the type. Workaround for #557.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -1410,6 +1410,11 @@ Type *GenIR::getClassType(CORINFO_CLASS_HANDLE ClassHandle, bool IsRefClass,
 
     for (uint32_t I = 0; I < NumDerivedFields; I++) {
       CORINFO_FIELD_HANDLE FieldHandle = getFieldInClass(ClassHandle, I);
+      if (FieldHandle == nullptr) {
+        // Likely a class that derives from System.__ComObject. See
+        // LLILC issue #557.
+        throw NotYetImplementedException("can't account for all fields");
+      }
       ASSERT(FieldHandle != nullptr);
       const uint32_t FieldOffset = getFieldOffset(FieldHandle);
       DerivedFields.push_back(std::make_pair(FieldOffset, FieldHandle));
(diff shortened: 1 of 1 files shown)


pgavlin:
LGTM.

PR #575, opened 2015-05-15
Enable zero and sign extension for call args/results.

pgavlin:
RyuJIT expects all calls to/from managed code to follow a producer-extends
model for integral arguments that are less than four bytes in size.

### DIFF
--- a/include/Reader/abi.h
+++ b/include/Reader/abi.h
@@ -29,6 +29,10 @@ public:
     Direct, ///< Pass the argument directly, optionally coercing it to a
             ///< different type.
 
+    ZeroExtend, ///< Pass the argument directly with zero-extension.
+
+    SignExtend, ///< Pass the argument directly with sign-extension.
+
     Indirect, ///< Pass the argument indirectly via a hidden pointer
   };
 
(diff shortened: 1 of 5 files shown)


pgavlin:
@erozenfeld @russellhadley @JosephTremoulet PTAL

JosephTremoulet:
LGTM, assuming I'm following correctly that:
1. Applying Attribute::SExt will sign-extend to 64 bits (not just to 32 bits), and either that's what we require or we don't care what's in the top 32 bits.
2. The fact that you're applying the attributes to formal parameters in `createFunction` and formal return values in `emitCall` means that we can take advantage of the guarantees on the consumer side (e.g. that we can eliminate the sign/zero-extend we insert as part of ldarg [after eliminating load, in cases where the arg isn't aliased])

Also, I'm curious why RyuJIT does this.

pgavlin:
> Applying Attribute::SExt will sign-extend to 64 bits (not just to 32 bits

From what I can tell, applying either attribute sign-/zero-extends to the size of the nearest representative register (i.e. the smallest register that is not a subregister). For X86, this is generally a 32-bit GPR, so we get 32-bit sign-/zero-extension. This is what RyuJIT expects.

> The fact that you're applying the attributes to formal parameters in createFunction and formal return values in emitCall means that we can take advantage of the guarantees on the consumer side

Yes, that's correct.

> Also, I'm curious why RyuJIT does this.

I'll defer to @cmckinsey on this one.

JosephTremoulet:
> For X86, this is generally a 32-bit GPR, so we get 32-bit sign-/zero-extension

Sorry, I meant to ask about x64 -- we don't extend 32-bit values, but we extend 16-bit and 8-bit values to 64 rather than 32 bits, do I have that right?

pgavlin:
> but we extend...to 64 rather than 32 bits

No, the situation is the same for x64 (i.e. on extension is to 32 bits).

russellhadley:
Looks good.  Go for it.

PR #579, opened 2015-05-18
Fix a bug in GenIR::arraySet.

pgavlin:
GenIR::arraySet was not converting the value to be stored from its stack
type to its actual type. This change adds the missing call to
convertFromStackType.


pgavlin:
@erozenfeld @AndyAyersMS PTAL

erozenfeld:
LGTM.

AndyAyersMS:
LGTM too.

PR #580, opened 2015-05-19
Disallow first-class aggregates in LLVM IR we are generating.

### LINKED ISSUE #407: Challenges with named types and shared generics
We're starting to see cases now where we have types that are structurally equivalent but not name equivalent. This happens when we have shared generics. This leads to typing violations in LLVM IR, since LLVM uses name equivalence if you give types names. We've gotten away with it for a while because with generic ref classes we could cast our way out of trouble (though Eugene has an example where even this fails). But for value classes we can't fix things easily by casting.

One example is from TinyCsc.Main in Roslyn, here we have:
```
.locals 
   ....
           valuetype [System.Collections.Immutable]System.Collections.Immutable.ImmutableArray`1/Enumerator<class [Microsoft.CodeAnalysis]Microsoft.CodeAnalysis.Diagnostic> V_16,
   ....
    IL_018e:  call       instance valuetype [System.Collections.Immutable]System.Collections.Immutable.ImmutableArray`1/Enumerator<!0> valuetype [System.Collections.Immutable]System.Collections.Immutable.ImmutableArray`1<class [Microsoft.CodeAnalysis]Microsoft.CodeAnalysis.Diagnostic>::GetEnumerator()
    IL_0193:  stloc.s    V_16
```
At the `stloc` we have type 
```
%"System.Collections.Immutable.ImmutableArray`1+Enumerator[System.__Canon]" = type <{ %"System.__Canon[]" addrspace(1)*, i32, [4 x i8] }>
``` 
on the stack, and type 
```
%"System.Collections.Immutable.ImmutableArray`1+Enumerator[Microsoft.CodeAnalysis.Diagnostic]" = type <{ %"System.__Canon[]" addrspace(1)*, i32, [4 x i8] }>
```
for the local.

Since these are structs, we can't just cast from one type to the other.

Either we need to canonicalize the names somehow (which seems tricky since we don't know offhand when the EE will decide to share) or we need to give up on having the names there in the first place. Or maybe some other clever fix.

### LINKED ISSUE #441: Reconsider allowing first-class aggregates in LLVM IR
We currently allow first-class aggregate (FCA) values to appear in LLVM IR, from value classes on the evaluation stack or passed/return to calls. This is not wrong but causes a couple of problems now and will likely cause more problems later.

The immediate problem is type mismatches both in straight-line IR and at PHIs See for instance #407. This isn't super common (seemingly requires a generic value type with a ref class field that the EE decides to share) but it does happen.

The downstream problem is potential poor optimization of FCAs by the LLVM optimizer. The canonical example of this is ldobj / ldfld, we need to ensure the entire object isn't copied just to access one field. In general getting away from FCAs allows us to control where and when copies are made and enables us to look at things like RVO or not making defensive copies for call arguments if we know the callee  is not going to write to the value type.

erozenfeld:
They were causing struct type mismatches due to generic sharing
and also had potential optimization issues.

No values of StructType are allowed on the operand stack.
Instead pointers to structs are pushed to the stack and
recorded in StructPointers set.

When a struct needs to be loaded we make a copy of the struct
and return a pointer to the copy.

addressOfValue doesn't need to copy the struct. Instead a pointer
to the struct is copied.

Struct type mismatches on the stack and at PHI joins
are handled by casting the pointers representing the structs.

Closes #407.
Closes #441.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -2769,7 +2769,7 @@ public:
   /// \param Class  Class handle that corresponds to RuntimeTypeHandle,
   ///               RuntimeMethodHandle, or RuntimeFieldHandle.
   ///
-  /// \returns Runtime handle corresponding to the token node..
+  /// \returns Runtime handle corresponding to the token node.
   virtual IRNode *convertHandle(IRNode *RuntimeTokenNode,
                                 CorInfoHelpFunc HelperID,
                                 CORINFO_CLASS_HANDLE Class) = 0;
@@ -2793,6 +2793,16 @@ public:
     return loadPrimitiveType(Address, CorType, Alignment, IsVolatile,
                              IsInterfConst, false);
   }
+
+  /// \brief Load a non-primitive object (i.e., a struct).
+  ///
+  /// \param Address Address of the struct.
+  /// \param Class Class handle corresponding to the struct.
+  /// \param Alignment Alignment of the load.
+  /// \param IsVolatile true iff this is a volatile load.
+  /// \param AddressMayBeNull true iff Address may be null.
+  ///
+  /// \returns Node representing loaded struct.
   virtual IRNode *loadNonPrimitiveObj(IRNode *Address,
                                       CORINFO_CLASS_HANDLE Class,
                                       ReaderAlignType Alignment,
@@ -3240,6 +3250,13 @@ public:
   // Used to expand multidimensional array access intrinsics
   virtual bool arrayGet(CORINFO_SIG_INFO *Sig, IRNode **RetVal) = 0;
   virtual bool arraySet(CORINFO_SIG_INFO *Sig) = 0;
+
+  /// \brief Check whether structs are represented by pointers on the operand
+  /// stack.
+  ///
+  /// \returns true iff structs are represented by pointers on the operand
+  /// stack.
+  virtual bool structsAreRepresentedByPointers() = 0;
 
 #if !defined(NDEBUG)
   virtual void dbDumpFunction(void) = 0;
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -419,6 +419,25 @@ public:
                               ReaderAlignType Alignment, bool IsVolatile,
                               bool AddressMayBeNull = true) override;
 
+  /// \brief Load a non-primitive object (i.e., a struct).
+  ///
+  /// \param StructTy Type of the struct.
+  /// \param Address Address of the struct.
+  /// \param Alignment Alignment of the load.
+  /// \param IsVolatile true iff this is a volatile load.
+  /// \param AddressMayBeNull true iff Address may be null.
+  ///
+  /// \returns Node representing loaded struct.
+  IRNode *loadNonPrimitiveObj(llvm::StructType *StructTy, IRNode *Address,
+                              ReaderAlignType Alignment, bool IsVolatile,
+                              bool AddressMayBeNull = true);
+
+  IRNode *loadNonPrimitiveObjNonNull(llvm::StructType *StructTy,
+                                     IRNode *Address, ReaderAlignType Alignment,
+                                     bool IsVolatile) {
+    return loadNonPrimitiveObj(StructTy, Address, Alignment, IsVolatile, false);
+  }
+
   IRNode *makeRefAny(CORINFO_RESOLVED_TOKEN *ResolvedToken,
                      IRNode *Object) override {
     throw NotYetImplementedException("makeRefAny");
@@ -822,6 +841,8 @@ public:
   bool arrayGet(CORINFO_SIG_INFO *Sig, IRNode **RetVal) override;
   bool arraySet(CORINFO_SIG_INFO *Sig) override;
 
+  bool structsAreRepresentedByPointers() override { return true; }
+
 #if !defined(NDEBUG)
   void dbDumpFunction(void) override {
     throw NotYetImplementedException("dbDumpFunction");
@@ -922,6 +943,18 @@ private:
                          IsVolatile, false);
   }
 
+  IRNode *loadAtAddress(IRNode *Address, llvm::Type *Ty, CorInfoType CorType,
+                        ReaderAlignType AlignmentPrefix, bool IsVolatile,
+                        bool AddressMayBeNull = true);
+
+  IRNode *loadAtAddressNonNull(IRNode *Address, llvm::Type *Ty,
+                               CorInfoType CorType,
+                               ReaderAlignType AlignmentPrefix,
+                               bool IsVolatile) {
+    return loadAtAddress(Address, Ty, CorType, AlignmentPrefix, IsVolatile,
+                         false);
+  }
+
   /// Generate instructions for storing value of the specified type at the
   /// specified address.
   ///
@@ -947,6 +980,17 @@ private:
     return storeAtAddress(Address, ValueToStore, Ty, ResolvedToken,
                           AlignmentPrefix, IsVolatile, IsField, false);
   }
+
+  /// Generate instructions for storing value of the specified type at the
+  /// specified address. The caller must guarantee that address is not null and
+  /// no write barrier is needed.
+  ///
+  /// \param Address Address to store to.
+  /// \param ValueToStore Value to store.
+  /// \param Ty llvm type of the value to store.
+  /// \param IsVolatile true iff the store is volatile.
+  void storeAtAddressNoBarrierNonNull(IRNode *Address, IRNode *ValueToStore,
+                                      llvm::Type *Ty, bool IsVolatile);
 
   void classifyCmpType(llvm::Type *Ty, uint32_t &Size, bool &IsPointer,
                        bool &IsFloat);
@@ -1145,8 +1189,13 @@ private:
   ///
   /// \param Ty1 Type of the first value.
   /// \param Ty1 Type of the second value.
+  /// \param IsStruct1 true iff value corresponding to Ty1 logically represents
+  /// a struct rather than a pointer to struct.
+  /// \param IsStruct2 true iff value corresponding to Ty2 logically represents
+  /// a struct rather than a pointer to struct.
   /// \returns The result type.
-  llvm::Type *getStackMergeType(llvm::Type *Ty1, llvm::Type *Ty2);
+  llvm::Type *getStackMergeType(llvm::Type *Ty1, llvm::Type *Ty2,
+                                bool IsStruct1, bool isStruct2);
 
   bool objIsThis(IRNode *Obj);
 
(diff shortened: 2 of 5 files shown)


erozenfeld:
@AndyAyersMS @pgavlin PTAL.

lib/Reader/readerir.cpp
```diff
@@ -6588,7 +6704,8 @@ Value *GenIR::ChangePHIOperandType(Value *Operand, BasicBlock *OperandBlock,
   }
 }
 
-Type *GenIR::getStackMergeType(Type *Ty1, Type *Ty2) {
+Type *GenIR::getStackMergeType(Type *Ty1, Type *Ty2, bool IsStruct1,
+                               bool IsStruct2) {
```
AndyAyersMS: Seems like IsStruct1 must == IsStruct2, otherwise we have a problem.

include/Reader/reader.h
```diff
@@ -3241,6 +3251,13 @@ class ReaderBase {
   virtual bool arrayGet(CORINFO_SIG_INFO *Sig, IRNode **RetVal) = 0;
   virtual bool arraySet(CORINFO_SIG_INFO *Sig) = 0;
 
+  /// \brief Check whether structs are represented by pointers on the operand
+  /// stack.
+  ///
+  /// \returns true iff structs are represented by pointers on the operand
+  /// stack.
+  virtual bool structsAreRepresentedByPointers() = 0;
```
AndyAyersMS: It feels like support for the "old way" where we allowed first class structs is only partially plumbed through. I'm not sure it adds much value. So I'd suggest removing this and cleaning out the code that this guards.

erozenfeld: There is one place in ReaderBase (in storeObj) that needs this info and it didn't fell right to hard-code it to true there. GenIR has it hard-coded to true and never calls this method.

lib/Reader/readerir.cpp
erozenfeld: Yes, I will add an assert.

lib/Reader/readerir.cpp
```diff
@@ -2069,20 +2094,29 @@ IRNode *GenIR::convertFromStackType(IRNode *Node, CorInfoType CorType,
   }
 
   case Type::TypeID::PointerTyID: {
+    bool PointerRepresentsStruct = doesValueRepresentStruct(Node);
+    if (PointerRepresentsStruct) {
+      if (Ty->getPointerElementType() != ResultTy) {
```
AndyAyersMS: I would add an assert here that `ResultTy` is a `StructType` and that the layout of the two structs is the same (similar to the merge case). We don't want to reinterpret a struct's layout here.

lib/Reader/readerir.cpp
```diff
@@ -3675,6 +3746,16 @@ IRNode *GenIR::addressOfValue(IRNode *Leaf) {
     StoreInst *Store = LLVMBuilder->CreateStore(Leaf, Alloc);
     return (IRNode *)Alloc;
   }
+  case Type::TypeID::PointerTyID: {
```
AndyAyersMS: Should we make struct types invalid here?

lib/Reader/readerir.cpp
```diff
@@ -3675,6 +3746,16 @@ IRNode *GenIR::addressOfValue(IRNode *Leaf) {
...
+    assert(doesValueRepresentStruct(Leaf));
+
+    // Create a new pointer to the struct. The new pointer will point to the
+    // same struct as the original pointer but doesValueRepresentStruct will
+    // return false for the new pointer.
+    Instruction *Alloc = createTemporary(LeafTy);
+    StoreInst *Store = LLVMBuilder->CreateStore(Leaf, Alloc);
+    return (IRNode *)LLVMBuilder->CreateLoad(Alloc);
+  }
```
AndyAyersMS: I don't understand why we can't just return the existing pointer.

include/Reader/reader.h
AndyAyersMS: I suppose it's ok, my guess is the factoring between reader base and reader is off somewhere and reader base should just be told this and not have to know it (via something like the `IsValueIsPointer` parameter), but let's not worry about it now.

pgavlin: I agree with Andy. `ReaderBase` shouldn't need to know how values are represented by its subclasses. This does feel like something we can fix up later, however.

AndyAyersMS:
A couple of questions but overall looks pretty close.

This intersects with the `mkrefany` that I have coded up; I'll hold off on that one until this is in.

include/Reader/readerir.h
```diff
@@ -1453,6 +1524,15 @@ class GenIR : public ReaderBase {
...
   llvm::DenseMap<uint32_t, llvm::StoreInst *> ContinuationStoreMap;
+  llvm::SmallPtrSet<llvm::Value *, 5> StructPointers; ///< This set contains
+                                                      ///< pointers to structs
+                                                      ///< that we put on
+                                                      ///< the operand stack
+                                                      ///< to represent structs
+                                                      ///< since we don't allow
+                                                      ///< first-class
+                                                      ///< aggregates in LLVM IR
+                                                      ///< we are generating.
```
pgavlin: This comment feels a bit misleading: it looks like we track more than just what we put on the operand stack. Also, where did '5' come from?

lib/Reader/abisignature.cpp
```diff
@@ -96,18 +96,33 @@ ABISignature::ABISignature(const ReaderCallSignature &Signature, GenIR &Reader,
...
-  Type *ValueType = TheValue->getType();
-
-  if (TheType == ValueType) {
-    return TheValue;
+  bool PointerRepresentsStruct = Reader.doesValueRepresentStruct(TheValue);
+  Type *ValueType = nullptr;
+  Value *ValuePtr = nullptr;
+  if (PointerRepresentsStruct) {
+    ValueType = TheValue->getType()->getPointerElementType();
+    if ((TheType == ValueType) && TheType->isStructTy()) {
```
pgavlin: Should the second half of this condition just be an assert in the `if` body?

lib/Reader/abisignature.cpp
```diff
@@ -436,7 +461,13 @@ Value *ABICallSignature::emitCall(GenIR &Reader, Value *Target, bool MayThrow,
       ResultNode = Call.getInstruction();
     }
   } else {
-    ResultNode = Builder.CreateLoad(ResultNode);
+    Type *ResultTy = ResultNode->getType();
+    if (!ResultTy->isPointerTy() ||
+        !ResultTy->getPointerElementType()->isStructTy()) {
+      ResultNode = Builder.CreateLoad(ResultNode);
+    } else {
+      Reader.setValueRepresentsStruct(ResultNode);
```
pgavlin: I wonder if it might be better to move this up into the indirect result setup (see lines 351-354). If you call `GenIR::setValueRepresentsStruct` up there, this becomes

```
if (!Reader.doesValueRepresentStruct(ResultNode)) {
    ResultNode = Builder.CreateLoad(ResultNode);
}
```

include/Reader/reader.h
erozenfeld: storeObj is called not only from GenIR but also from ReaderBase directly. So ReaderBase needs some way to ask GenIR about how it should interpret the node representing struct that it pops from the stack. Do you have a suggestion on how this can be done cleaner?

AndyAyersMS: I haven't worked this out completely, but if `addressOfValue` is suitably savvy then we perhaps could always call it from `rdrCallWriteBarrierHelper` for the value class case, and (depending on client strategy) it would either just return the address it already has or save the struct to a temp and return that temp's address. Then we don't need to have this info visible at the base layer.

I'd be fine if you leave what you have for this and we can fix this part up later.

lib/Reader/readerir.cpp
erozenfeld: I want to keep an invariant that a pointer value either represents a struct or a pointer to struct. If we don't have that invariant code that makes decisions based on whether a pointer represents a struct will get confused. The input value represents a struct. The return value represents a pointer to struct so I'm making a copy of the pointer.

include/Reader/readerir.h
erozenfeld: I'll update the comment. 5 is an arbitrary small int. I think in the common case we won't have many of these in a method. Do you have a suggestion for a different const or a different data structure here?

lib/Reader/abisignature.cpp
erozenfeld: No. TheType can be a primitive type in case of a small struct.

pgavlin: But wouldn't `TheType == ValueType` be false in that case? I didn't think that we ever coerced structs to primitives until this point...

erozenfeld: I'm pretty sure we do. Let me double check.

lib/Reader/readerir.cpp
AndyAyersMS: Ok, makes sense now.

include/Reader/readerir.h
pgavlin: I think you've got the right data structure--I was just wondering whether or not '5' was arbitrary.

lib/Reader/readerir.cpp
erozenfeld: Done.

lib/Reader/readerir.cpp
erozenfeld: Yes. Done.

lib/Reader/abisignature.cpp
erozenfeld: I was wrong, I got this case confused with something else. I took your suggestion.

lib/Reader/abisignature.cpp
erozenfeld: Done.

erozenfeld:
@pgavlin @AndyAyersMS I addressed all feedback. PTAL

AndyAyersMS:
LGTM.

lib/Reader/abisignature.cpp
```diff
@@ -338,6 +354,12 @@ Value *ABICallSignature::emitCall(GenIR &Reader, Value *Target, bool MayThrow,
         Reader.getUnmanagedPointerType(Result.getType());
     Arguments[ResultIndex] = ResultNode =
         Reader.createTemporary(Result.getType());
+
+    Type *ResultTy = ResultNode->getType();
+    if (ResultTy->isPointerTy() &&
```
pgavlin: I would rewrite this check as `if (Result.getType()->isStructTy())`; the `isPointerTy()` check is redundant.

pgavlin:
LGTM as well. The test failure looks spurious.

lib/Reader/abisignature.cpp
erozenfeld: Done.

erozenfeld:
The same spurious failure on retest with the update. Merging now.

PR #581, opened 2015-05-19
Improve llilc_checkpass failure detection.

pgavlin:
llilc_checkpass was previously checking for failures by hunting for lines
in a diff that began with '-'. Unfortunately, this is not accurate: most
importantly, it fails to account for the fact that methods may simply no
longer be submitted for compilation (generally due to changes to the
libraries or tests).

This change switches to a set-based implementation in order to accurately
detect methods meeting any of the following criteria:
- Compiled cleanly in the baseline and are now failing to compile
- Failed to compile in the baseline and are now compiling cleanly
- Processed in the baseline but not processed now
- Not processed in the baseline and processed now

A run is only a failure if there are methods that used to compile cleanly
that now fail to compile. A count for each category is reported in the output.

### DIFF
--- a/test/llilc_checkpass.py
+++ b/test/llilc_checkpass.py
@@ -4,18 +4,21 @@
 #description     :
 # 
 # llilc_checkpass compares two test results, no matter summary or verbose,
-# and checks if any function successfully jitted by LLILC in base result
-# failed jitting in diff result.
+# and checks if any function successfully compiled by LLILC in base result
+# failed to compile in the target result or if any function that was newly
+# submitted for compilation failed to compile. Either of these situations
+# is considered an unexpected failure. The script returns the number of
+# unexpected failures, where 0 indicates that all tests passed. If the script
+# fails due to an unexpected environmental issue, the return value will be
+# negative, and has no relation to the number of unexpected failures.
 #
-# If the check passed, the script return 0. In case of 0, it might have
-# jitted more functions in diff result than in base result. The two results
-# are not necessary the same in this case. Counting result is reported.
+# Whether or not all tests passed, the set of methods submitted for compilation
+# may differ between the base result and the target result. The number of
+# methods that were only submitted to the base compiler and the number of
+# methods that were onty submitted to the target compiler are reported on
+# stdout.
 # 
-# If the check failed due to newly faily routine, return 1.
-#     
-# If the check failed due to unexpected reason, return negative numbers.
-#
-# usage: llilc_checkpass.py [-h] -b BASE_RESULT_PATH -d DIFF_RESULT_PATH
+# usage: llilc_checkpass.py [-h] -b BASE_RESULT_PATH -d TARGET_RESULT_PATH
 # 
 # optional arguments:
 #   -h, --help            show this help message and exit
@@ -23,8 +26,8 @@
 # required arguments:
 #   -b BASE_RESULT_PATH, --base-result-path BASE_RESULT_PATH
 #                         full path to base result
-#   -d DIFF_RESULT_PATH, --diff-result-path DIFF_RESULT_PATH
-#                         full path to diff result
+#   -d TARGET_RESULT_PATH, --diff-result-path TARGET_RESULT_PATH
+#                         full path to target result
 #
 #==========================================================================================
 
@@ -32,13 +35,11 @@ import argparse
 import re
 import sys
 import os
-import difflib
 import const
+import traceback
 
 def main(argv):
-    # define return code const value
-    const.CheckPassFail = 1
-    const.CheckPassOK = 0
+    # Define return code constants
     const.GeneralError = -1
     const.UnknownArguments = -2
     const.MissingResult = -3
@@ -49,11 +50,11 @@ def main(argv):
     required.add_argument('-b', '--base-result-path', type=str, required=True,
                         help='full path to base result')
     required.add_argument('-d', '--diff-result-path', type=str, required=True,
-                        help='full path to diff result')
+                        help='full path to target result')
     args, unknown = parser.parse_known_args(argv)
 
     if unknown:
-        print('Unknown argument(s): ', ', '.join(unknown))
+        print 'Unknown argument(s): ', ', '.join(unknown)
         return const.UnknownArguments
 
     try:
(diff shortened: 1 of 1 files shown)


pgavlin:
@libengu @JosephTremoulet @AndyAyersMS PTAL

JosephTremoulet:
Couldn't hurt to add comments describing e.g. what the summary output looks like (next to where the regex is defined), what the two maps represent (looks like one's keys are the passing baseline methods, the other's keys are the failing baseline methods, and the values are a visited bit when you walk the diff results?), what the parameters are to the inline function (FWIW, naming one "diff_set" even though it's one of the base_ sets threw me for a loop).

JosephTremoulet:
Seems a bit fast&loose w.r.t. overloads.  Maybe we should change the summary output generation in LLILC to include the method token or something?  (and to be pedantic, maybe the assembly name and parent type name if it's a nested type, but maybe that's overkill for what we need for this script)

JosephTremoulet:
It's an interesting question whether "not processed -> failing" should be flagged as a failure or not.  I think we use this script to track progress in two "modes":
 - bring-up mode: tests are failing in the baseline, when you unblock one you start processing its callees, which may fail and that's still forward progress so it shouldn't be flagged as a failure
 - regression checking mode: tests are passing in the baseline (and we believe we're 100% good on this test), it's a bit weird if a method starts being processed as a result of your change (maybe less weird once we get inlining up), but if there's a new method and it fails then it merits taking a look and so should probably get flagged

Not sure what to do about that... one option would be to try to infer whether to flag these based on whether all methods in the baseline succeeded, another option would be to have this script take a flag telling it whether to report "not processed -> failing" as an error, and a third option would be to have this script just report the transitions and leave it up to calling scripts to parse the output and determine if it constitutes a pass or a failure.

pgavlin:
> Couldn't hurt to add comments describing e.g. what the summary output looks like...

Sure. You've got it right, FWIW. The maps are maps in order to avoid counting things twice in the case of name collisions (e.g. overloads). I thought about using a count instead of a bit for the value, but it didn't quite seem like it was worth it.

> Seems a bit fast&loose w.r.t. overloads. Maybe we should change the summary output generation in LLILC to include the method token or something? (and to be pedantic, maybe the assembly name and parent type name...

This is certainly something worth thinking about. It would be nice if there was some easy way to do name mangling that took all of these things into account.

> It's an interesting question whether "not processed -> failing" should be flagged as a failure or not

"not processed" was giving me a bit of trouble in general. It's very possible that a large change in either missing or new methods is interesting, but I wasn't sure that either condition warranted a 'failure'. Your point about "not processed" -> "failing" is interesting, however, and I think I agree that such a change should be marked as a failure.

As far as the modes you described: I like the idea of having the script report the transitions and leaving it up to the caller. How would you feel about having the script print its usual output and then return the number of unexpected failures (i.e. the sum of "methods that compiled in the baseline and failed in the diff" and "newly processed methods that failed to compile")?

JosephTremoulet:
> I thought about using a count instead of a bit for the value, but it didn't quite seem like it was worth it

Seems worth it to me just to avoid having to reason about how much could slip through with the bools, but I agree it seems like not much bad would get through.

> How would you feel about having the script print its usual output and then return the number of unexpected failures (i.e. the sum of "methods that compiled in the baseline and failed in the diff" and "newly processed methods that failed to compile")?

Sounds fine to me.

pgavlin:
Addressed feedback; PTAL.

test/llilc_checkpass.py
```diff
@@ -66,62 +67,131 @@ def main(argv):
...
-    if len(diff_files) == 0:
-        print('Error: diff result is empty')
+    if len(target_files) == 0:
+        print 'Error: target result is empty'
         return const.MissingResult
 
-    # Counting the newly failed or passed test cases
-    count_new_failed = 0
-    count_new_passed = 0
+    # Track the number of methods meeting any of the following criteria:
```
JosephTremoulet: I think this comment is now stale.

test/llilc_checkpass.py
```diff
@@ -66,62 +67,131 @@ def main(argv):
...
+    print 'CheckPass:', new_target_failures, 'method(s) not processed in base result that failed to compile in target result.'
+
+    unexpected_failures = target_failure_diff + new_target_failures
+
+    if unexpected_failures == 0:
+        print 'CheckPass Passed.'
     else:
-        print('CheckPass Failed.\n')
-        return const.CheckPassFail
+        print 'CheckPass Failed.'
```
JosephTremoulet: I'd maybe avoid "passed"/"failed" here and say something like "there were unexpected failures" / "there were no unexpected failures".

test/llilc_checkpass.py
```diff
@@ -66,62 +67,131 @@ def main(argv):
...
 
-    print('CheckPass: ', count_new_failed, ' test cases passed in base result but failed in diff result.')
-    print('CheckPass: ', count_new_passed, ' test cases failed in base result but passed in diff result.') 
-    if count_new_failed == 0:
-        print('CheckPass Passed.')
-        return const.CheckPassOK
+    print 'CheckPass:', target_failure_diff, 'method(s) compiled successfully in base result that failed in target result.'
+    print 'CheckPass:', target_success_diff, 'method(s) failed in base result that compiled successfully in target result.'
+    print 'CheckPass:', missing_base_methods, 'method(s) processed in base result that were not processed in target result.'
+    print 'CheckPass:', new_target_methods, 'method(s) not processed in base result that were processed in target result.'
```
JosephTremoulet: Seems funny to me that the other four categories you report are disjoint but this one is a superset of new_target_failures; I'd have computed and reported new_target_successes rather than new_target_methods to keep them all disjoint (and FWIW that would let you move the code on 164-168 up to 158 and make it symmetric with the code on 161-162).  But I don't think it's wrong to do it this way if you really prefer it.

test/llilc_checkpass.py
JosephTremoulet: (and ditto for the comment up on line 100 that you changed while I was writing that... :) )

JosephTremoulet:
Thanks for the updates.  LGTM modulo some nits.

PR #583, opened 2015-05-19
Handle ref classes with missing fields.

AndyAyersMS:
This is a more aggressive workaround for #557. We already have the ability to tolerate missing fields in codegen. So we fall back to that if we're building up a model of a ref class and can't account for all of its fields.

For value classes we still insist that we get a complete picture, because we intend to use the LLVM types to produce GC info.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -1412,10 +1412,11 @@ Type *GenIR::getClassType(CORINFO_CLASS_HANDLE ClassHandle, bool IsRefClass,
       CORINFO_FIELD_HANDLE FieldHandle = getFieldInClass(ClassHandle, I);
       if (FieldHandle == nullptr) {
         // Likely a class that derives from System.__ComObject. See
-        // LLILC issue #557.
-        throw NotYetImplementedException("can't account for all fields");
+        // LLILC issue #557. We'll just have to cope with an incomplete
+        // picture of this type.
+        assert(IsRefClass && "need to see all fields of value classes");
+        break;
       }
-      ASSERT(FieldHandle != nullptr);
       const uint32_t FieldOffset = getFieldOffset(FieldHandle);
       DerivedFields.push_back(std::make_pair(FieldOffset, FieldHandle));
     }


AndyAyersMS:
@erozenfeld this fixes a failure issue in Roslyn. PTAL.

lib/Reader/readerir.cpp
```diff
@@ -1411,8 +1411,14 @@ Type *GenIR::getClassType(CORINFO_CLASS_HANDLE ClassHandle, bool IsRefClass,
...
-        // LLILC issue #557.
+        if (IsRefClass) {
+          // Likely a class that derives from System.__ComObject. See
+          // LLILC issue #557. We'll just have to cope with an incomplete
+          // picture of this type.
+          break;
+        }
+        // For value classes we use the LLVM type to derive the GC reporting.
+        // So we can't tolerate any missing fields.
         throw NotYetImplementedException("can't account for all fields");
```
erozenfeld: Should this be an assert instead of an NYI? We don't know of any existing value classes that have this issue, right?

AndyAyersMS: Yeah, let me update this.

AndyAyersMS:
@erozenfeld updated...

lib/Reader/readerir.cpp
```diff
@@ -1412,8 +1412,10 @@ Type *GenIR::getClassType(CORINFO_CLASS_HANDLE ClassHandle, bool IsRefClass,
...
       if (FieldHandle == nullptr) {
         // Likely a class that derives from System.__ComObject. See
-        // LLILC issue #557.
-        throw NotYetImplementedException("can't account for all fields");
+        // LLILC issue #557. We'll just have to cope with an incomplete
+        // picture of this type.
+        assert(IsRefClass && "need to see all fields of value classes");
+        break;
       }
       ASSERT(FieldHandle != nullptr);
```
erozenfeld: Nit: this ASSERT is not needed. We really need to make asserts uniform. Here we have ASSERT and assert almost next to each other.

erozenfeld: LGTM with one nit.

AndyAyersMS:
FWIW I've been gravitating towards `assert(...)`. Removed the `ASSERT`. CI seems confused, will let this update get retested before merging.

PR #586, opened 2015-05-20
Implement support for mkrefany.

### LINKED ISSUE #298: Reader: implement mkrefany msil instruction

AndyAyersMS:
This implements basic support for the `mkrefany` opcode. I created a simple test but System.TypedReference is not exposed by contract assemblies so getting the test to build as part of core CLR tests is nontrivial. Note `refanyval` was buggy before the change to not allow structs on the stack. Now it works ok but is somewhat fragile.

Closes #298.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -2023,10 +2023,10 @@ IRNode *GenIR::convertToStackType(IRNode *Node, CorInfoType CorType) {
   case Type::TypeID::PointerTyID:
   case Type::TypeID::FloatTyID:
   case Type::TypeID::DoubleTyID:
-  case Type::TypeID::StructTyID:
     // Already a valid stack type.
     break;
 
+  case Type::TypeID::StructTyID:
   default:
     // An invalid type
     ASSERTNR(UNREACHED);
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@erozenfeld PTAL.

This fixes the last Roslyn issue...
1 tests, 9861 methods, 9861 good, 0 bad (100% good)

erozenfeld:
I'm probably being blind but the change in RefAnyVal looks like a simple renaming to me. But you said that "refanyval was buggy before the change to not allow structs on the stack. Now it works ok but is somewhat fragile." so I'm confused.

lib/Reader/readerir.cpp
```diff
@@ -6543,6 +6543,40 @@ IRNode *GenIR::loadAndBox(CORINFO_RESOLVED_TOKEN *ResolvedToken, IRNode *Addr,
   return box(ResolvedToken, Value);
 }
 
+IRNode *GenIR::makeRefAny(CORINFO_RESOLVED_TOKEN *ResolvedToken,
+                          IRNode *Object) {
+
+  // Create a new temporary of the right type.
+  CORINFO_CLASS_HANDLE RefAnyHandle = getBuiltinClass(CLASSID_TYPED_BYREF);
+  Type *RefAnyTy = getType(CORINFO_TYPE_VALUECLASS, RefAnyHandle);
+  assert(RefAnyTy->isStructTy() && "expect struct type for refany");
```
erozenfeld: This assert looks redundant both because getType always returns StructTypes for CORINFO_TYPE_VALUECLASS and because you are casting on the next line.

AndyAyersMS:
Before your change I had to call address of leaf here. I left some vestiges of this behind. I can undo it if you think it's confusing.

erozenfeld:
The renaming change is fine; the commit/PR description that I quoted is confusing. It would be good to change it since you didn't fix any refanyval bugs in this change.

lib/Reader/readerir.cpp
```diff
@@ -6543,6 +6543,40 @@ IRNode *GenIR::loadAndBox(CORINFO_RESOLVED_TOKEN *ResolvedToken, IRNode *Addr,
...
+
+  // Store the type handle in the type field.
+  Value *TypeFieldAddress = LLVMBuilder->CreateStructGEP(RefAnyTy, RefAny, 1);
+  Value *TypeHandle = genericTokenToNode(ResolvedToken);
+  assert((TypeHandle->getType() == RefAnyStructTy->getContainedType(1)) &&
+         "wrong type for refany type");
+  makeStoreNonNull(TypeHandle, TypeFieldAddress, IsVolatile);
+
+  // Load the refany as the result.
+  Value *Result = makeLoadNonNull(RefAny, IsVolatile);
```
erozenfeld: This is loading a struct value, which is not allowed after my change. Are you not getting an assert when trying to push this to the stack?

erozenfeld: The correct way to return is
setValueRepresentsStruct(RefAny);
return RefAny;

lib/Reader/readerir.cpp
```diff
@@ -6543,6 +6543,40 @@ IRNode *GenIR::loadAndBox(CORINFO_RESOLVED_TOKEN *ResolvedToken, IRNode *Addr,
...
+  StructType *RefAnyStructTy = cast<StructType>(RefAnyTy);
+  Value *RefAny = createTemporary(RefAnyTy, "RefAny");
+  const bool IsVolatile = false;
+
+  // Store the object in the value field. Object should either be an object
+  // reference or the address of a local or param, both of which we type as
+  // managed pointers.
+  assert(isManagedPointerType(Object->getType()) &&
+         "wrong type for refany value");
+  Value *ObjectFieldAddress = LLVMBuilder->CreateStructGEP(RefAnyTy, RefAny, 0);
```
erozenfeld: I would make 0 and 1 symbolic constants here (e.g., ObjectFieldIndex and TypeFieldIndex)

lib/Reader/readerir.cpp
AndyAyersMS: There's nothing in `convertToStackType` to assert that structs are not allowed. Seems like we should have one there, right? On the popping side we should see it but I ran Roslyn with a rel w/asserts build and I'm guessing the ASSERTNR there went away.

erozenfeld: GenStack::push has 
ASSERT(GenIR::isValidStackType(NewVal));
It should fire in this case.

AndyAyersMS: The issue is that ASSERTNR doesn't fire in RelWithDebInfo builds. I have reworked this and am testing with full debug builds.... should have an update out shortly.

AndyAyersMS:
@erozenfeld updated per your feedback... Roslyn and my test case pass with debug builds.

erozenfeld:
LGTM.

Issue #592, opened 2015-05-21
Bad interaction between ldsflda and mkrefany


pgavlin:
ldsflda sometimes returns an unmanaged pointer, but mkref any expects managed pointers only. This causes Methodical/refany/format to fail with an assert.

erozenfeld:
@pgavlin This is likely the same as #588. I have an open PR #589 with a fix (waiting for a green CI).

pgavlin:
Could be. I'll rerun the test once #589 is merged.

AndyAyersMS:
Note that the code that asserts that  'mkrefany' expects a managed pointer is our convention. Refanys can only live on the stack and the value field must be reported as an interior pointer. So I've modeled it as a managed pointer so that when we got to generate the GC references on the frame we don't need to special case the refany type. However other things (local addresses and evidently static addresses) can be pointed to by the refany's value field.

I suspect our typing of `ldsflda` is suspect and it should be returning managed pointers more often.

AndyAyersMS:
On the latter point I think we should be creating a managed pointer for field addresses unless `CORINFO_FLG_FIELD_UNMANAGED` is set in the field's flags.

AndyAyersMS:
I'll take a look at this one.

AndyAyersMS:
For now, retyping any unmanaged pointer passed into the `mkrefany`.

PR #595, opened 2015-05-21
Add exclusions for new JIT tests.

pgavlin:
This covers the tests under JIT/Methodical. All but one of the exclusions are
due to issues involving EH or GC. The outlier is tracked by issue #592.


pgavlin:
@AndyAyersMS PTAL

AndyAyersMS:
LGTM.

pgavlin:
Test failure is spurious; it appears to be due to some packaging problems.

PR #597, opened 2015-05-22
Fix the type of refanyval.

pgavlin:
refanyval was returning an interior pointer to a byte instead of
an interior pointer to the type given as an operand to the
refanyval instruction. This was causing downstream code to assert.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -3211,9 +3211,17 @@ public:
 
   /// Create an operand that will be used to pass to the boxing helper
   ///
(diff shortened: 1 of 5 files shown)


pgavlin:
@AndyAyersMS @erozenfeld PTAL

AndyAyersMS:
LGTM.

pgavlin:

pgavlin:
The build failures are due to problems with BCLRewriter that we've been seeing.

PR #600, opened 2015-05-23
Implement refAnyType and fix llilc_checkpass.

pgavlin:
This cleans up an additional six NYIs in the new tests; the remaining
four are still under investigation. This also fixes llilc_checkpass for
compatibility with Python 3.x, as well as correcting an inverted
condition in success/failure diff measurement.


pgavlin:
@AndyAyersMS @erozenfeld PTAL

AndyAyersMS:
LGTM

pgavlin:
CI failures are due to problems running the BCL rewriter.

PR #606, opened 2015-05-29
Emit catch clauses

JosephTremoulet:
This is a PR for changes going into the EH branch.  They enable emission and execution of simple catch handlers (nested handlers and references to locals NYI) when built against LLVM with corresponding changes (Microsoft/llvm#43).  The approach to Windows-style EH support in LLVM is [undergoing a change](http://thread.gmane.org/gmane.comp.compilers.llvm.devel/85783) that will necessitate reworking at least the reader portion of these changes.  The EEMemoryManager changes, or something along these lines, will likely survive (but see [this llvmdev thread](http://thread.gmane.org/gmane.comp.compilers.llvm.devel/86316) for discussion of the approach).  I'm parking these changes in the EH branch so anyone curious can see where things are/were heading and provide feedback, and planning shift my immediate focus to aspects of EH that are independent of the impending rewrite (e.g. filter function outlining).

### DIFF
--- a/lib/Jit/EEMemoryManager.cpp
+++ b/lib/Jit/EEMemoryManager.cpp
@@ -26,6 +26,7 @@
 #include "llvm/IR/Verifier.h"
 #include "llvm/IRReader/IRReader.h"
 #include "llvm/IR/PassManager.h"
+#include "llvm/Object/ObjectFile.h"
 #include "llvm/Support/SourceMgr.h"
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Support/Timer.h"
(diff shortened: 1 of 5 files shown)


libengu:
retest this please

JosephTremoulet:
I've reworked the reserveUnwindInfo portion according to the switch to ORC jit.
@AndyAyersMS PTAL.

lib/Jit/EEMemoryManager.cpp
```diff
@@ -74,15 +75,65 @@ bool EEMemoryManager::finalizeMemory(std::string *ErrMsg) {
   return true;
 }
 
+static size_t getXdataSize(const uint8_t *PdataPtr) {
```
AndyAyersMS: Naming nit: `PdataPtr` seems unfortunate because it sounds like it is referring to the .pdata section. Since the .pdata points at the .xdata doing things this way is not all that implausible. But it's not pointing there.

Also given that this is a file static it needs doxygen markup here.

lib/Jit/EEMemoryManager.cpp
```diff
@@ -130,13 +181,40 @@ void EEMemoryManager::registerEHFrames(uint8_t *Addr, uint64_t LoadAddr,
...
-  // we have multiple unwind regions we'll need something more clever.
-  uint32_t StartOffset = 0;
-  uint32_t EndOffset = this->Context->HotCodeSize;
-  this->Context->JitInfo->allocUnwindInfo(
-      this->HotCodeBlock, nullptr, StartOffset, EndOffset, Size,
-      (BYTE *)LoadAddr, CorJitFuncKind::CORJIT_FUNC_ROOT);
+  uint32_t *DataPtr = reinterpret_cast<uint32_t *>(Addr);
+  CorJitFuncKind FuncKind = CorJitFuncKind::CORJIT_FUNC_ROOT;
+  uint32_t IsLast;
+  do {
```
AndyAyersMS: If we ever had chained unwind in the main root (eg from shrink wrap or separation) this would get confused, right? Is there any way for us to know if we see such cases here?

AndyAyersMS:
Overall this looks good but you might want to think about future-proofing if we're ever in a mode where there are more unwind records then there are functions/funclets.

lib/Jit/EEMemoryManager.cpp
JosephTremoulet: Oops, yeah, that was supposed to be XdataPtr.  Good catch, thanks; fixed.

JosephTremoulet:
Ok, I've updated Microsoft/llvm#43 to put the sentinel indicating last-funclet-ness before the funclet start/end offsets rather than after.  Neither of the sentinel values (0 and 0xffffffff) agree with what we'd expect to see if a chained unwind info were present (the low 3 bits would be 001), so that let me add some assertions to the parsing code that should fail if we ever see chained unwind.  I've pushed this up as a separate commit (7e22eb6), but intend to squash it before merging.
@AndyAyersMS PTAL.

AndyAyersMS:
LGTM.

PR #607, opened 2015-05-29
LLILC changes needed to merge latest LLVM.

richardlford:
LLVM now include a coff.h file that defines in enumeration types
constants with the same names that Windows defines using
in the scope of the windows #defines. To avoid that we
include related include files early before anyone has includes
windows.h.

Two other small changes made:

1. The TargetOptions.NoFramePointerElim field no longer exists.
For now I just commented out references to it, but we need to
decide if there is something else we need to do to
ensure we get a frame pointer.

2. The parameters of the LLVMBUILDER->CreateCall method changed.
Now need to pass args. We only used this for making NOPs, so
pass empty args.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -13,6 +13,7 @@
 ///
 //===----------------------------------------------------------------------===//
 
+#include "earlyincludes.h"
 #include "readerir.h"
 #include "imeta.h"
 #include "newvstate.h"
@@ -526,7 +527,7 @@ void GenIR::insertIRToKeepGenericContextAlive() {
 
   // This method now requires a frame pointer.
   TargetMachine *TM = JitContext->EE->getTargetMachine();
-  TM->Options.NoFramePointerElim = true;
+  // TM->Options.NoFramePointerElim = true;
 
   // TODO: we must convey the offset of this local to the runtime
   // via the GC encoding.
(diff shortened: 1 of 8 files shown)


lib/Reader/readerir.cpp
```diff
@@ -5308,7 +5309,8 @@ void GenIR::nop() {
     // We may need to pick something else that survives lowering.
     Value *DoNothing = Intrinsic::getDeclaration(JitContext->CurrentModule,
                                                  Intrinsic::donothing);
-    LLVMBuilder->CreateCall(DoNothing);
+    ArrayRef<Value *> Args;
```
pgavlin: I'm a little surprised that this works. I would have expected something like `Value *Args[] = {};`.

pgavlin:
> The TargetOptions.NoFramePointerElim field no longer exists.
For now I just commented out references to it, but we need to
decide if there is something else we need to do to
ensure we get a frame pointer.

Is there an issue tracking this?

lib/Reader/readerir.cpp
richardlford: ArrayRef has a default constructor that yields an empty list of arg.

richardlford:
I added issue #608 to track follow up on deletion of TargetOptions.NoFramePointerElim.

lib/Reader/readerir.cpp
pgavlin: OK, great.

pgavlin:
LGTM.

jasonwilliams200OK:
Side note: If you reference the issue number in your commit message (for example at the end \n\nIssue-URL: #608), it gives two advantages:

* Keeps the repository history well connected (with back reference to the original issue).
* When the PR is merged, it automatically closes the issue, and hence prevent you manually close the issue. 

Since it wasn't the case here, so #608 needs to be manually closed :)

PR #611, opened 2015-05-31
Fix filename and shorten length of first line of earlyincludes.h

richardlford:
- It erroneously still was named WindowsPal.h, the name
of an earlier attempt to fix the include compatibility problem.

- The header line was too long, causing the formatter to
break the line and thus the formatting job to fail.


richardlford:
I'm going to go ahead and merge this as it should fix a build break.

PR #613, opened 2015-06-01
Support {add,sub}.ovf.un when the result is a pointer.

pgavlin:
ECMA-335 supports addition and subtraction with overflow under the rules
given in table III.8. Because overflowing arithmetic in LLVM requires
calling an intrinsic that only accepts integral arguments, supporting
these operations requires converting the incoming operands to an
intermediate integral type, then converting the result to the ultimate
result type.

This also includes a small fix for errors encountered when compiling with
GCC 4.9.2 on Linux.

### DIFF
--- a/include/clr/misc.h
+++ b/include/clr/misc.h
@@ -21,6 +21,7 @@
 
 //-- Bit Manipulation --
 
+#ifndef _rotl
 inline unsigned int __cdecl _rotl(unsigned int value, int shift)
 {
   unsigned int retval = 0;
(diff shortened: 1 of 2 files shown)


pgavlin:
@AndyAyersMS @erozenfeld PTAL

pgavlin:
Jenkins is having some trouble updating PR statuses. The (green) build CI is here: http://dotnet-ci.cloudapp.net/job/dotnet_llilc_prtest/725/

erozenfeld:
LGTM.

PR #614, opened 2015-06-03
Switch to ORC jit infrastructure

### LINKED ISSUE #610: Switch from MCJit to ORC
The [ORC facilities](http://article.gmane.org/gmane.comp.compilers.llvm.devel/80639) in LLVM provide a newer framework for constructing JITs than the MCJit class that LLILC is currently built on, and are the focus of new JIT development.  Migrating LLILC to ORC therefore makes sense as our long-term plan.  It now makes sense in the near-term, too; EH support will require funclet generation, and the handling of multiple LLVM `Function`s and `MachineFunction`s per MSIL method will look different in ORC vs MCJit, so switching to ORC now will ensure we don't have to do that work twice.

JosephTremoulet:
This will simplify some aspects of funclet management, and also gets LLILC
on top of the newer jitting framework that active development is focused
on.

ORC jits are created by composing `layers`.  LLILC's functionality is
covered by two "off-the-shelf" layers: the IRCompileLayer (using the
SimpleComplier utility) for compiling IR to machine code, and the
ObjectLinkingLayer for resolving fixups and relocations and interfacing
with the EEMemoryManager to load the machine code into appropriate memory.

Closes #610

### DIFF
--- a/include/Jit/LLILCJit.h
+++ b/include/Jit/LLILCJit.h
@@ -22,6 +22,8 @@
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/ThreadLocal.h"
+#include "llvm/ExecutionEngine/Orc/ObjectLinkingLayer.h"
+#include "llvm/ExecutionEngine/Orc/IRCompileLayer.h"
 
 class ABIInfo;
 struct LLILCJitPerThreadState;
@@ -79,7 +81,7 @@ public:
   //@{
   llvm::LLVMContext *LLVMContext; ///< LLVM context for types and similar.
   llvm::Module *CurrentModule;    ///< Module holding LLVM IR.
-  llvm::ExecutionEngine *EE;      ///< MCJIT execution engine.
+  llvm::TargetMachine *TM;        ///< Target characteristics
   bool HasLoadedBitCode;          ///< Flag for side-loaded LLVM IR.
   //@}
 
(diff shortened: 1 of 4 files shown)


JosephTremoulet:
@russellhadley PTAL

include/Jit/LLILCJit.h
```diff
@@ -167,6 +186,10 @@ struct LLILCJitPerThreadState {
 /// top-level invocations of the jit is held in thread local storage.
 class LLILCJit : public ICorJitCompiler {
 public:
+  typedef llvm::orc::ObjectLinkingLayer<> LoadLayerT;
+  typedef llvm::orc::IRCompileLayer<LoadLayerT> CompileLayerT;
+  typedef CompileLayerT::ModuleSetHandleT ModuleHandleT;
```
pgavlin: Nit: I assume you chose these names for consistency with ORC, but would you object to removing the terminal `T`?

pgavlin:
LGTM aside from one nit.

russellhadley:
@JosephTremoulet well that looks like it was easier than I expected.  I'm assuming that there will be some more follow on work for the outlining correct?  

Code LGTM.  Thanks.

include/Jit/LLILCJit.h
JosephTremoulet: While I'm not a fan of terminal `T`s in general, I really think it's helpful to have them here, and yes that's because the ORC code uses them.  The ORC code makes heavy use of type parameterization, using the terminal `T` to signal both the type parameters to templates and the exported typedefs like these that make the type expressions more manageable (both by reducing their size and by abstracting away implementation details of lower layers, which are typically template arguments of higher layers).  Also I think in the case of typedefs for layers, there's a good chance that the name without the T would be the same as the template name; following convention here avoids any such conflicts (which didn't happen here, but may happen as we add layers, and maybe I should have used `ObjectLinkingLayerT` and `IRCompileLayerT`...).
Taking a closer look at this, though, made me realize that the ModuleHandleT I defined was vestigial and nowhere referenced, so I'll push up an update that omits it.

JosephTremoulet:
@russellhadley yes, I'll redo the provisional changes I have for the EH branch to do the obj cracking for xdata reporting on top of ORC rather than MCJit modifications.  I did a quick proof-of-concept of that before submitting this, and it works.  And yes I'd imagine we'll be adding layers over time (e.g. it may be a natural way to ensure that outlined filters are laid out just before their handlers), though I don't know concretely what.

JosephTremoulet:
[and yes, it was easier than I expected as well :) ]

JosephTremoulet:
@dotnet-bot re-test this please

JosephTremoulet:
CI failures match baselines; going ahead with merge.

PR #615, opened 2015-06-03
Initialize Scalar opts

swaroop-sridhar:
Call Core and Scalar opts initializers so that the
dependencies for Safepoint placement and rewriting 
are initialized correctly.


swaroop-sridhar:
@pgavlin  please review, Thanks.

pgavlin:
LGTM

Issue #616, opened 2015-06-03
Investigate Regression\CLR-x86-EJIT test case that has issue with LLILC


libengu:
Core CLR PR https://github.com/dotnet/coreclr/pull/1088 added multiple test directories under Regression. Three of the test cases failed if run with LLILC. Investigate the causes.

JIT\Regression\CLR-x86-EJIT\V1-M11-Beta1\b40138\b40138
JIT\Regression\CLR-x86-EJIT\V1-M11-Beta1\b45679\b45679
JIT\Regression\CLR-x86-EJIT\V1-M12-Beta2\b46847\b46847

AndyAyersMS:
Will reassess these.

AndyAyersMS:
```
JIT\Regression\CLR-x86-EJIT\V1-M11-Beta1\b40138\b40138.cmd -- EH
JIT\Regression\CLR-x86-EJIT\V1-M11-Beta1\b45679\b45679.cmd -- EH
JIT\Regression\CLR-x86-EJIT\V1-M12-Beta2\b46847\b46847.cmd -- EH
```

PR #617, opened 2015-06-03
Exclude Regression\CLR-x86-EJIT test case that has issue with LLILC

libengu:
Core CLR PR dotnet/coreclr#1088 added multiple test directories under Regression. Three of the test cases failed if run with LLILC. https://github.com/dotnet/llilc/issues/616 is created for investigation. Exclude the test case for now.


libengu:
@pgavlin PTAL

pgavlin:
LGTM.

libengu:
retest this please

libengu:
5c709f8 added two more test cases in exclusion.

Issue #380, opened 2015-03-31
Standardize assertions


JosephTremoulet:
LLVM uses `assert`, LLILC defines `ASSERT` and `ASSERTNR`, and it's worth checking what the CLR exposes and what the production JIT uses.  We should make a determination when to use which and apply it to the existing assertions in our codebase.

JosephTremoulet:
FYI:
LLVM's official guidance on assert is [here](http://llvm.org/docs/CodingStandards.html#assert-liberally); the recommendation is to use the assert macro (from \<cassert\>), which is [spec'ed](http://en.cppreference.com/w/cpp/error/assert) to write to stderr and call `std::abort`.  The [MS CRT implementation](https://msdn.microsoft.com/en-us/library/9sb57dw4.aspx) uses stderr or a message box, depending on the application type:

> The destination of the diagnostic message depends on the type of application that called the routine. Console applications always receive the message through stderr. In a Windows-based application, assert calls the Windows MessageBox function to create a message box to display the message along with an OK button. When the user clicks OK, the program aborts immediately.

And, indeed, the MessageBox popup is what we see when testing through corerun.exe on Windows.

I'm not seeing in any of that documentation a way to override where the diagnostic output goes, but maybe there's something clever we can do...

Still haven't checked on what the CLR exposes / what RyuJit uses.

JosephTremoulet:
I noticed today that there's no popup when Clang fails an assertion in LLVM.  The `main` function in the driver calls `llvm::sys::PrintStackTraceOnErrorSignal`, which in turn calls `DisableSystemDialogsOnCrash` and `RegisterHandler`.  Maybe we could use the same mechanism to redirect assertion failure output somewhere appropriate in LLILC.

I don't think that's the whole story -- if I comment out the call to `PrintStackTraceOnErrorSignal`, I start getting a popup on assertion failures, but it's not the same as the one I see when an assertion fails on LLILC (the Clang one seems to be responding to an abort, and the LLILC one seems to be responding directly to the assert).  But it's at least a place to start looking.

JosephTremoulet:
see also #584 regarding disabling assert popups

JosephTremoulet:
[...]

PR #618, opened 2015-06-03
Multiple small bug fixes.

pgavlin:
- Change the ASSERTM macro in the reader to use `assert` rather than
  `debugError` in order to avoid losing the message argument
- In GenIR::convertFromStackType, allow zero extension to native uint
- In GenIR::binaryOpType, handle subtraction of one managed pointer
  from another managed pointer as per ECMA-335 Table III.2
- In GenIR::mdArrayRefAddr, convert array indicies from stack types to
  actual types for consitency with ECMA-335 III.1.6
- In GenIR::arrayGetDimLength, handle cases where the array argument is
  an opaque System.Array reference
- In GenIR::cmp, allow comparison between floats and doubles by extending
  the former to the size of the latter

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -47,12 +47,7 @@
 // Debugging
 // -----------------------------------------------------------------
 
-#define ASSERTM(Predicate, Message)                                            \
-  do {                                                                         \
-    if (!(Predicate)) {                                                        \
-      ReaderBase::debugError(__FILE__, __LINE__, Message);                     \
-    }                                                                          \
-  } while (0)
+#define ASSERTM(Predicate, Message) assert((Predicate) && Message)
 #define ASSERT(Predicate) ASSERTM(Predicate, #Predicate)
 #define UNREACHED 0
 #ifndef _MSC_VER
@@ -2956,10 +2951,6 @@ public:
 
   // Remove all IRNodes from block (for verification error processing.)
   virtual void clearCurrentBlock() = 0;
-
-  // Called when an assert occurs (debug only)
-  static void debugError(const char *Filename, uint32_t LineNumber,
-                         const char *Message);
 
   // Notify client of alignment problem
   virtual void verifyStaticAlignment(void *Pointer, CorInfoType CorType,
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -49,8 +49,9 @@ void GenStack::push(IRNode *NewVal) {
 
 IRNode *GenStack::pop() {
   ASSERTM(size() > 0, "stack underflow");
-  if (size() == 0)
+  if (size() == 0) {
     LLILCJit::fatal(CORJIT_BADCODE);
+  }
 
   IRNode *result = Stack.back();
   Stack.pop_back();
@@ -1085,14 +1086,6 @@ void GenIR::verifyStaticAlignment(void *FieldAddress, CorInfoType CorType,
 #endif
 }
 
-void ReaderBase::debugError(const char *Filename, unsigned Linenumber,
-                            const char *S) {
-  assert(0);
-  // TODO
-  // if (s) JitContext->JitInfo->doAssert(Filename, Linenumber, S);
-  // ASSERTNR(UNREACHED);
-}
-
 // Fatal error, reader cannot continue.
 void ReaderBase::fatal(int ErrNum) { LLILCJit::fatal(LLILCJIT_FATAL_ERROR); }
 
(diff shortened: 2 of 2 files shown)


pgavlin:
@erozenfeld @JosephTremoulet PTAL

lib/Reader/readerir.cpp
```diff
@@ -4099,18 +4105,24 @@ bool GenIR::canExpandMDArrayRef(CORINFO_SIG_INFO *Sig, bool IsStore,
...
   if (IsStore) {
-    CORINFO_ARG_LIST_HANDLE ArgLst = Sig->args;
+    CORINFO_ARG_LIST_HANDLE ArgList = Sig->args;
+    CORINFO_CLASS_HANDLE ArgType;
+
     // Skip arguments for each dimension.
     for (uint32_t R = 0; R < *Rank; R++) {
-      ArgLst = getArgNext(ArgLst);
+      CorType = strip(getArgType(Sig, ArgList, &ArgType));
+      if (CorType != CorInfoType::CORINFO_TYPE_INT) {
```
erozenfeld: Should this be an assert? Do you see cases where we get a different type?

pgavlin: I don't see such cases, but this whole codepath is really just optimization. I think that falling back to the helper in corner cases is okay. If you feel differently, however--or if you'd like the assert simply to help find the corner cases--I can make this change.

erozenfeld: Yes, my preference is to assert if we don't believe the corner case is legal and don't see it anywhere.

pgavlin: I don't know that this case is illegal; I just haven't seen it.

erozenfeld: We would only see a different type if the signature of System.Array.SetValue(object, int32[]) changed. That won't happen but I guess I'm ok with either an assert or what you have.

erozenfeld:
LGTM.

include/Reader/reader.h
```diff
@@ -47,12 +47,7 @@
 // Debugging
 // -----------------------------------------------------------------
 
-#define ASSERTM(Predicate, Message)                                            \
-  do {                                                                         \
-    if (!(Predicate)) {                                                        \
-      ReaderBase::debugError(__FILE__, __LINE__, Message);                     \
```
JosephTremoulet: I believe this was the only caller of ReaderBase::debugError, so it could be removed now.

include/Reader/reader.h
```diff
@@ -47,12 +47,7 @@
...
-    }                                                                          \
-  } while (0)
+#define ASSERTM(Predicate, Message) assert((Predicate) && Message)
```
JosephTremoulet: We really might as well just go ahead and inline and remove this macro at this point...

lib/Reader/readerir.cpp
```diff
@@ -2069,9 +2069,10 @@ IRNode *GenIR::convertFromStackType(IRNode *Node, CorInfoType CorType,
       Result = (IRNode *)LLVMBuilder->CreateIntCast(Node, ResultTy, IsSigned);
     } else if (NeedsExtension) {
       assert(!NeedsReinterpret && "cannot reinterpret and extend");
-      assert((CorType == CorInfoType::CORINFO_TYPE_NATIVEINT) &&
-             "only expect to extend to native int");
-      const bool IsSigned = true;
+      assert((CorType == CorInfoType::CORINFO_TYPE_NATIVEINT ||
+              CorType == CorInfoType::CORINFO_TYPE_NATIVEUINT) &&
+             "only expect to extend to native int or uint");
+      const bool IsSigned = CorType == CorInfoType::CORINFO_TYPE_NATIVEINT;
```
JosephTremoulet: Do you know whether RyuJit zero-extends in these cases?

lib/Reader/readerir.cpp
```diff
@@ -3082,48 +3083,53 @@ Type *GenIR::binaryOpType(ReaderBaseNS::BinaryOpcode Opcode, Type *Type1,
...
+      if ((Size1 <= TargetPointerSizeInBits) &&
+          (Size2 <= TargetPointerSizeInBits)) {
+        return Type::getIntNTy(*JitContext->LLVMContext,
+                               TargetPointerSizeInBits);
+      }
+    } else if (isManagedPointerType(Type1)) {
+      if (IsSub && isManagedPointerType(Type2)) {
+        // The difference of two managed pointers is a native int.
+        return Type::getIntNTy(*JitContext->LLVMContext,
+                               TargetPointerSizeInBits);
```
JosephTremoulet: The caller is going to create `PtrToInt`s over the two sources and then do an integer subtract.  Moving one `PtrToInt` across a GC safepoint but not the other would be bad.  Does safepoint insertion run before any optimizations might try to do that?

JosephTremoulet:
Looks good to me, just a couple questions and some thoughts on asserts that could be done later / as #380.

libengu:
retest this please

include/Reader/reader.h
pgavlin: Done.

include/Reader/reader.h
pgavlin: I think that it's better to do this as part of #380, as it will cause cascading changes.

lib/Reader/readerir.cpp
pgavlin: That's a very good question. LLILC currently runs safepoint insertion as the very first pass, so the question really boils down to "are there any optimizations that might try to [move `PtrToInt`s]". Are there any particular optimizations that you're thinking of?

JosephTremoulet: No, just wasn't sure where we are with "late safepoint insertion" and how late "late" means.

include/Reader/reader.h
JosephTremoulet: Makes sense.

libengu:
Looks like there is formatting issue in your commit. I believe new test cases and exclusion should work as it was verified by CI in previous PR.

lib/Reader/readerir.cpp
pgavlin: RyuJIT sign-extends in both cases, but this appears to be a bug: ECMA-335 Table III.9 prescribes sign-extension in the case of native int and zero-extension in the case of native unsigned int.

lib/Reader/readerir.cpp
pgavlin: It turns out that indices that are not int32s would be illegal w.r.t. ECMA, so I changed this to an assert.

PR #621, opened 2015-06-06
Rework circular type resolution.

erozenfeld:
We had some code to deal with possible cycles in type graphs but it wasn't complete.

This commit has the following changes:

1. The mechanism for deferring and resolving types of aggregate fields is changed.
We now pass around a list that collects all aggregates whose fields weren't resolved.
These aggregates are resolved only after all fields of the top-level aggregate are resolved.

2. When resolving fields, in addition to deferring resolution of class types , we defer resolution
of both class types and struct types when the field is of a pointer to class or struct type.

3. We have the same rules for deferring resolution of array element types as we do for field types.

Examples that these changes fix:

================
class C {
  B b;
}

struct A {
  C c;
}

struct B {
  A a;
}
(Resolution of B was resulting in an assert).

=================
unsafe struct S
{
  S* ptr;
}
(Resolution of S was resulting in an assert).

=================
struct S {
  T[] array;
}

struct T {
  S s;
}

(Resolution of T was resulting in an assert).

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -857,11 +857,44 @@ public:
 #endif
 
 private:
-  llvm::Type *getType(CorInfoType Type, CORINFO_CLASS_HANDLE ClassHandle,
-                      bool GetRefClassFields = true);
+  /// \brief Get llvm type corresponding to Type and ClassHandle.
+  ///
+  /// \param Type  Type to get llvm type for.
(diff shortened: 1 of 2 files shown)


erozenfeld:
@AndyAyersMS PTAL

AndyAyersMS:
LGTM. Hopefully this is general enough now....

PR #623, opened 2015-06-08
Update EH doc to reflect new WinEH proposal

JosephTremoulet:
The plan is to build on the model proposed in
http://thread.gmane.org/gmane.comp.compilers.llvm.devel/85783 - in
particular, EH edges will no longer target `landingpad`, but rather will
target `catchblock`/`cleanupblock` instructions that can guarantee
preserving the structure of dispatch code (which ultimately will be
perfomed by the runtime) during optimization.


russellhadley:
LGTM

PR #647, opened 2015-06-12
Add Debug Line Information Support to LLILC

adiaaida:
This change adds support to LLILC to send source line information to
LLVM and then on to the CLR EE to be used in debugging. This change
adds:

* Support to the Reader to capture IL offsets and set them as debug
  locations for instructions. To do so, this change adds calls to the
  DIBuilder to create debug information for functions, and then to set
  the DebugLoc information for each instruction. The debugger expects
  offsets for the following: the start of a function, when the stack is
  empty, and when we have a call instruction. To get this all piped
  through LLVM properly, we had to insert extra nops for instances where
  the stack is empty, but we were looking at a store for a local (this
  gets swallowed by the following instruction). We also required extra
  nops around branches to get information about breaks piped through
  correctly. Finally, the debugger expects a location of where the
  return MSIL instruction was, but it expects the actual return logic to
  be considered part of the epilog. This is handled by the CoreCLR jit
  by inserting a nop with the IL offset of the return instruction, and
  everything after the nop has IL offset of -3 (epilog). Additionally,
  we had to change the logic for creating nops. The logic we previously
  had created a DoNothing intrinsic, which LLVM throws away. We needed
  these nops to make it all the way through to the AsmPrinter, so we
  changed nop creation to insert an inline asm nop, which LLVM preserves.
* Adds an object loaded layer to capture native and IL offsets and send
  them to the EE.

### DIFF
--- a/include/Jit/LLILCJit.h
+++ b/include/Jit/LLILCJit.h
@@ -19,13 +19,16 @@
 #include "Pal/LLILCPal.h"
 #include "options.h"
 #include "llvm/ExecutionEngine/ExecutionEngine.h"
+#include "llvm/ExecutionEngine/RuntimeDyld.h"
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/ThreadLocal.h"
 #include "llvm/ExecutionEngine/Orc/ObjectLinkingLayer.h"
 #include "llvm/ExecutionEngine/Orc/IRCompileLayer.h"
+#include "llvm/Config/config.h"
 
 class ABIInfo;
+class ObjectLoadListener;
 struct LLILCJitPerThreadState;
 
 /// \brief This struct holds per-jit request state.
@@ -186,7 +189,7 @@ public:
 /// top-level invocations of the jit is held in thread local storage.
 class LLILCJit : public ICorJitCompiler {
 public:
-  typedef llvm::orc::ObjectLinkingLayer<> LoadLayerT;
+  typedef llvm::orc::ObjectLinkingLayer<ObjectLoadListener> LoadLayerT;
   typedef llvm::orc::IRCompileLayer<LoadLayerT> CompileLayerT;
 
   /// \brief Construct a new jit instance.
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -25,6 +25,8 @@
 #include "llvm/IR/Verifier.h"
 #include "llvm/IR/CFG.h"
 #include "llvm/IR/CallSite.h"
+#include "llvm/IR/DIBuilder.h"
+#include "llvm/IR/DebugInfoMetadata.h"
 #include "reader.h"
 #include "abi.h"
 #include "abisignature.h"
@@ -558,6 +560,10 @@ public:
 
   EHRegion *rgnAllocateRegion() override;
   EHRegionList *rgnAllocateRegionList() override;
+  void setDebugLocation(uint32_t CurrOffset, bool IsCall) override;
+  llvm::DISubroutineType *createFunctionType(llvm::Function *F,
+                                             llvm::DIFile *Unit);
+  llvm::DIType *convertType(llvm::Type *Ty);
 
   //
   // REQUIRED Flow and Region Graph Manipulation Routines
@@ -1555,6 +1561,7 @@ private:
   // where they should be inserted (the gen- methods do not take explicit
   // insertion point parameters).
   llvm::IRBuilder<> *LLVMBuilder;
+  llvm::DIBuilder *DBuilder;
   std::map<CORINFO_CLASS_HANDLE, llvm::Type *> *ClassTypeMap;
   std::map<llvm::Type *, CORINFO_CLASS_HANDLE> *ReverseClassTypeMap;
   std::map<CORINFO_CLASS_HANDLE, llvm::Type *> *BoxedTypeMap;
@@ -1610,6 +1617,10 @@ private:
                                                 ///< helper function.
                                                 ///< This constant is from the
                                                 ///< legacy jit.
+  struct DebugInfo {
+    llvm::DICompileUnit *TheCU;
+    llvm::DIScope *FunctionScope;
+  } LLILCDebugInfo;
 };
 
 #endif // MSIL_READER_IR_H
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -22,30 +22,67 @@
 #include "abi.h"
 #include "EEMemoryManager.h"
 #include "llvm/CodeGen/GCs.h"
+#include "llvm/Config/llvm-config.h"
+#include "llvm/DebugInfo/DIContext.h"
+#include "llvm/DebugInfo/DWARF/DWARFContext.h"
+#include "llvm/ExecutionEngine/Orc/CompileUtils.h"
 #include "llvm/IR/DataLayout.h"
+#include "llvm/Support/DataTypes.h"
+#include "llvm/IR/DebugInfo.h"
+#include "llvm/IR/DebugLoc.h"
 #include "llvm/IR/DerivedTypes.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IRReader/IRReader.h"
 #include "llvm/IR/LegacyPassManager.h"
 #include "llvm/IR/LLVMContext.h"
+#include "llvm/IR/Metadata.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Verifier.h"
 #include "llvm/IR/PassManager.h"
+#include "llvm/Object/ObjectFile.h"
 #include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/Errno.h"
+#include "llvm/Support/Format.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/Signals.h"
 #include "llvm/Support/SourceMgr.h"
+#include "llvm/Support/TargetRegistry.h"
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Support/Timer.h"
-#include "llvm/Support/Debug.h"
-#include "llvm/Support/Format.h"
-#include "llvm/Support/Signals.h"
-#include "llvm/Support/CommandLine.h"
-#include "llvm/Support/TargetRegistry.h"
 #include "llvm/Transforms/Scalar.h"
 #include "llvm/Transforms/IPO/PassManagerBuilder.h"
-#include "llvm/ExecutionEngine/Orc/CompileUtils.h"
 #include <string>
 
 using namespace llvm;
+using namespace llvm::object;
+
+class ObjectLoadListener {
+public:
+  ObjectLoadListener(LLILCJitContext *Context) { this->Context = Context; }
+
+  template <typename ObjSetT, typename LoadResult>
+  void operator()(llvm::orc::ObjectLinkingLayerBase::ObjSetHandleT ObjHandles,
+                  const ObjSetT &Objs, const LoadResult &LoadedObjInfos) {
+    int I = 0;
+
+    for (const auto &PObj : Objs) {
+
+      const object::ObjectFile &Obj = *PObj;
+      const RuntimeDyld::LoadedObjectInfo &L = *LoadedObjInfos[I];
+
+      getDebugInfoForObject(Obj, L);
+
+      ++I;
+    }
+  }
+
+  void getDebugInfoForObject(const ObjectFile &Obj,
+                             const RuntimeDyld::LoadedObjectInfo &L);
+
+private:
+  LLILCJitContext *Context;
+};
 
 // The one and only Jit Object.
 LLILCJit *LLILCJit::TheJit = nullptr;
@@ -199,7 +236,8 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
 
   // Construct the jitting layers.
   EEMemoryManager MM(&Context);
-  LoadLayerT Loader;
+  ObjectLoadListener Listener(&Context);
+  LoadLayerT Loader(Listener);
   CompileLayerT Compiler(Loader, orc::SimpleCompiler(*TM));
 
   // Now jit the method.
@@ -393,4 +431,98 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
 //  this handler avoids a crash when LLVM goes down.
 void LLILCJit::signalHandler(void *Cookie) {
   // do nothing
+}
+
+void ObjectLoadListener::getDebugInfoForObject(
+    const ObjectFile &Obj, const RuntimeDyld::LoadedObjectInfo &L) {
+  OwningBinary<ObjectFile> DebugObjOwner = L.getObjectForDebug(Obj);
+  const ObjectFile &DebugObj = *DebugObjOwner.getBinary();
+
+  // Get the address of the object image for use as a unique identifier
+  const void *ObjData = DebugObj.getData().data();
+
+  // TODO: This extracts DWARF information from the object file, but we will
+  // want to also be able to eventually extract WinCodeView information as well
+  DWARFContextInMemory DwarfContext(DebugObj);
+
+  // Use symbol info to iterate functions in the object.
+  // TODO: This may have to change when we have funclets
+
+  for (symbol_iterator SI = DebugObj.symbol_begin(), E = DebugObj.symbol_end();
+       SI != E; ++SI) {
+    SymbolRef::Type SymType;
+    if (SI->getType(SymType))
+      continue;
+    if (SymType != SymbolRef::ST_Function)
+      continue;
+
+    // Function info
+    StringRef Name;
+    uint64_t Addr;
+    if (SI->getName(Name))
+      continue;
+    if (SI->getAddress(Addr))
+      continue;
+    uint64_t Size = SI->getSize();
+
+    unsigned LastDebugOffset = -1;
+    unsigned NumDebugRanges = 0;
+    ICorDebugInfo::OffsetMapping *OM;
+
+    DILineInfoTable Lines = DwarfContext.getLineInfoForAddressRange(Addr, Size);
+
+    DILineInfoTable::iterator Begin = Lines.begin();
+    DILineInfoTable::iterator End = Lines.end();
+
+    // Count offset entries. Will skip an entry if the current IL offset
+    // matches the previous offset.
+    for (DILineInfoTable::iterator It = Begin; It != End; ++It) {
+      int LineNumber = (It->second).Line;
+
+      if (LineNumber != LastDebugOffset) {
+        NumDebugRanges++;
+        LastDebugOffset = LineNumber;
+      }
+    }
+
+    // Reset offset
+    LastDebugOffset = -1;
+
+    if (NumDebugRanges > 0) {
+      // Allocate OffsetMapping array
+      unsigned SizeOfArray =
+          (NumDebugRanges) * sizeof(ICorDebugInfo::OffsetMapping);
+      OM = (ICorDebugInfo::OffsetMapping *)Context->JitInfo->allocateArray(
+          SizeOfArray);
+
+      unsigned CurrentDebugEntry = 0;
+
+      // Iterate through the debug entries and save IL offset, native
+      // offset, and source reason
+      for (DILineInfoTable::iterator It = Begin; It != End; ++It) {
+        int Offset = It->first;
+        int LineNumber = (It->second).Line;
+
+        // We store info about if the instruction is being recorded because
+        // it is a call in the column field
+        bool IsCall = (It->second).Column == 1;
+
+        if (LineNumber != LastDebugOffset) {
+          LastDebugOffset = LineNumber;
+          OM[CurrentDebugEntry].nativeOffset = Offset;
+          OM[CurrentDebugEntry].ilOffset = LineNumber;
+          OM[CurrentDebugEntry].source = IsCall
+                                             ? ICorDebugInfo::CALL_INSTRUCTION
+                                             : ICorDebugInfo::STACK_EMPTY;
+          CurrentDebugEntry++;
+        }
+      }
+
+      // Send array of OffsetMappings to CLR EE
+      CORINFO_METHOD_INFO *MethodInfo = Context->MethodInfo;
+      CORINFO_METHOD_HANDLE MethodHandle = MethodInfo->ftn;
+
+      Context->JitInfo->setBoundaries(MethodHandle, NumDebugRanges, OM);
+    }
+  }
 }
--- a/lib/Reader/reader.cpp
+++ b/lib/Reader/reader.cpp
@@ -1,4 +1,4 @@
-//===---- lib/MSILReader/reader.cpp -----------------------------*- C++ -*-===//
+//===---- lib/Reader/reader.cpp ---------------------------------*- C++ -*-===//
 //
 // LLILC
 //
@@ -402,7 +402,7 @@ void ReaderBase::printMSIL(uint8_t *Buf, uint32_t StartOffset,
   while (Offset < NumBytes) {
     dbPrint("0x%-4x: ", StartOffset + Offset);
     Offset = parseILOpcode(Buf, Offset, NumBytes, this, &Opcode, &Operand);
-    dbPrint("%-10s ", OpcodeName[Opcode]);
+    dbPrint("%d: %-10s ", Offset, OpcodeName[Opcode]);
 
     switch (Opcode) {
     default:
@@ -6410,6 +6410,12 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
     CurrInstrOffset = CurrentOffset;
     NextInstrOffset = NextOffset;
 
+    // Set debug locations
+    if (ReaderOperandStack->empty()) {
+      const bool IsCall = false;
+      setDebugLocation(CurrentOffset, IsCall);
+    }
+
     // If we have cached a LoadFtnToken from LDFTN or LDVIRTFTN
     // then clear it if the next opcode is not NEWOBJ
     if (Opcode != ReaderBaseNS::CEE_NEWOBJ) {
@@ -6564,6 +6570,8 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
       goto GEN_BRANCH;
 
     GEN_BRANCH:
+      // We need to record the source location of branches, so insert a nop
+      nop();
       Param->HasFallThrough = false;
       Param->VerifiedEndBlock = true;
       verifyFinishBlock(TheVerificationState,
@@ -6588,6 +6596,8 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
     case ReaderBaseNS::CEE_CALL:
     case ReaderBaseNS::CEE_CALLI:
     case ReaderBaseNS::CEE_CALLVIRT: {
+      const bool IsCall = true;
+      setDebugLocation(CurrentOffset, IsCall);
       bool IsUnmarkedTailCall = false;
       Token = readValue<mdToken>(Operand);
 
@@ -6994,6 +7004,9 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
     case ReaderBaseNS::CEE_LDC_I4_8:
     case ReaderBaseNS::CEE_LDC_I4_M1:
     LOAD_CONSTANT:
+      // We need to record the source locations of ldcs, so insert a nop
+      // to record the source location.
+      nop();
       verifyLoadConstant(TheVerificationState, Opcode);
       BREAK_ON_VERIFY_ONLY;
 
@@ -7342,6 +7355,11 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
       break;
 
     case ReaderBaseNS::CEE_RET: {
+      // Insert a nop for debug and set return's IL offset to signify
+      // that it is part of the epilogue
+      nop();
+      const bool IsCall = false;
+      setDebugLocation(ICorDebugInfo::EPILOG, IsCall);
       CorInfoCallConv Conv;
       CorInfoType CorType;
       CORINFO_CLASS_HANDLE RetTypeClass;
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -1,4 +1,4 @@
-//===---- lib/MSILReader/readerir.cpp ---------------------------*- C++ -*-===//
+//===---- lib/Reader/readerir.cpp -------------------------------*- C++ -*-===//
 //
 // LLILC
 //
@@ -18,11 +18,15 @@
 #include "imeta.h"
 #include "newvstate.h"
 #include "llvm/ADT/Triple.h"
+#include "llvm/IR/DebugLoc.h"
+#include "llvm/IR/InlineAsm.h"
 #include "llvm/IR/Intrinsics.h"
 #include "llvm/Support/Debug.h"          // for dbgs()
 #include "llvm/Support/raw_ostream.h"    // for errs()
 #include "llvm/Support/ConvertUTF.h"     // for ConvertUTF16toUTF8
 #include "llvm/Transforms/Utils/Local.h" // for removeUnreachableBlocks
+#include "llvm/IR/DebugInfo.h"
+#include "llvm/IR/DebugInfoMetadata.h"
 #include <cstdlib>
 #include <new>
 
@@ -312,6 +316,11 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
   EntryBlock = BasicBlock::Create(LLVMContext, "entry", Function);
 
   LLVMBuilder = new IRBuilder<>(LLVMContext);
+  DBuilder = new DIBuilder(*JitContext->CurrentModule);
+  LLILCDebugInfo.TheCU = DBuilder->createCompileUnit(dwarf::DW_LANG_C_plus_plus,
+                                                     Function->getName().str(),
+                                                     ".", "LLILCJit", 0, "", 0);
+
   LLVMBuilder->SetInsertPoint(EntryBlock);
 
   // Note numArgs may exceed the IL argument count when there
@@ -428,6 +437,21 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
     }
   }
 
+  // Setup function for emiting debug locations
+  DIFile *Unit = DBuilder->createFile(LLILCDebugInfo.TheCU->getFilename(),
+                                      LLILCDebugInfo.TheCU->getDirectory());
+  bool IsOptimized = (JitContext->Flags & CORJIT_FLG_DEBUG_CODE) == 0;
+  DIScope *FContext = Unit;
+  unsigned LineNo = 0;
+  unsigned ScopeLine = ICorDebugInfo::PROLOG;
+  bool IsDefinition = true;
+  DISubprogram *SP = DBuilder->createFunction(
+      FContext, Function->getName(), StringRef(), Unit, LineNo,
+      createFunctionType(Function, Unit), Function->hasInternalLinkage(),
+      IsDefinition, ScopeLine, DINode::FlagPrototyped, IsOptimized, Function);
+
+  LLILCDebugInfo.FunctionScope = SP;
+
   // TODO: Insert class initialization check if necessary
   CorInfoInitClassResult InitResult =
       initClass(nullptr, getCurrentMethodHandle(), getCurrentContext());
@@ -484,6 +508,7 @@ void GenIR::readerPostPass(bool IsImportOnly) {
   }
 
   // Cleanup the memory we've been using.
+  DBuilder->finalize();
   delete LLVMBuilder;
 }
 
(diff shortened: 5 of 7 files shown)


adiaaida:
@AndyAyersMS @pgavlin @russellhadley PTAL

include/Reader/readerir.h
```diff
@@ -1610,6 +1619,11 @@ class GenIR : public ReaderBase {
                                                 ///< helper function.
                                                 ///< This constant is from the
                                                 ///< legacy jit.
+  struct DbgInfo {
```
pgavlin: Nit: s/Dbg/Debug/ here and in `LLILCDbgInfo`

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+              LastDebugOffset = LineNumber;
+            }
+          }
+
+          // Reset offset
+          LastDebugOffset = -1;
+
+          if (NumDebugRanges > 0) {
+
+            // Allocate OffsetMapping array
```
pgavlin: Why not just allocate this lazily to avoid looping over the debug info twice?

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
 void LLILCJit::signalHandler(void *Cookie) {
   // do nothing
 }
+
+template <typename ObjSetT, typename LoadResult>
+void ObjectLoadListener::operator()(llvm::orc::ObjectLinkingLayerBase::ObjSetHandleT ObjHandles,
```
pgavlin: The `llvm::` qualification here is unnecessary,

include/Reader/readerir.h
```diff
@@ -558,6 +560,12 @@ class GenIR : public ReaderBase {
 
   EHRegion *rgnAllocateRegion() override;
   EHRegionList *rgnAllocateRegionList() override;
+  void setDebugLocation(uint32_t CurrOffset, uint32_t IsCall) override;
```
pgavlin: Please change `IsCall` to a `bool`.

lib/Jit/LLILCJit.cpp
adiaaida: How? OffsetMapping has to be allocated before we can start writing to it (and this is how it was in Ryujit when I did it originally).

pgavlin: Yeah, disregard that comment--I misread the code below.

lib/Reader/reader.cpp
```diff
@@ -6410,6 +6410,11 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
     CurrInstrOffset = CurrentOffset;
     NextInstrOffset = NextOffset;
 
+    // Set debug locations
+    if (ReaderOperandStack->empty()) {
+      setDebugLocation(CurrentOffset, 0);
```
AndyAyersMS: Agree with Pat that that second arg needs to convey information at call sites. If you make it a bool then change this to 
```
const bool IsCall = false;
setDebugLocation(CurentOffset, IsCall);
```
or similar.

lib/Reader/reader.cpp
```diff
@@ -8015,6 +8027,8 @@ void ReaderBase::msilToIR(void) {
     ++it;
   }
 
+  popDebugScope();
```
AndyAyersMS: It seems odd to me that `popDebugScope()` is here in the base class but `push` is only in the derived class. I would have expected them both to be at the same level...

lib/Reader/readerir.cpp
```diff
@@ -5381,11 +5519,15 @@ void GenIR::nop() {
   // Preserve Nops in debug builds since they may carry unique source positions.
   if ((JitContext->Flags & CORJIT_FLG_DEBUG_CODE) != 0) {
     // LLVM has no high-level NOP instruction. Put in a placeholder for now.
-    // We may need to pick something else that survives lowering.
-    Value *DoNothing = Intrinsic::getDeclaration(JitContext->CurrentModule,
-                                                 Intrinsic::donothing);
+    // This will survive lowering, but we may want to do something else that
+    // is cleaner
+    llvm::FunctionType *FTy = llvm::FunctionType::get(
+        llvm::Type::getVoidTy(*(JitContext->LLVMContext)), /*Variadic=*/false);
```
pgavlin: The preferred style is to avoid the inline comment and use a const local, e.g.

```
const bool Variadic = false;
llvm::FunctionType *FTy = llvm::FunctionType::get(
        llvm::Type::getVoidTy(*(JitContext->LLVMContext)), Variadic);
```

lib/Reader/readerir.cpp
```diff
@@ -428,6 +437,22 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
     }
   }
 
+  // Setup function for emiting debug locations
+  DIFile *Unit = DBuilder->createFile(LLILCDbgInfo.TheCU->getFilename(),
+                                      LLILCDbgInfo.TheCU->getDirectory());
+  bool isOptimized = (JitContext->Flags & CORJIT_FLG_DEBUG_CODE) == 0;
```
pgavlin: s/isOptimized/IsOptimized/

lib/Reader/readerir.cpp
```diff
@@ -428,6 +437,22 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
...
+  DIScope *FContext = Unit;
+  unsigned LineNo = 0;
+  unsigned ScopeLine = ICorDebugInfo::PROLOG;
+  DISubprogram *SP = DBuilder->createFunction(
+      FContext, Function->getName(), StringRef(), Unit, LineNo,
+      createFunctionType(Function, Unit), Function->hasInternalLinkage(),
+      true /* definition */, ScopeLine, DINode::FlagPrototyped, isOptimized,
```
pgavlin: Same issue with the inline comment here as below.

lib/Reader/reader.cpp
adiaaida: This is being deleted as Pat pointed out that we never have more than one scope, so there's no point in having the stack.

lib/Reader/reader.cpp
```diff
@@ -6994,6 +7001,7 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
     case ReaderBaseNS::CEE_LDC_I4_8:
     case ReaderBaseNS::CEE_LDC_I4_M1:
     LOAD_CONSTANT:
+      nop();
```
AndyAyersMS: I'm confused why we need a `nop` here.

lib/Reader/readerir.cpp
```diff
@@ -945,6 +971,118 @@ void GenIR::createSafepointPoll() {
...
+  // for each param, etc. convert the type to DIType and add it to the array.
+  FunctionType *FunctionTy = F->getFunctionType();
+
+  unsigned i = 0;
+
+  DIType *ReturnTy = convertType(Function->getReturnType());
+  EltTys.push_back(ReturnTy);
+
+  for (auto ParamIterator = FunctionTy->param_begin();
+       ParamIterator != FunctionTy->param_end(); ParamIterator++) {
```
pgavlin: Nit: lift the call to `param_end()` out of the loop or just use `for (Type *Ty : FunctionTy->params())`

lib/Reader/readerir.cpp
```diff
@@ -945,6 +971,118 @@ void GenIR::createSafepointPoll() {
...
+  LLVMBuilder->SetCurrentDebugLocation(Loc);
+}
+
+llvm::DISubroutineType *GenIR::createFunctionType(llvm::Function *F,
+                                                  DIFile *Unit) {
+  SmallVector<Metadata *, 8> EltTys;
+  // for each param, etc. convert the type to DIType and add it to the array.
+  FunctionType *FunctionTy = F->getFunctionType();
+
+  unsigned i = 0;
```
pgavlin: I think that this is unused aside from the increment in the loop below.

lib/Reader/readerir.cpp
```diff
@@ -945,6 +971,118 @@ void GenIR::createSafepointPoll() {
...
+    llvm::DIType *DbgTy =
+        DBuilder->createBasicType(TyName, BitWidth, BitWidth, Encoding);
+    return DbgTy;
+  }
+
+  if (Ty->isPointerTy()) {
+    uint64_t Size = JitContext->TM->getDataLayout()->getPointerTypeSize(Ty);
+    uint64_t Align = JitContext->TM->getDataLayout()->getPrefTypeAlignment(Ty);
+    llvm::DIType *DbgTy = DBuilder->createPointerType(
+        convertType(Ty->getContainedType(0)), Size, Align);
```
pgavlin: Please replace `getContainedType(0)` with `getPointerElementType()`.

lib/Reader/reader.cpp
```diff
@@ -6564,6 +6569,7 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
       goto GEN_BRANCH;
 
     GEN_BRANCH:
+      nop();
```
AndyAyersMS: Would love to see a comment here as to why we need a `nop`, even if it will just be there in a debug compile. I think I know why...maybe...

lib/Reader/readerir.cpp
```diff
@@ -945,6 +971,118 @@ void GenIR::createSafepointPoll() {
...
+
+    return DbgTy;
+  }
+  // TODO: add support for aggregate types
+  return nullptr;
```
pgavlin: Is it OK to just return `nullptr` from this function? Do we understand the impact of doing so? I don't think that we should see FCAs in our function signatures at the moment in any case, so it might be better to just turn this into an `assert(UNREACHED)` or `llvm_unreachable()`.

include/Reader/readerir.h
AndyAyersMS: Or maybe a enum with `IsCall` and `IsStackEmpty` the two possible values? Then the uses are self-documenting.

lib/Reader/reader.cpp
pgavlin: FWIW, it might be nice to change the name of this function to `debug_nop` or similar to indicate that nothing is emitted unless we're compiling for debug.

lib/Reader/reader.cpp
adiaaida: This is to make sure that the store of the constant gets an ILOffset. What I was seeing is that when we have a LDC followed by a call, the call's ILOffset would be used instead of the load's. The EE expects both the call's offset (with the CALL_INSTRUCTION sourcetype) and the load's offset because the stack is empty. Generally speaking, LDC's are at locations where the stack is empty, and the debugger expects il offsets for all points where the stack is empty. I could add a comment to make that clear.

include/Reader/readerir.h
pgavlin: Andy's suggestion is better: I hadn't really realized that `IsCall == 0` meant that the call originated from an empty operand stack.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+    const RuntimeDyld::LoadedObjectInfo &L = *LoadedObjInfos[I];
+
+    OwningBinary<ObjectFile> DebugObjOwner = L.getObjectForDebug(Obj);
+    const ObjectFile &DebugObj = *DebugObjOwner.getBinary();
+
+    // Get the address of the object image for use as a unique identifier
+    const void *ObjData = DebugObj.getData().data();
+    DIContext *DwarfContext = new DWARFContextInMemory(DebugObj);
+
+    // Use symbol info to iterate functions in the object.
```
AndyAyersMS: Any idea how this will need to change when we have funclets running around? If not, please leave a TODO and open an issue for debug with EH.

lib/Reader/readerir.cpp
adiaaida: Yep, this was an artifact of some of my debugging. Removing.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+        uint64_t Size = SI->getSize();
+
+        unsigned LastDebugOffset = -1;
+        unsigned CurrentDebugEntry = 0;
+        unsigned NumDebugRanges = 0;
+        unsigned Offset = 0;
+        ICorDebugInfo::OffsetMapping *OM;
+        // unsigned LastDebugEntry = 0;
+
+        if (DwarfContext) {
```
pgavlin: Can this condition ever be false? It looks like `DwarfContext` is only ever set once.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+  for (const std::unique_ptr<object::ObjectFile> &PObj : Objs) {
+    const object::ObjectFile &Obj = *PObj;
+    const RuntimeDyld::LoadedObjectInfo &L = *LoadedObjInfos[I];
+
+    OwningBinary<ObjectFile> DebugObjOwner = L.getObjectForDebug(Obj);
+    const ObjectFile &DebugObj = *DebugObjOwner.getBinary();
+
+    // Get the address of the object image for use as a unique identifier
+    const void *ObjData = DebugObj.getData().data();
+    DIContext *DwarfContext = new DWARFContextInMemory(DebugObj);
```
pgavlin: Can this be stack-allocated? `DWARFContextInMemory DwarfContext(DebugObj);`

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
 void LLILCJit::signalHandler(void *Cookie) {
   // do nothing
 }
+
+template <typename ObjSetT, typename LoadResult>
+void ObjectLoadListener::operator()(llvm::orc::ObjectLinkingLayerBase::ObjSetHandleT ObjHandles,
+                const ObjSetT &Objs, const LoadResult &LoadedObjInfos) {
+  int I = 0;
```
pgavlin: I think that this `I` is also unused except for a later increment.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+        if (SI->getAddress(Addr))
+          continue;
+        uint64_t Size = SI->getSize();
+
+        unsigned LastDebugOffset = -1;
+        unsigned CurrentDebugEntry = 0;
+        unsigned NumDebugRanges = 0;
+        unsigned Offset = 0;
+        ICorDebugInfo::OffsetMapping *OM;
+        // unsigned LastDebugEntry = 0;
```
pgavlin: Remove this line.

lib/Reader/readerir.cpp
```diff
@@ -945,6 +971,118 @@ void GenIR::createSafepointPoll() {
...
+  if (LLILCDbgInfo.LexicalBlocks.empty())
+    Scope = LLILCDbgInfo.TheCU;
+  else
+    Scope = LLILCDbgInfo.LexicalBlocks.back();
+  DebugLoc Loc = DebugLoc::get(CurrOffset, IsCall, Scope);
+
+  LLVMBuilder->SetCurrentDebugLocation(Loc);
+}
+
+llvm::DISubroutineType *GenIR::createFunctionType(llvm::Function *F,
```
AndyAyersMS: We should leave a note somewhere about these dwarf type mappings.

The CLR/EE will never look at them, so (I assume) we have them because if we run on linux and use native debugging then gdb or similar is informed about the debug for the jitted code we can get some information when debugging without using the "standard" xplat approach -- though it is debatable how useful this will be is given that there is no ability to map back to source.

I'm fine with keeping this code around but we should add a note about its purpose and why it is ok for it to be incomplete.

lib/Reader/reader.cpp
adiaaida: This was another instance of needed a nop to make the debugger happy, in particular with breaks. This was caught by one of the automated debugger tests that had breaks in it. Basically this nop maps to the il offset of where the curly brace source line is.

lib/Reader/readerir.cpp
```diff
@@ -5381,11 +5519,15 @@ void GenIR::nop() {
   // Preserve Nops in debug builds since they may carry unique source positions.
   if ((JitContext->Flags & CORJIT_FLG_DEBUG_CODE) != 0) {
     // LLVM has no high-level NOP instruction. Put in a placeholder for now.
-    // We may need to pick something else that survives lowering.
-    Value *DoNothing = Intrinsic::getDeclaration(JitContext->CurrentModule,
-                                                 Intrinsic::donothing);
+    // This will survive lowering, but we may want to do something else that
+    // is cleaner
```
AndyAyersMS: I think we'll eventually want a high-level NOP that doesn't get removed and gets lowered to the right platform-specific encoding, so perhaps you can add a todo and an ISSUE to implement that in LLVM.

lib/Jit/LLILCJit.cpp
adiaaida: Which line?

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+        StringRef Name;
+        uint64_t Addr;
+        if (SI->getName(Name))
+          continue;
+        if (SI->getAddress(Addr))
+          continue;
+        uint64_t Size = SI->getSize();
+
+        unsigned LastDebugOffset = -1;
+        unsigned CurrentDebugEntry = 0;
```
pgavlin: Nit: I'd move this declaration closer to its use.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+        unsigned NumDebugRanges = 0;
+        unsigned Offset = 0;
```
pgavlin: I think that this local is unused as well (it ends up shadowed by the declaration of `int Offset` on line 471.

lib/Jit/LLILCJit.cpp
pgavlin: The commented out declaration.

lib/Jit/LLILCJit.cpp
adiaaida: I don't know. I got this from the IntelJITEventListener, and this is how they have it. I can try removing it and seeing if anything breaks.

lib/Jit/LLILCJit.cpp
adiaaida: got it.

include/Reader/readerir.h
adiaaida: I can do IsStackEmpty, but we only set debug locations if StackEmpty is true or if it's a call instruction. So I thought the one bool would cover it.

lib/Reader/reader.cpp
adiaaida: Done.

lib/Reader/readerir.cpp
adiaaida: done.

lib/Reader/readerir.cpp
adiaaida: done.

lib/Reader/readerir.cpp
adiaaida: done.

lib/Reader/readerir.cpp
adiaaida: done.

lib/Reader/readerir.cpp
adiaaida: done.

lib/Jit/LLILCJit.cpp
adiaaida: I don't know. I'll add a TODO.

lib/Jit/LLILCJit.cpp
adiaaida: Done.

lib/Jit/LLILCJit.cpp
adiaaida: Done.

lib/Reader/readerir.cpp
adiaaida: We have it because I needed something for the functiontype, and I wasn't sure if/how this was used throughout LLVM. I don't think it is necessary for us. I'm also not sure it's necessary for GDB for JIT (looking at the GDB EventListener)

lib/Reader/reader.cpp
AndyAyersMS: Ah, I think I see what you're getting at ...certain MSIL operations generate operands, not instructions. So if we have an MSIL operation that only creates an operand, and it has a reportable debug location, and the following MSIL operation creates an instruction and has some other reportable debug location, that first debug location doesn't get recorded anywhere and is lost.

So arguably we only need this `nop` if we're at a stack empty point, right? And presumably `ldc` is the only case where we generate an operand and not an instruction?

You might be able to make this more robust in the main IL processing loop -- if you are at a stack empty, down at the bottom of the loop after we've generated the IR, if the last instruction in the current block doesn't have the stack empty debug info, you could generate a `nop` there. But if `ldc` is truly the only case of this then adding a comment here (and possibly only adding a `nop`\ for stack empty) is ok by me.

lib/Reader/readerir.cpp
adiaaida: Adding a TODO.

lib/Jit/LLILCJit.cpp
adiaaida: Done

lib/Jit/LLILCJit.cpp
adiaaida: Done.

lib/Reader/reader.cpp
adiaaida: I saw this with breaks too, which is where that nop for branches came from. I'm not sure how I would go about doing the check of if the last instruction in the current block doesn't have the stack empty debug info.

lib/Jit/LLILCJit.cpp
adiaaida: Actually, nope. It's used on line 406.

lib/Reader/readerir.cpp
adiaaida: This isn't unreachable. There are other types that I do not have support for that do end up reaching this point. It hasn't proved to be any sort of problem yet.

lib/Reader/reader.cpp
AndyAyersMS: For control flow I'd imagine that LLVM might optimize away "branch to next" even in debug codegen mode, so you'd lose the debug location if that happened.

If you're confident the only special case here (aside from branches) is `ldc` then just add a comment explaining why. If there are more special cases then I'd consider trying to do something more general.

lib/Jit/LLILCJit.cpp
pgavlin: Ah, yep. Good catch.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
 void LLILCJit::signalHandler(void *Cookie) {
   // do nothing
 }
+
+template <typename ObjSetT, typename LoadResult>
```
JosephTremoulet: As a rule, template function definitions should be in headers rather than cpp, no?

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +395,107 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+void ObjectLoadListener::operator()(llvm::orc::ObjectLinkingLayerBase::ObjSetHandleT ObjHandles,
+                const ObjSetT &Objs, const LoadResult &LoadedObjInfos) {
+  int I = 0;
+  for (const std::unique_ptr<object::ObjectFile> &PObj : Objs) {
```
JosephTremoulet: Since this is written as implementing a contract where the pointer kind (and collection type) of the `ObjectFile` pointer collection is (are) abstracted away in the type parameter, seems like you may as well use `const auto &PObj` instead here, to accommodate other pointer kinds.

include/Jit/LLILCJit.h
```diff
@@ -19,11 +19,32 @@
...
+#include "llvm/Support/Format.h"
+#include "llvm/Support/Errno.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/IR/DebugLoc.h"
+#include "llvm/Support/DataTypes.h"
+#include "llvm/Config/llvm-config.h"
+#include <vector>
+
+using namespace llvm;
+using namespace llvm::object;
```
JosephTremoulet: You don't seem to be making use of these new #includes or using statements later in this file; should they be here?

JosephTremoulet:
I seem to have suggested moving `ObjectLoadListener::operator()` from the .cpp to the .h in one comment, and moving the #includes it relies on from the .h to the .cpp in another comment.  It might make the most sense to move just the iteration part (that is templated over the pointer kind and set type) into the header file, but pull out the guts of the method into a non-template method that just takes an `ObjectFile &` and a `LoadedObjectInfo &`, whose definition would stay in the .cpp.

lib/Jit/LLILCJit.cpp
pgavlin: Only if they're intended for use in more than one translation unit. In this case, the function is local to LLILCJit.cpp, as so does not need to be defined in the header file. It might be better, in fact, if `ObjectLoadListener` was simply a forward declaration in LLILCJit.h and defined completely in LLILCJit.cpp.

adiaaida:
I'm pretty sure I got most of these comments. I have probably missed something. Please take a look and let me know.

@pgavlin, @JosephTremoulet I went with Joe's suggestion to split the operator() to the .h file and create a function for doing the actual debug information communication which is in the .cpp.

include/Jit/LLILCJit.h
```diff
@@ -174,6 +180,32 @@ class NullResolver : public llvm::RuntimeDyld::SymbolResolver {
   }
 };
 
+class ObjectLoadListener {
```
pgavlin: I still think that this type should be moved into LLILCJit.cpp, as it's not used outside that file. Its presence here is only to satisfy the typedef on line 221--this whole definition can be replaced with a forward declaration (i.e. `class ObjectLoadListener;`), which is all that is necessary for the header file.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+    const ObjectFile &Obj, const RuntimeDyld::LoadedObjectInfo &L) {
+  OwningBinary<ObjectFile> DebugObjOwner = L.getObjectForDebug(Obj);
+  const ObjectFile &DebugObj = *DebugObjOwner.getBinary();
+
+  // Get the address of the object image for use as a unique identifier
+  const void *ObjData = DebugObj.getData().data();
+  DIContext *DwarfContext = new DWARFContextInMemory(DebugObj);
+
+  // Use symbol info to iterate functions in the object.
+  // TODO: This may have to change when we have funclets
```
pgavlin: Should there be an issue for this?

adiaaida: Adding to my list. I haven't gotten around to adding issues.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
 }
+
+void ObjectLoadListener::getDebugInfoForObject(
+    const ObjectFile &Obj, const RuntimeDyld::LoadedObjectInfo &L) {
+  OwningBinary<ObjectFile> DebugObjOwner = L.getObjectForDebug(Obj);
+  const ObjectFile &DebugObj = *DebugObjOwner.getBinary();
+
+  // Get the address of the object image for use as a unique identifier
+  const void *ObjData = DebugObj.getData().data();
+  DIContext *DwarfContext = new DWARFContextInMemory(DebugObj);
```
pgavlin: I think that the heap allocation here can be removed. Instead of `DIContext *DwarfContext = new DWARFContextInMemory(DebugObj);`, `DWARFContextInMemory DwarfContext(DebugObj)` should do.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+
+  // Use symbol info to iterate functions in the object.
+  // TODO: This may have to change when we have funclets
+
+  for (symbol_iterator SI = DebugObj.symbol_begin(), E = DebugObj.symbol_end();
+       SI != E; ++SI) {
+    SymbolRef::Type SymType;
+    if (SI->getType(SymType))
+      continue;
+    if (SymType == SymbolRef::ST_Function) {
```
pgavlin: Nit: maybe change this block to an early-continue style for consistency with the other checks (i.e. instead of `if (SymType == SymbolRef::ST_Function) { ... }`, make this `if (SymType != SymbolRef::ST_Function) continue;`)

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+        continue;
+      if (SI->getAddress(Addr))
+        continue;
+      uint64_t Size = SI->getSize();
+
+      unsigned LastDebugOffset = -1;
+      unsigned NumDebugRanges = 0;
+      ICorDebugInfo::OffsetMapping *OM;
+
+      if (DwarfContext) {
```
pgavlin: I don't think that this check can every fail (it certainly can't if you change `DwarfContext` from a pointer to a value as suggested).

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+        DILineInfoTable Lines =
+            DwarfContext->getLineInfoForAddressRange(Addr, Size);
+
+        DILineInfoTable::iterator Begin = Lines.begin();
+        DILineInfoTable::iterator End = Lines.end();
+
+        // Count offset entries. Will skip an entry if the current IL offset
+        // matches
+        // the previous offset.
```
pgavlin: Formatting: please move `the previous offset` to the preceding line.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+            LastDebugOffset = LineNumber;
+          }
+        }
+
+        // Reset offset
+        LastDebugOffset = -1;
+        unsigned CurrentDebugEntry = 0;
+
+        if (NumDebugRanges > 0) {
+
```
pgavlin: Nit: extra blank line.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+          // Allocate OffsetMapping array
+          unsigned SizeOfArray =
+              (NumDebugRanges) * sizeof(ICorDebugInfo::OffsetMapping);
+          OM = (ICorDebugInfo::OffsetMapping *)Context->JitInfo->allocateArray(
+              SizeOfArray);
+
+          Begin = Lines.begin();
```
pgavlin: Isn't this redundant? `Begin` should already be positioned at `Lines.begin()`.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+          // Iterate through the debug entries and save IL offset, native
+          // offset, and
+          // source reason
+          for (DILineInfoTable::iterator It = Begin; It != End; ++It) {
+            int Offset = It->first;
+            int LineNumber = (It->second).Line;
+
+            // We store info about if the instruction is being recorded because
+            // it is a call in the column field
+            int IsCall = (It->second).Column;
```
pgavlin: I'd change this to `bool IsCall = (It->second).Column == 1;` to avoid the comparison in the ternary on line 491.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +407,100 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+
+            if (LineNumber != LastDebugOffset) {
+              LastDebugOffset = LineNumber;
+              OM[CurrentDebugEntry].nativeOffset = Offset;
+              OM[CurrentDebugEntry].ilOffset = LineNumber;
+              OM[CurrentDebugEntry].source =
+                  IsCall == 1 ? ICorDebugInfo::CALL_INSTRUCTION
+                              : ICorDebugInfo::STACK_EMPTY;
```
pgavlin: Is this always correct? I see a few other source types listed in corinfo.h. Do we not generate any of those other types yet?

lib/Reader/reader.cpp
```diff
@@ -6399,6 +6399,7 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
 
   while (CurrentOffset < fgNodeGetEndMSILOffset(Fg)) {
     CORINFO_RESOLVED_TOKEN ResolvedToken;
+    bool IsCall = false;
```
pgavlin: I think that having this up so high is confusing. I'd just inline the declaration at the sites that use it (at that point, I think the resulting declarations could also be made const).

lib/Reader/readerir.cpp
pgavlin: Are those types only aggregates? It would be nice to know exactly what we're missing here.

lib/Jit/LLILCJit.cpp
adiaaida: Yes. Ryujit only generates Call_Instruction and Stack_Empty, so we only generate those as well.

lib/Jit/LLILCJit.cpp
adiaaida: Thanks. Fixed. ccformat fail.

include/Jit/LLILCJit.h
JosephTremoulet: FWIW, that would be fine with me -- so long as the template declaration and definition are in the same file, there's no worry about referencing an undefined instantiation.

include/Jit/LLILCJit.h
```diff
@@ -19,11 +19,17 @@
...
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/ThreadLocal.h"
 #include "llvm/ExecutionEngine/Orc/ObjectLinkingLayer.h"
 #include "llvm/ExecutionEngine/Orc/IRCompileLayer.h"
+#include "llvm/Config/config.h"
+#include "llvm/Object/ObjectFile.h"
+
+using namespace llvm;
+using namespace llvm::object;
```
JosephTremoulet: Nit: Per [the LLVM coding standards](http://llvm.org/docs/CodingStandards.html#do-not-use-using-namespace-std), I think it's best not to put using statements in headers:

> In header files, adding a `using namespace XXX` directive pollutes the namespace of any source file that #includes the header.

include/Jit/LLILCJit.h
```diff
@@ -174,6 +180,32 @@ class NullResolver : public llvm::RuntimeDyld::SymbolResolver {
...
 };
 
+class ObjectLoadListener {
+public:
+  ObjectLoadListener(LLILCJitContext *Context) { this->Context = Context; }
+
+  template <typename ObjSetT, typename LoadResult>
+  void operator()(llvm::orc::ObjectLinkingLayerBase::ObjSetHandleT ObjHandles,
+                  const ObjSetT &Objs, const LoadResult &LoadedObjInfos) {
+    for (const std::unique_ptr<object::ObjectFile> &PObj : Objs) {
```
JosephTremoulet: Nit: I'd still prefer `const auto &PObj` here, since we're going to the trouble of abstracting the set type and pointer kind to match the abstraction in the calling code.

JosephTremoulet:
LGTM aside from a couple nits.

lib/Jit/LLILCJit.cpp
adiaaida: done.

lib/Jit/LLILCJit.cpp
adiaaida: done.

lib/Jit/LLILCJit.cpp
adiaaida: done.

lib/Jit/LLILCJit.cpp
adiaaida: done.

lib/Jit/LLILCJit.cpp
adiaaida: done.

lib/Jit/LLILCJit.cpp
adiaaida: done.

lib/Reader/reader.cpp
adiaaida: done.

include/Jit/LLILCJit.h
adiaaida: done.

include/Jit/LLILCJit.h
adiaaida: done.

include/Jit/LLILCJit.h
adiaaida: done.

adiaaida:
I think I got everything. Please let me know if you have anything else.

lib/Jit/LLILCJit.cpp
```diff
@@ -22,30 +22,66 @@
...
+
+class ObjectLoadListener {
+public:
+  ObjectLoadListener(LLILCJitContext *Context) { this->Context = Context; }
+
+  template <typename ObjSetT, typename LoadResult>
+  void operator()(llvm::orc::ObjectLinkingLayerBase::ObjSetHandleT ObjHandles,
+                  const ObjSetT &Objs, const LoadResult &LoadedObjInfos) {
+    for (const auto &PObj : Objs) {
+      int I = 0;
```
JosephTremoulet: This needs to be initialized outside the `for` loop.

adiaaida: fixed.

JosephTremoulet:
LGTM (once you push that last fix).  Thanks for the updates.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +431,95 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+    if (NumDebugRanges > 0) {
+      // Allocate OffsetMapping array
+      unsigned SizeOfArray =
+          (NumDebugRanges) * sizeof(ICorDebugInfo::OffsetMapping);
+      OM = (ICorDebugInfo::OffsetMapping *)Context->JitInfo->allocateArray(
+          SizeOfArray);
+
+      // Iterate through the debug entries and save IL offset, native
+      // offset, and
+      // source reason
```
pgavlin: Comment formatting is off--please move `source reason` into the preceeding line.

lib/Jit/LLILCJit.cpp
```diff
@@ -394,3 +431,95 @@ void __cdecl LLILCJit::fatal(int Errnum, ...) {
...
+
+      if (LineNumber != LastDebugOffset) {
+        NumDebugRanges++;
+        LastDebugOffset = LineNumber;
+      }
+    }
+
+    // Reset offset
+    LastDebugOffset = -1;
+    unsigned CurrentDebugEntry = 0;
```
pgavlin: Minor nit: I'd sink this declaration down to ~line 499 or so (immediately preceding `for (DILineInfoTable::iterator It = Begin; It != End; ++It) {`).

lib/Reader/reader.cpp
```diff
@@ -7342,6 +7355,10 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
       break;
 
     case ReaderBaseNS::CEE_RET: {
+      // Insert a nop for debug and set return's IL offset to signify
+      // that it is part of the epilogue
+      nop();
+      setDebugLocation(ICorDebugInfo::EPILOG, false);
```
pgavlin: Nit: please life the constant `false` into a local similar to the other call sites: `const bool IsCall = false`.

lib/Reader/readerir.cpp
pgavlin: Any more info here?

pgavlin:
LGTM modulo two nits and a request that we add more info about which types we're missing in `GenIR::convertType`.

adiaaida:
Added a comment about which types we're missing and fixed all the comments above. FWIW, we are missing LabelTy, MetadataTy, X86_MMXTy, FunctionTy, StructTy, ArrayTy, VectoryTy

pgavlin:
Do you know which of those we actually see at runtime?

adiaaida:
Definitely StructTy. Not sure about the others. When I was running and printing things, I was only running on a small number of functions. I haven't run it on anything terribly complicated.

adiaaida:
@dotnet-bot test this please

PR #649, opened 2015-06-15
Add ability to specify CoreCLR tests directory

adiaaida:
For users who have cloned CoreCLR into a directory named something other
than coreclr, the current checking of the current directory would fail.
This change allows the user to specify where their coreclr\tests
directory can be found, and allows users to run llilc_runtests.py
outside of coreclr\tests. The script will abort if runtest.cmd is not
found in the directory specified as the coreclr\tests path (or the
current working directory if none is specified). We then run runtest
specifying the full path.

### DIFF
--- a/test/llilc_runtest.py
+++ b/test/llilc_runtest.py
@@ -115,6 +115,8 @@ def main(argv):
                         help='the dump level: summary, or verbose')
     parser.add_argument('-r', '--result-path', type=str, 
                         default=DefaultResultPath(), help='the path to runtest result output directory')
+    parser.add_argument('-t', '--runtest-path', type=str,
+                        default=None, help='the full path to the CoreCLR\\tests directory')
     required = parser.add_argument_group('required arguments')
     required.add_argument('-j', '--jit-path', required=True, 
                         help='full path to jit .dll')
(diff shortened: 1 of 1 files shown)


adiaaida:
@pgavlin PTAL

test/llilc_runtest.py
```diff
@@ -171,7 +173,8 @@ def main(argv):
     # Run the test
     return_code = const.RunTestOK
     exclusion = os.path.join(os.path.dirname(__file__), 'exclusion.targets')
-    runtest_command = 'runtest ' + args.arch + ' ' + args.build 
+    runtest_full_path = os.path.join(runtest_dir, 'runtest')
```
pgavlin: I'd change `'runtest'` to `'runtest.cmd`', since that's what we're testing for the existence of above.

test/llilc_runtest.py
```diff
@@ -127,12 +129,12 @@ def main(argv):
...
-    parent_path, parent_folder = os.path.split(path)
-    if (not folder == 'tests' or
-        not parent_folder == 'coreclr'):
-        print('This script is required to run from tests directory in a CoreCLR repository.' )    
+    runtest_dir = args.runtest_path
+    if (args.runtest_path is None):
+        runtest_dir = os.getcwd()
+
+    if (not os.path.isfile(os.path.join(runtest_dir, "runtest.cmd"))):
+        print('Please specify --runtest-path or run from tests directory in a CoreCLR repository')
```
pgavlin: s/from tests directory/from the tests directory/

pgavlin:
LGTM aside from the comments above.

test/llilc_runtest.py
adiaaida: I think this is meant to also be run on Linux, which will run the runtest.sh script (whenever we get testing on Linux). I test for the cmd above because it should always be there (and is currently the only script there). I can change this, but I think the intent was to be ready for cross plat once we got testing there.

pgavlin: The current approach won't work on Linux, which requires that we append the file extension (i.e. `.sh`) to the path. IMHO we should always test for the script that we're going to run.

adiaaida: good to know. I will update this.

test/llilc_runtest.py
```diff
@@ -127,12 +129,19 @@ def main(argv):
...
+        runtest_dir = os.getcwd()
+
+    # All platforms other than Windows run runtest.sh
+    runtest_full_path = os.path.join(runtest_dir, 'runtest.sh')
+
+    # On Windows, we will run runtest.cmd
+    if sys.platform == "win32":
+      runtest_full_path = os.path.join(runtest_dir, 'runtest.cmd')
+
+    if (not os.path.isfile(os.path.join(runtest_full_path))):
```
pgavlin: There's a spurious call to `os.path.join` in here.

adiaaida: fixed.

pgavlin:
LGTM. Thanks for doing this.

adiaaida:
Checking in as at least one of the tests that run llilc_runtest.py completed successfully.

PR #651, opened 2015-06-15
Fix unwind info reporting.

### LINKED ISSUE #650: Unwind info size reported via allocUnwindInfo shouldn't include padding entry
Size passed to EEMemoryManager::registerEHFrames
includes the size of a padding UNWIND_CODE entry
so that the number of elements of the UnwindCode array is even.
Jit interface (allocUnwindInfo) expects the size reported to it not to include the padding
entry.

erozenfeld:
Size passed to EEMemoryManager::registerEHFrames
includes the size of a padding UNWIND_CODE entry
so that the number of elements of the UnwindCode array is even.
Jit interface (allocUnwindInfo) expects the size reported to it not to include the padding
entry.

This was causing an assert in crossgen.exe when using llilcjit.

Closes #650.


erozenfeld:
@JosephTremoulet PTAL

JosephTremoulet:
LGTM, thanks.

erozenfeld:
Retest this please

PR #661, opened 2015-06-20
Remove typedefs for jit layers

JosephTremoulet:
The jit layers are really an implementation detail, and don't need to be
exposed from LLILCJit.h.  Even in LLILCJit.cpp, they're only used to
declare the layers; this change inlines them into the declarations (using
decltype(PreviousLayer) to avoid excessive repetition of lower layer
types) and removes them entirely.

This avoids needing to make parallel changes in the typedefs and
declarations when adding/changing layers.

### DIFF
--- a/include/Jit/LLILCJit.h
+++ b/include/Jit/LLILCJit.h
@@ -28,7 +28,6 @@
 #include "llvm/Config/config.h"
 
 class ABIInfo;
-class ObjectLoadListener;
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
@pgavlin PTAL

pgavlin:
LGTM. Thanks for doing this!

PR #662, opened 2015-06-23
Fix Cmake for upcoming LLVM merge

AndyAyersMS:
In Microsoft/llvm@84b1665, LLVM stopped supplying a custom argument parser in favor of using the CMake builtin. We need to do likewise in LLIIC.

This is a prerequisite to merging latest LLVM sources.


AndyAyersMS:
@pgavlin PTAL.

pgavlin:
LGTM

PR #665, opened 2015-06-25
Use orc::NullResolver

JosephTremoulet:
LLILC doesn't need its own NullResolver class now that LLVM provides one as a utility.
Also remove some vestigial MCJit references.


JosephTremoulet:
@erozenfeld PTAL; not sure how this would interact with your pending changes.

pgavlin:
LGTM assuming that @erozenfeld is OK with the change.

erozenfeld:
I'm ok with the change. I may have to re-introduce a custom resolver depending on the outcome of my thread with Lang.

PR #666, opened 2015-06-26
Change getSize to getCommonSize to match LLVM change.

richardlford:
LLVM renamed SymbolRef::getSize to SymbolRef::getCommonSize,
so change LLILC code to use the new name.


richardlford:
@adiaaida, PTAL.

richardlford:
To pass testing this will need the corresponding LLVM change.

adiaaida:
LGTM

richardlford:
LLVM change has been merged, so now will do this merge which should pass in lab.

PR #667, opened 2015-06-26
Compute symbol sizes using common utility

JosephTremoulet:
The current call to getCommonSize fails because our function symbols are
not common symbols.

PR #668, opened 2015-06-26
Fix a small bug in getDebugInfoForObject.

pgavlin:
`getCommonSize` is only valid on symbols that have the `SF_Common` flag
set. Check for this flag before attempting to get symbol sizes.


pgavlin:
@adiaaida PTAL

adiaaida:
Joe actually has a different fix for this. see #667

pgavlin:
That's probably a better fix.

russellhadley:
Yeah.  I think Joe's is the one we should go with.

PR #671, opened 2015-06-27
Use ObjectTransformLayer to reserve unwind space

JosephTremoulet:
This layer just needs to call the memory manager's reserveUnwindSpace
method on each object as it passes through; use the
orc::ObjectTransformLayer utility rather than a custom object layer.


JosephTremoulet:
Note that this is a PR for a change to go into the EH branch.
@pgavlin PTAL.

pgavlin:
LGTM.

PR #673, opened 2015-06-30
Change call to SymbolRef::getType() to match LLVM changed signature.

richardlford:
LLVM changed the signature of SymbolRef::getType() by removing
its parameter, so modify call to match.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -455,9 +455,7 @@ void ObjectLoadListener::getDebugInfoForObject(
 
   for (const auto &Pair : SymbolSizes) {
     object::SymbolRef Symbol = Pair.first;
-    SymbolRef::Type SymType;
-    if (Symbol.getType(SymType))
-      continue;
+    SymbolRef::Type SymType = Symbol.getType();
     if (SymType != SymbolRef::ST_Function)
       continue;
 


richardlford:
This needs matching changes in the LLVM repo.

richardlford:
@adiaaida, @swaroop-sridhar PTAL.

adiaaida:
We need to be able to check SymType. Before, SI->getType(SymType) set SymType to the type of SI. So this needs to be changed to:

SymType = SI->getType();

and the if can be removed.

richardlford:
This passed a checked build locally but will not pass in the lab without the corresponding changes from the other repos, so I'll merge this without waiting for lab build to pass.

adiaaida:
Please see my comment. This does not have the correct semantics.

richardlford:
@adiaaida, I've corrected as you suggested. How does this look?

adiaaida:
LGTM.

richardlford:
Local testing has again passed (though the prior testing did not detect the semantic error), I'll be merging shortly.

swaroop-sridhar:
LGTM2

adiaaida:
The reason it didn't catch it is that this is the code that communicates the debug line info back to the CLR. You'd only see that it wasn't working correctly if you tried to step through something in the debugger. We're still working on getting automated debugger tests running in the lab (it's a long term thing. MDBG needs to be ported), so until we do, it's a good bet to try stepping through something to make sure debugging isn't broken.

Issue #676, opened 2015-07-01
Type substitution and cast needed for return value from shared method


richardlford:
When calling a shared method, the return type in the signature of the called method may involve
type variables. To get the effective type for the calling method a type substitution 
needs to be performed on the returned value. 

For example, a generic type ImmutableArray<T> can have a "T get_Item(int)" method
that returns an element of type T. If a caller has an ImmutableArray<char[]>, then 
get_Item should return a char[]. However when a shared version is called, the 
signature of the called method is "canon get_item(int)". So after the call the 
result needs a cast from type "canon" to type "char[]". 

We currently have some hacks that fix up return values to be the expected type
but we probably have not covered all cases, so this work is needed.

erozenfeld:
I don't think we can do such a substitution after the call with what JIT-EE interface provides. In particular, I don't think we can figure out the target type (char[]) in your example.

richardlford:
Move to next sprint.

PR #678, opened 2015-07-01
Ensure Array operations use correct Array type.

### LINKED ISSUE #630: Bad codegen for Microsoft.CodeAnalysis.Text.LargeEncodedText.get_Item
This item is part of #629, testing LLILC by using it to compile Roslyn that in turn compiles Roslyn.

@erozenfeld did the initial investigation. Quoting from his email:

> I was able to set up a Roslyn build, then extracted csc command for building 
> Microsoft.CodeAnalysis.CSharp.dll (biggest Roslyn assembly). I ran into a couple of asserts related 
> to type translation and checked in fixes for them. Roslyn csc.exe has /parallel[+|-] flag. The default is > /parallel+. With /parallel- there is just one issue: we have bad codegen for 
> Microsoft.CodeAnalysis.Text.LargeEncodedText.get_Item. If we exclude that method we successfully 
> complete compilation:
>
> 1 tests, 20569 methods, 20569 good, 0 bad (100% good)

This task is to identify the reason for the bad code and fix it.

richardlford:
Because we do not yet properly do type substitution on the
result of calls to generic-shared methods (#676), sometimes
the result of a call is not the correct type. In particular
sometimes we attempt an array operation on such a result
when the type of the result is not an array type. Typically
it is something like the "canonical" type used to represent
any reference type.

In such cases we need to cast the value to an LLVM type that
represents the correct MSIL type. Prior to this fix we were
just using a single "array of reference" type for these,
but that gives incorrect address arithmetic when the element
type has a size different from the size of a reference.

This change restructures and extends a number of methods
so that array types referencing the proper element type
are created.

For the case of ldelem and stelem instructions we know
what the element type should be, except for the
ldelem.ref and stelem.ref cases where the element type
should be taken from the input Array. In that case if
the input is not an array we settle for using
"array of ref".

For the ldlen instruction that loads the length of an
array, the result does not depend on the element type,
so if the element type is not known we just use
"array of ref" again.

In addition to these fixes, this change also modifies
LLILCJit::compileMethod so that it uses
JitOptions.IsAltJit so that only methods selected
by the ComPlusAltJit environment variables are
JITed by LLILC.

Resolves #630.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -260,7 +260,7 @@ public:
       : ReaderBase(JitContext->JitInfo, JitContext->MethodInfo,
                    JitContext->Flags),
         UnmanagedCallFrame(nullptr), ThreadPointer(nullptr),
-        ArrayOfReferenceType(nullptr) {
+        BuiltinObjectType(nullptr), ElementToArrayTypeMap() {
     this->JitContext = JitContext;
     this->ClassTypeMap = ClassTypeMap;
     this->ReverseClassTypeMap = ReverseClassTypeMap;
@@ -1312,19 +1312,25 @@ private:
   /// \returns Alignment in bytes.
   uint32_t convertReaderAlignment(ReaderAlignType ReaderAlignment);
 
-  /// Get array element type.
+  /// \brief Get array element type expected for an MSIL instruction.
   ///
-  /// \param Array Array node.
+  /// Note: we avoid the name getArrayElementType to avoid confusion
+  /// with llvm::Type::getArrayElementType.
+  ///
+  /// \param Array         When the CorInfoType is CORINFO_TYPE_CLASS this
+  ///                      is a call for ldelem.ref or stelem.ref. In that
+  ///                      case the element type is the same as element type
+  ///                      of the input array, so we need to pass it in.
   /// \param ResolvedToken Resolved token from ldelem or stelem instruction.
-  /// \param CorInfoType [IN/OUT] Type of the element (will be resolved for
+  /// \param CorType - [IN/OUT] Type of the element (will be resolved for
   /// CORINFO_TYPE_UNDEF).
   /// \param Alignment - [IN/OUT] Alignment that will be updated for value
   /// classes.
-  /// \returns Array element type.
-  llvm::Type *getArrayElementType(IRNode *Array,
-                                  CORINFO_RESOLVED_TOKEN *ResolvedToken,
-                                  CorInfoType *CorType,
-                                  ReaderAlignType *Alignment);
+  /// \returns LLVM Array element type.
+  llvm::Type *getMSILArrayElementType(IRNode *Array,
+                                      CORINFO_RESOLVED_TOKEN *ResolvedToken,
+                                      CorInfoType *CorType,
+                                      ReaderAlignType *Alignment);
 
   /// Check whether access to this multidimensional array can be expanded as an
   /// intrinsic.
@@ -1470,9 +1476,22 @@ private:
 
   /// Check if this LLVM type appears to be a CLR array type.
   ///
-  /// \param Type   Type to examine.
+  /// \param Type       Type to examine.
+  /// \param ElementTy  If non-null, the element type that must match.
+  ///                   otherwise any element type will do.
   /// \returns      True if this type looks like a CLR array type.
-  bool isArrayType(llvm::Type *Type);
+  bool isArrayType(llvm::Type *Type, llvm::Type *ElementTy);
+
+  /// If the Array is not typed as an MSIL array with specified element type
+  /// cast it to that type.
+  ///
+  /// \param Array     The array for which we want to ensure its typing.
+  /// \param ElementTy The LLVM element type desired. If null, means we
+  ///                  do not care about element type. In that case if
+  ///                  we need to make an array type we will use
+  ///                  Sytem.Object as the element type.
+  /// \returns         The possibly-casted input array.
+  IRNode *ensureIsArray(IRNode *Array, llvm::Type *ElementTy);
 
   /// Get the LLVM type for the built-in string type.
   ///
@@ -1484,16 +1503,20 @@ private:
   /// \returns    LLVM type that models the built-in object type.
   llvm::Type *getBuiltInObjectType();
 
-  /// Get the LLVM type for an array of references.
+  /// Get the LLVM type for an array of given element type.
   ///
   /// Used when we know that some type must be an array but our local
   /// type information thinks otherwise.
   ///
-  /// \returns    LLVM type that models array of references.
-  llvm::PointerType *getArrayOfReferenceType();
+  /// \param ElementTy                 The desired element type for the array,
+  /// \returns                         LLVM type that models the desired array
+  llvm::PointerType *getArrayOfElementType(llvm::Type *ElementTy);
 
-  /// Create the LLVM type for an array of references.
-  void createArrayOfReferenceType();
+  /// \brief Creates the type of the array with specified element type.
+  ///
+  /// \param ElementTy The LLVM element type that is desired for the array type.
+  /// \returns The LLVM representation of the MSIL array type.
+  llvm::PointerType *createArrayOfElementType(llvm::Type *ElementTy);
 
   /// Create the length, padding, and elements fields for an array type.
   ///
@@ -1502,22 +1525,12 @@ private:
   /// \param IsVector                  true iff this is a zero-lower-bound
   ///                                  single-dimensional array.
   /// \param ArrayRank                 Rank of the array.
-  /// \param ElementCorType            CorType for the array elements.
-  /// \param ElementClassHandle        Class handle for the array elements.
-  /// \param GetElementAggregateFields true iff this method should get type
-  ///                                  details for fields in case the element
-  ///                                  type is an aggregate or a pointer to an
-  ///                                  aggregate.
-  /// \param DeferredDetailClasses     List to append aggregates whose details
-  ///                                  weren't fully resolved.
+  /// \param ElementTy                 The LLVM element type desired.
   /// \returns                         Byte size of the fields added. Fields
   ///                                  updated with length, padding (if needed),
   ///                                  and the array itself.
-  uint32_t addArrayFields(
-      std::vector<llvm::Type *> &Fields, bool IsVector, uint32_t ArrayRank,
-      CorInfoType ElementCorType, CORINFO_CLASS_HANDLE ElementClassHandle,
-      bool GetElementAggregateFields,
-      std::list<CORINFO_CLASS_HANDLE> *DeferredDetailClasses = nullptr);
+  uint32_t addArrayFields(std::vector<llvm::Type *> &Fields, bool IsVector,
+                          uint32_t ArrayRank, llvm::Type *ElementTy);
 
   /// Add fields of a type to the field vector, expanding structures
   /// (recursively) to the types they contain.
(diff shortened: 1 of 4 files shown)


richardlford:
@erozenfeld, PTAL. @jimhogg, please look at changes to use AltJit filtering.

include/Reader/readerir.h
```diff
@@ -1484,16 +1501,21 @@ class GenIR : public ReaderBase {
...
+  /// \param ElementTy                 The desired element type for the array, 
+  ///                                  or if null, use array of references.
+  /// \returns                         LLVM type that models the desired array
+  llvm::PointerType *getArrayOfElementType(llvm::Type *ElementTy = nullptr);
 
-  /// Create the LLVM type for an array of references.
-  void createArrayOfReferenceType();
+  /// \brief Creates the type of the array with specified element type.
+  /// 
+  /// \param ElementTy The llvm element type that is desired for the array type.
```
erozenfeld: Capitalize llvm.

include/Reader/readerir.h
```diff
@@ -1515,6 +1540,7 @@ class GenIR : public ReaderBase {
   ///                                  and the array itself.
   uint32_t addArrayFields(
       std::vector<llvm::Type *> &Fields, bool IsVector, uint32_t ArrayRank,
+      llvm::Type *ElementTy,
```
erozenfeld: I think this signature is confusing. I would prefer to have just ElementTy parameter and remove ElementCorType and ElementClassHandle and have the caller always calculate ElementTy. That should also allow you to remove GetElementAggregateFields and DeferredDetailClasses.

lib/Jit/LLILCJit.cpp
```diff
@@ -212,91 +223,101 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
...
+    if (!TheTarget) {
+      errs() << "Could not create Target: " << ErrStr << "\n";
+      return CORJIT_INTERNALERROR;
+    }
+    TargetOptions Options;
+    CodeGenOpt::Level OptLevel;
+    if (Context.Options->OptLevel != ::OptLevel::DEBUG_CODE) {
+      OptLevel = CodeGenOpt::Level::Default;
+    }
+    else {
```
erozenfeld: This will fail ccformat.

lib/Jit/LLILCJit.cpp
```diff
@@ -212,91 +223,101 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
...
-  }
-  bool HasMethod = this->readMethod(&Context);
+    Context.TM = TM;
+
+    // Construct the jitting layers.
+    EEMemoryManager MM(&Context);
+    ObjectLoadListener Listener(&Context);
+    orc::ObjectLinkingLayer<decltype(Listener)> Loader(Listener);
+    orc::IRCompileLayer<decltype(Loader)> Compiler(Loader,
+      orc::SimpleCompiler(*TM));
```
erozenfeld: This will fail ccformat.

lib/Reader/readerir.cpp
```diff
@@ -2327,8 +2328,9 @@ bool GenIR::isUnmanagedPointerType(llvm::Type *Type) {
 
 uint32_t
 GenIR::addArrayFields(std::vector<llvm::Type *> &Fields, bool IsVector,
-                      uint32_t ArrayRank, CorInfoType ElementCorType,
-                      CORINFO_CLASS_HANDLE ElementHandle,
+                      uint32_t ArrayRank, 
+                      Type *ElementTy,
+                      CorInfoType ElementCorType, CORINFO_CLASS_HANDLE ElementHandle,
                       bool GetElementAggregateFields,
```
erozenfeld: This will fail ccformat.

lib/Reader/readerir.cpp
```diff
@@ -2327,8 +2328,9 @@ bool GenIR::isUnmanagedPointerType(llvm::Type *Type) {
 
 uint32_t
 GenIR::addArrayFields(std::vector<llvm::Type *> &Fields, bool IsVector,
-                      uint32_t ArrayRank, CorInfoType ElementCorType,
-                      CORINFO_CLASS_HANDLE ElementHandle,
+                      uint32_t ArrayRank, 
+                      Type *ElementTy,
+                      CorInfoType ElementCorType, CORINFO_CLASS_HANDLE ElementHandle,
```
erozenfeld: This will fail ccformat.

include/Reader/readerir.h
```diff
@@ -1484,16 +1501,21 @@ class GenIR : public ReaderBase {
...
   ///
   /// Used when we know that some type must be an array but our local
   /// type information thinks otherwise.
   ///
-  /// \returns    LLVM type that models array of references.
-  llvm::PointerType *getArrayOfReferenceType();
+  /// \param ElementTy                 The desired element type for the array, 
+  ///                                  or if null, use array of references.
+  /// \returns                         LLVM type that models the desired array
+  llvm::PointerType *getArrayOfElementType(llvm::Type *ElementTy = nullptr);
```
erozenfeld: My preference is for ElementTy not to be allowed to be nullptr. It should be Object if an array of references is asked for.

include/Reader/readerir.h
```diff
@@ -1484,16 +1501,21 @@ class GenIR : public ReaderBase {
...
 
-  /// Create the LLVM type for an array of references.
-  void createArrayOfReferenceType();
+  /// \brief Creates the type of the array with specified element type.
+  /// 
+  /// \param ElementTy The llvm element type that is desired for the array type.
+  /// \returns The LLVM representation of the MSIL array type.
+  llvm::PointerType *createArrayOfElementType(llvm::Type *ElementTy = nullptr);
```
erozenfeld: My preference is for ElementTy not to be allowed to be nullptr. It should be Object if an array of references is asked for.

lib/Reader/readerir.cpp
```diff
@@ -2422,20 +2451,40 @@ void GenIR::createArrayOfReferenceType() {
...
   // Install fields and give this a recognizable name.
   StructTy->setBody(Fields, true /* isPacked */);
-  StructTy->setName("__ArrayOfReferences");
+  std::string TypeName;
+
+  // We use a raw_string_ostream as a handly way to construct
+  // the name of the array type. This lets us use the
+  // Type::print method to get the element type name.
+  raw_string_ostream StringStream(TypeName);
+  StringStream << "__ArrayOf__";
```
erozenfeld: We should try to have names that look similar to what we get for arrays from getClassType. I think they look like object[].

lib/Reader/readerir.cpp
```diff
@@ -2395,22 +2399,47 @@ bool GenIR::isArrayType(llvm::Type *ArrayTy) {
...
+PointerType *GenIR::getArrayOfElementType(llvm::Type *ElementTy) {
+  PointerType *Array;
+  if (ElementTy == nullptr) {
+    if (ArrayOfReferenceType != nullptr) {
+      return ArrayOfReferenceType;
+    }
+  } else {
+    auto it = ElementToArrayTypeMap.find(ElementTy);
+    if (it != ElementToArrayTypeMap.end()) {
+      return (*it).second;
```
erozenfeld: Nit: why not it->second?

erozenfeld:
I think this is a correct fix but I left a number of refactoring suggestions and noted a few nits.

include/Reader/readerir.h
richardlford: Done.

include/Reader/readerir.h
richardlford: Agree. I've fixed it and adjusted callers.

lib/Jit/LLILCJit.cpp
richardlford: I'll run the formatter before the next push.

include/Reader/readerir.h
richardlford: Agreed for getArrayOfElementType. However in the case of isArrayType and ensureIsArray, 
we want to allow the user to specify a "don't care" for the element type. 
I've adjusted the code accordingly.

include/Reader/readerir.h
richardlford: Agreed and fixed for this as well.

lib/Reader/readerir.cpp
richardlford: I've modified so just append "[]" to end of element type name.

lib/Reader/readerir.cpp
richardlford: Changed.

richardlford:
I added the above comments while editing. I'm now going to reformat and then will do another squash and push.

richardlford:
I've now pushed the updates and locally ran the formatter. It did change some lines that I had not changed but hopefully the lab format checker will be ok with those changes.

richardlford:
@erozenfeld, please take another look.

richardlford:
I've pushed fixes for the formatting problems.

richardlford:
@dotnet-bot, retest please.

include/Reader/readerir.h
```diff
@@ -1312,19 +1312,25 @@ class GenIR : public ReaderBase {
...
-  /// \param Array Array node.
+  /// Note: we avoid the name getArrayElementType to avoid confusion
+  /// with llvm::Type::getArrayElementType.
+  ///
+  /// \param Array         When the CorInfoType is CORINFO_TYPE_CLASS this
+  ///                      is a call for ldelem.ref or stelem.ref. In that
+  ///                      case the element type is the same as element type
+  ///                      of the input array, so we need to pass it in.
   /// \param ResolvedToken Resolved token from ldelem or stelem instruction.
   /// \param CorInfoType [IN/OUT] Type of the element (will be resolved for
```
erozenfeld: CorInfoType --> CorType

include/Reader/readerir.h
```diff
@@ -1605,9 +1618,12 @@ class GenIR : public ReaderBase {
   uint32_t TargetPointerSizeInBits;
   const uint32_t UnmanagedAddressSpace = 0;
   const uint32_t ManagedAddressSpace = 1;
-  llvm::PointerType *ArrayOfReferenceType; ///< Array type for use in casting
-                                           ///< non-array types seen in array
-                                           ///< contexts.
+  llvm::Type *BuiltinObjectType; ///< Cached LLVM representation of object.
+
+  /// \brief Map from Element types to the LLVM representation of the
```
erozenfeld: We are using a different commenting style for fields.

include/Reader/readerir.h
richardlford: Fixed.

include/Reader/readerir.h
richardlford: When a declaration is short enough to leave room for a comment after it, that seems appropriate. But if a declaration is long I think it is better to have a preceding comment.

lib/Reader/readerir.cpp
```diff
@@ -4070,19 +4084,19 @@ void GenIR::storeElem(ReaderBaseNS::StElemOpcode Opcode,
...
   CORINFO_CLASS_HANDLE ClassHandle = nullptr;
   if (*CorType == CorInfoType::CORINFO_TYPE_CLASS) {
     PointerType *Ty = cast<PointerType>(Array->getType());
-    if (!isArrayType(Ty)) {
+    if (!isArrayType(Ty, nullptr)) {
       // Likely System.__Canon*. Assume array of object.
-      Ty = getArrayOfReferenceType();
+      Ty = getArrayOfElementType(this->getBuiltInObjectType());
     }
     StructType *ReferentTy = cast<StructType>(Ty->getPointerElementType());
```
erozenfeld: Can we use ensureIsArray here?

richardlford:
@dotnet-bot  retest please.

lib/Reader/readerir.cpp
richardlford: No. ensureIsArray mutates an array value. But this can be simplified. We just want the element type, so there is no need to actually construct the array of object just to then deduce that the element type is object. I'll simplify the code.

richardlford:
I've pushed the simplification of getMSILArrayElementType suggested by @erozenfeld's last comment.

erozenfeld:
LGTM. Thank you for making the changes.

pgavlin:
One more style comment: please do not prefix field accesses/method calls with `this` unless it is necessary for purposes of disambiguation.

richardlford:
@dotnet-bot, retest please. Got the following failure: 
"Nondeterministic failure occasionally seen building coreclr. Rerunning should fix the issue"
on the http://dotnet-ci.cloudapp.net/job/dotnet_llilc_check_win64_prtest/187/ leg.

richardlford:
@dotnet-bot, please retest.

richardlford:
@dotnet-bot, retest this please.

PR #680, opened 2015-07-02
Implement GC-Table generation

swaroop-sridhar:
Translate Gc-pointer liveness information from __llvm.stackmaps section 
to CoreCLR GCInfo Encoding.

Testing:
1) We can now build and a few tests in a precise-GC environment.
   The correctness of generated GCInfo is verified by:
   a) Examining the GcInfo dump in the debugger (SOS) and comparing
      it against the assembly dump.
   b) Stepping through the debugger and watch it scan through the
      reported GC-Roots.
   c) The execution crashes for the same functions if the liveness
      information is not reported.

2) The execution crashes if we try to build MsCorLib for precise-GC.
There are a few known issues (as noted in the comments) and the
remaining issues need to be fixed via debugging.

3) The standard LLILC testing is still done using conservative-gc.
So, no testing impact is expected for the lab. A separate test-leg will
be added to run tests with precise-GC.


swaroop-sridhar:
@AndyAyersMS, @pgavlin: Please review. Thanks

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+    dbgs() << "GcTable for Function: " << JitContext->MethodName << "\n"
+           << "  Size: " << JitContext->HotCodeSize << "\n";
+  }
+
+  // TODO: Set Code Length accurately.
+  // https://github.com/dotnet/llilc/issues/679
+  // JitContext->HotCodeSize is the size of the allocated code block.
+  // It is not the actual length of the current function's code.
+  Encoder->SetCodeLength(JitContext->HotCodeSize);
+  Encoder->SetSizeOfStackOutgoingAndScratchArea(0);
```
AndyAyersMS: This needs to be updated to have the true size at some point, right? Should there be an issue for this?

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+#if defined(BIGENDIAN)
+  typedef StackMapV1Parser<support::big> StackMapParserType;
+#else
+  typedef StackMapV1Parser<support::little> StackMapParserType;
+#endif
+  StackMapParserType StackMapParser(StackMapContentsArray);
+
+  assert(StackMapParser.getNumFunctions() == 1 &&
+         "Expect only one function with GcInfo in the module");
+
```
AndyAyersMS: Won't be true once EH gets a bit further; the funclets will look like separate functions.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+    // + OffsetCorrection: to account for any bytes before the start
+    //                     of the function.
+    // (-1) CLR expects the statepoint to be one byte before the start of the
+    //     instruction following the Call-instruction.
+    // (-1) Safepoints are encoded with a -1 adjustment.
+    unsigned InstructionOffset =
+        R.getInstructionOffset() + OffsetCorrection - 2;
+
+    CallSites[RecordIndex] = InstructionOffset;
+    CallSiteSizes[RecordIndex] = 2;
```
AndyAyersMS: Should this be `CallSiteSize` instead of `2`?

include/GcInfo/GcInfo.h
```diff
@@ -9,14 +9,44 @@
...
+  void emitGCInfo();
+
+private:
+  void encodeHeader();
+  void encodeLiveness();
+  void emitEncoding();
+
+  LLILCJitContext *JitContext;
+  uint8_t *LLVMStackMapData;
+  GcInfoEncoder *Encoder;
```
pgavlin: Can this be made a value rather than a pointer (i.e. `GcInfoEncoder`)?

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+  // CoreCLR's GCTables record pointer birth/deaths per Safepoint.
+  // So, we do the translation using old/new live-pointer-sets
+  //
+  // We need bit-sets for recording the liveness -- one bit per slot.
+  // But std::BitSet only supports fixed size bitsets, so we use
+  // vector<bool> which should be optimized by the compiler to use
+  // a bit-wide representation.
+
+  vector<bool> OldLiveSet;
+  vector<bool> NewLiveSet;
```
AndyAyersMS: `vector<bool>` is usually avoided in C++, see for instance [here](https://isocpp.org/blog/2012/11/on-vectorbool). LLVM may offer something better we can use.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+using namespace std;
+
+GCInfo::GCInfo(LLILCJitContext *JitContext, uint8_t *LLVMStackMapData,
+               uint8_t *CodeBlockStart) {
+  this->JitContext = JitContext;
+  this->LLVMStackMapData = LLVMStackMapData;
+  this->CodeBlockStart = CodeBlockStart;
+
+  this->Encoder = new GcInfoEncoder(JitContext->JitInfo, JitContext->MethodInfo,
+                                    new GcInfoAllocator());
```
pgavlin: Instead of allocating a `GcInfoAllocator` on the heap, I'd simply add a field of type `GcInfoAllocator` to `GCInfo` and pass in a pointer to that field.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+  Encoder->SetSizeOfStackOutgoingAndScratchArea(0);
+}
+
+void GCInfo::encodeLiveness() {
+  if (LLVMStackMapData == nullptr) {
+    return;
+  }
+
+  ArrayRef<uint8_t> StackMapContentsArray(
+      reinterpret_cast<const uint8_t *>(LLVMStackMapData),
```
pgavlin: You shouldn't need the `reinterpret_cast` here.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+  size_t OffsetCorrection = FunctionEntry - CodeBlockStart;
+
+  // Loop over LLVM StackMap records to:
+  // 1) Note CallSites (safepoints)
+  // 2) Assign Slot-IDs to each unique gc-pointer location (slot)
+  // 3) Record liveness (birth/death) of slots per call-site.
+
+  size_t NumCallSites = StackMapParser.getNumRecords();
+  unsigned *CallSites = new unsigned[NumCallSites];
+  BYTE *CallSiteSizes = new BYTE[NumCallSites];
```
pgavlin: Instead of always allocating arrays on the heap, you might want to use `llvm::SmallVector` with a size of 16 or so. Values of that type will use the stack until it runs out of room, and will deallocate any storage when they go out of scope. FWIW, these values are currently being leaked (they're never `delete[]`d).

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+  // interior pointers
+  GcSlotFlags SlotFlags = (GcSlotFlags)GC_SLOT_INTERIOR;
+
+  // TODO: Determine call-site-size accurately
+  // https://github.com/Microsoft/llvm/issues/56
+  // Call-site size is not yet available in LLVM's StackMap,
+  // so make up a value for now.
+  // The Call-instruction generated by LLILC on X86/X64 is typically
+  // Call [rax], which has a two-byte encoding.
+  uint8_t CallSiteSize = 2;
```
pgavlin: Might as well make this `const` for now.

pgavlin: Same goes for `SlotFlags` above.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+  // Loop over LLVM StackMap records to:
+  // 1) Note CallSites (safepoints)
+  // 2) Assign Slot-IDs to each unique gc-pointer location (slot)
+  // 3) Record liveness (birth/death) of slots per call-site.
+
+  size_t NumCallSites = StackMapParser.getNumRecords();
+  unsigned *CallSites = new unsigned[NumCallSites];
+  BYTE *CallSiteSizes = new BYTE[NumCallSites];
+  map<int32_t, GcSlotId> SlotMap;
```
pgavlin: An `llvm::DenseMap` is probably a better choice here.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+      case StackMapParserType::LocationKind::Register:
+        // TODO: Report Live - GC values in Registers
+        // https://github.com/dotnet/llilc/issues/474
+        // Live gc-pointers are currently spilled to the stack at Safepoints.
+        assert(false && "GC-Pointer Live in Register");
+        break;
+
+      case StackMapParserType::LocationKind::Direct: {
+
```
pgavlin: Nit: extra blank line.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+        default:
+          assert(false && "Unexpected stack base-pointer");
+        }
+
+        GcSlotId SlotID;
+        int32_t Offset = Loc.getOffset();
+        map<int32_t, GcSlotId>::iterator ExistingSlot = SlotMap.find(Offset);
+        if (ExistingSlot == SlotMap.end()) {
+          SlotID = Encoder->GetStackSlotId(Offset, SlotFlags, SpBase);
+
```
pgavlin: Nit: extra blank line.

lib/GcInfo/GcInfo.cpp
pgavlin: I think `llvm::BitVector` is probably what you want here.

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,271 @@
...
+          break;
+        default:
+          assert(false && "Unexpected stack base-pointer");
+        }
+
+        GcSlotId SlotID;
+        int32_t Offset = Loc.getOffset();
+        map<int32_t, GcSlotId>::iterator ExistingSlot = SlotMap.find(Offset);
+        if (ExistingSlot == SlotMap.end()) {
+          SlotID = Encoder->GetStackSlotId(Offset, SlotFlags, SpBase);
```
pgavlin: Given the way that these slot IDs are used, an assertion that the newly returned slot id is exactly `NumSlots` would be useful.

lib/GcInfo/GcInfo.cpp
swaroop-sridhar: The CallSites and CallSiteSizes arrays are dictated by the encoder's DefineCallSites() API.
I'll add a destructor for GcInfo to delete the allocated CallSite structures.

lib/GcInfo/GcInfo.cpp
swaroop-sridhar: I'll handle this case in a separate checkin.
https://github.com/dotnet/llilc/issues/687

lib/GcInfo/GcInfo.cpp
```diff
@@ -0,0 +1,296 @@
...
+#include "llvm/Object/StackMapParser.h"
+#include <sstream>
+
+using namespace llvm;
+
+GCInfo::GCInfo(LLILCJitContext *JitCtx, uint8_t *StackMapData,
+               uint8_t *CodeBlkStart, GcInfoAllocator *Allocator)
+    : JitContext(JitCtx), LLVMStackMapData(StackMapData),
+      CodeBlockStart(CodeBlkStart), 
+      Encoder(JitContext->JitInfo, JitContext->MethodInfo, Allocator) {
```
swaroop-sridhar: The encoder constructor causes allocation using the GcInfoAllocator passed in.

lib/GcInfo/GcInfo.cpp
swaroop-sridhar: To clarify, this will need additional work on the LLVM side to associate records with functions in the StackMap section.

swaroop-sridhar:
@pgavlin, @AndyAyersMS:

Please review changes in: https://github.com/swaroop-sridhar/llilc/commit/a0b00ac64799646f753e531acca21676e1488405

pgavlin:
LGTM.

AndyAyersMS:
LGTM too.

PR #682, opened 2015-07-02
Fix build flags and several compiler warnings

JosephTremoulet:
Our cmake files were inadvertently clobbering build flags set by LLVM's cmake files when trying to append -EHsc.  Fix that, and the ensuing fall-out from enabling -W4.

### DIFF
--- a/include/Jit/EEMemoryManager.h
+++ b/include/Jit/EEMemoryManager.h
@@ -18,7 +18,7 @@
 
 #include "llvm/ExecutionEngine/RTDyldMemoryManager.h"
 
-class LLILCJitContext;
+struct LLILCJitContext;
 
 namespace llvm {
 
--- a/include/Reader/abi.h
+++ b/include/Reader/abi.h
@@ -139,7 +139,8 @@ public:
 /// The \p ABIInfo class provides ABI-specific services. Currently, this is
 /// limited to computing the details of how arguments are passed to functions
 /// for a given platform.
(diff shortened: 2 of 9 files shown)


JosephTremoulet:
This makes the warnings we're getting about AlignOf go away (one of the LLVM flags getting clobbered was the flag to disable that warning), but brings in some new warnings due to unreferenced locals in the gc encoder file we compile from the CoreCLR repo.  I'm not sure if we should try to match the CoreCLR's build flags for that file, or suppress the warning for that file, or change the file in the CoreCLR repo... this change to compile LLVM with the flags it expects and fix the warnings in LLILC code is at least a step in the right direction.

JosephTremoulet:
@adiaaida PTAL.  It's all pretty mechanical, but you may want to look at the change to ObjectLoadListner::getDebugInfoForObject; the comment there made it seem like an odd line to remove.

russellhadley:
@JosephTremoulet do we see similar shake out in warnings on Linux?  I'd expect that we'd see some other warnings churn from clang.

JosephTremoulet:
@russellhadley no, the part where we accidentally clobbered flags (to set -EHsc) is conditional on WIN32, so no change on Linux

JosephTremoulet:
I.e. we were already passing the right flags there

russellhadley:
LGTM

adiaaida:
LGTM

PR #689, opened 2015-07-06
Modify SymbolRef getName and getAddress calls for new signature.

richardlford:
The SymbolRef::getName and SymbolRef.getAddress methods
now use ErrorOr<T> return types to return either an
error code or the desired value. The calls to these
were changed to capture the ErrorOr<T> value and
then test for error. If no error then extract the
value and proceed to use it.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -475,12 +475,16 @@ void ObjectLoadListener::getDebugInfoForObject(
       continue;
 
     // Function info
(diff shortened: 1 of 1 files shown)


richardlford:
This change is needed for an LLVM FI. The change will not pass until the LLVM FI is merged.

pgavlin:
LGTM.

richardlford:
This change (along with the matching LLVM changes) has passed local testing with a checked build.

PR #690, opened 2015-07-07
Enable llilc as an ngen jit.

### LINKED ISSUE #11: Find a way to refer symbolically to the addresses handed out by the EE
We get a lot of raw addresses from the EE for things like methods and static fields. For instance:

  %1 = call %System.Text.Encoding addrspace(1)* inttoptr (i64 140733439908056 to %System.Text.Encoding addrspace(1)* (i32)*)(i32 %0)

They make the IR dumps somewhat cryptic. Also these addresses vary from run to run so we have to strip them out when we diff the dumps. Sure would be nice to have names here instead of numbers, like so:

  %1 = call %System.Text.Encoding addrspace(1)*System.Text.Encoding::GetEncoding(i32 %0)

### LINKED ISSUE #655: NGEN: record relocations via Jit interface
In the jit scenario we get final pointers for strings, methods, etc. from the EE. In NGEN scenario we get handles and need to call recordRelocation on the jit interface for each code location using those handles.

### LINKED ISSUE #656: NGEN: verify that we process handle indirections correctly
Any differences in how handles should be interpreted (no indirection, single indirection, double indirection) is already abstracted via InfoAccessType enum. When we get a handle from the CLR we also get an InfoAccessType value:

// Can a value be accessed directly from JITed code.
enum InfoAccessType
{
    IAT_VALUE,      // The info value is directly available
    IAT_PVALUE,     // The value needs to be accessed via an       indirection
    IAT_PPVALUE     // The value needs to be accessed via a double indirection
};

We need to make sure the existing code is correct for the new codepaths.

erozenfeld:
These changes depend on commit b2cc30c in coreclr (for ngen scenario)
and on commit d71fba5 in llvm (for both jit and ngen scenarios).

With these changes llilc jit can ngen mscorlib.dll (with some NYIs).
All CoreCLR tests enabled for llilc jit pass against mscorlib.dll
ngen-ed by llilc jit.

The numbers for ngen-ing mscorlib.dll are:
1 tests, 25329 methods, 24110 good, 1219 bad (95% good)

Changes include:

* Create GlobalVariables for handles and method addresses.
* Use a custom symbol resolver that tells LLVM to skip relocation
processing for symbols corresponding to GlobalVariables we create.
* Update ObjectLoadListener to register relocations for GlobalVariables
we create via Jit interface.
* Fix unwind info allocation calls to account for cases when
CountOfUnwindCodes is 0.
* Fix a bug in insertIRForSecurityObject (handles were transposed in
handleToIRNode call).
* Add a fix for ngen quirck in getBoxedType.
* Factor getClassNameWithNamespace.
* Add several NYIs for ngen scenario:
  * IL stubs
  * Floating point opcodes (ldc.r4, ldc.r8, conv.r4, conv.r8, conv.run)
(LLVM uses constant pulls)
  * Switch opcode (LLVM generates jump table accesses)

All bitcode dumps (both in jit and ngen scenarios) now have
function, helper, handle, etc. names instead of raw constants.
For example, here is the dump for HelloWorld (I wrapped some lines
manually):

define i32 @C.Main() {
entry:
  %loc0 = alloca i32
  store i32 0, i32* %loc0
  br label %0

; <label>:0                                       ; preds = %entry
  %1 = load %System.String addrspace(1)*,
       %System.String addrspace(1)** bitcast (i64* @"String(TK_70000001)"
       to %System.String addrspace(1)**), align 8, !dbg !8
  call void bitcast (i64* @"System.Console.WriteLine(TK_a000003)"
       to void (%System.String addrspace(1)*)*)(%System.String addrspace(1)* %1), !dbg !9
  store i32 0, i32* %loc0, !dbg !10
  br label %2

; <label>:2                                       ; preds = %0
  %3 = load i32, i32* %loc0, align 8, !dbg !11
  ret i32 %3, !dbg !12
}

Closes #655.
Closes #656.
Closes #11.


erozenfeld:
@pgavlin @JosephTremoulet PTAL

lib/Jit/EEMemoryManager.cpp
```diff
@@ -41,6 +41,11 @@ uint8_t *EEMemoryManager::allocateCodeSection(uintptr_t Size,
   return this->HotCodeBlock;
 }
 
+uint8_t *EEMemoryManager::getCodeSection() {
+  // ColdCodeBlock is not currently used.
```
pgavlin: Should this be marked with a TODO?

include/Jit/LLILCJit.h
```diff
@@ -164,6 +167,35 @@ struct LLILCJitPerThreadState {
...
+/// The ObjectLinkingLayer takes a SymbolResolver ctor parameter.
+class EESymbolResolver : public llvm::RuntimeDyld::SymbolResolver {
+public:
+  EESymbolResolver(llvm::StringMap<uint64_t> *NameToHandleMap) {
+    this->NameToHandleMap = NameToHandleMap;
+  }
+
+  llvm::RuntimeDyld::SymbolInfo findSymbol(const std::string &Name) final {
+    // Address 0 means that we don't know how to resolve this symbol, which will
+    // result in an LLVM error.
```
JosephTremoulet: Why delay the error?  Seems more straightforward to assert count == 1 here and unconditionally return UINT64_MAX as the address.

lib/Jit/LLILCJit.cpp
```diff
@@ -548,3 +565,60 @@ void ObjectLoadListener::getDebugInfoForObject(
...
+  switch (LLVMRelocationType) {
+  case IMAGE_REL_AMD64_ABSOLUTE:
+    return IMAGE_REL_BASED_ABSOLUTE;
+  case IMAGE_REL_AMD64_ADDR64:
+    return IMAGE_REL_BASED_DIR64;
+  case IMAGE_REL_AMD64_REL32:
+    return IMAGE_REL_BASED_REL32;
+
+  default:
+    assert(!"Unknown reloc type");
```
JosephTremoulet: You could use llvm_unreachable for this

lib/Reader/reader.cpp
```diff
@@ -3486,17 +3486,22 @@ IRNode *ReaderBase::genericTokenToNode(CORINFO_RESOLVED_TOKEN *ResolvedToken,
   // No runtime lookup is required
   if (!Result.lookup.lookupKind.needsRuntimeLookup) {
     if (NeedResult) {
+      mdToken Token = ResolvedToken->token;
+      if ((Result.handleType == CORINFO_HANDLETYPE_CLASS) &&
+          (TypeFromToken(Token) != mdtClassHandle)) {
```
JosephTremoulet: What's this checking for?  Comment might be helpful.

lib/Reader/readerir.cpp
```diff
@@ -1932,6 +1904,52 @@ void GenIR::createOverlapFields(
...
+  // Note some constructed types like arrays may not have names.
+  int32_t NameSize = 0;
+  const bool IncludeNamespace = true;
+  const bool FullInst = false;
+  const bool IncludeAssembly = false;
+  // We are using appendClassName instead of getClassName because
+  // getClassName omits namespaces from some types (e.g., nested classes).
+  // We may still get the same name for two different classes because
+  // two classes with the same fully-qualified names may live in different
+  // assemblies.
```
JosephTremoulet: What will happen if we get the same name for different classes, will that cause an error?  If so, would fixing it be as simple as passing true for IncludeAssembly here?

lib/Reader/readerir.cpp
```diff
@@ -4739,12 +4774,57 @@ bool GenIR::canMakeDirectCall(ReaderCallTargetData *CallTargetData) {
...
-  uint32_t NumBits = TargetPointerSizeInBits;
-  bool IsSigned = false;
+GlobalVariable *GenIR::GetGlobalVariable(uint64_t LookupHandle,
+                                         uint64_t ValueHandle, Type *Ty,
+                                         StringRef Name, bool IsConstant) {
+  GlobalVariable *GlobalVar = nullptr;
+  auto MapElement = HandleToGlobalVariableMap.find(LookupHandle);
+  if (MapElement != HandleToGlobalVariableMap.end()) {
+    GlobalVar = MapElement->second;
+  }
```
JosephTremoulet: I believe you could replace this with`GlobalVariable *&GlobalVar = HandleToGlobalVariableMap[LookupHandle]` and then just assign to GlobalVar below instead of repeating the lookup with operator[] there.

lib/Reader/readerir.cpp
```diff
@@ -4739,12 +4774,57 @@ bool GenIR::canMakeDirectCall(ReaderCallTargetData *CallTargetData) {
...
+  const char *ModuleName = nullptr;
+  const char *MethodName =
+      JitContext->JitInfo->getMethodName(MethodHandle, &ModuleName);
+
+  std::string FullName;
+  raw_string_ostream OS(FullName);
+  OS << format("%s.%s(TK_%x)", ModuleName, MethodName, MethodToken);
+  OS.flush();
+  GlobalVariable *GlobalVar =
+      GetGlobalVariable(Handle, Handle, CodeAddrTy, FullName, IsConstant);
```
JosephTremoulet: Should we be using `Function`s instead of `GlobalVariable`s for these?  (I'm not trying to say we should, just wondering what your thoughts are.  On the one hand, with Functions we could apply function-level attributes that analyses might pick up on; on the other hand, these really are pretty opaque to us so maybe it's safer to model them as indirect calls via function pointer global variables.)

lib/Jit/LLILCJit.cpp
```diff
@@ -548,3 +565,60 @@ void ObjectLoadListener::getDebugInfoForObject(
...
+        Symbol->getSection(SecI);
+        const bool IsExtern = SecI == Obj.section_end();
+        if (IsExtern) {
+          // This is an external symbol. Verify that it's one we created for
+          // a global variable and report the relocation via Jit interface.
+          uint64_t RelType = I->getType();
+          uint64_t Offset = I->getOffset();
+          ErrorOr<StringRef> NameOrError = Symbol->getName();
+          if (!NameOrError) {
+            continue; // Error.
```
JosephTremoulet: Why aren't we reporting the error?

JosephTremoulet:
LGTM, just some nits.

lib/Reader/reader.cpp
erozenfeld: Will add a comment:
      // If the handle really corresponds to a class and not a method, pass
      // mdtClassHandle to handleToIRNode so that a class name is used for the
      // name of the GlobalVariable created there instead of the method name.

lib/Reader/readerir.cpp
erozenfeld: No, that won't cause an error. The result of this method is used to set Struct names or GlobalVariable names. LLVM ensures that the names are unique by appending a numerical suffix on conflicting names. Passing true for IncludeAssembly will result in fully qualified class names (including version, public key token, etc.), which will clutter the dumps. I don't think that's what we should since conflicting class names are very rare.

lib/Reader/readerir.cpp
erozenfeld: There are some practical difficulties with using `Function`s here. `Function`s require the correct `FunctionType` when they are created and we don't have one yet. If we find a very good reason to use `Function`s that can probably be changed but it's not a trivial change.

lib/Jit/EEMemoryManager.cpp
erozenfeld: Done.

include/Jit/LLILCJit.h
erozenfeld: Done.

lib/Jit/LLILCJit.cpp
erozenfeld: Done.

lib/Reader/readerir.cpp
erozenfeld: Good suggestion! Done.

lib/Jit/LLILCJit.cpp
erozenfeld: Changed to an assert.

erozenfeld:
All feedback addressed. @JosephTremoulet @pgavlin PTAL.

pgavlin:
Are there bugs for the NYIs? LGTM otherwise.

erozenfeld:
I'll open bugs once these changes are in.

erozenfeld:
I pushed a fix for a linux build error below. @JosephTremoulet or @pgavlin PTAL.

/mnt/j/workspace/dotnet_llilc_linux_release_prtest/llvm/tools/llilc/lib/Jit/LLILCJit.cpp:220:19: error: call to implicitly-deleted copy constructor of 'LLILCJitContext'
16:32:11   LLILCJitContext Context = LLILCJitContext(PerThreadState);
16:32:11                   ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
16:32:11 /mnt/j/workspace/dotnet_llilc_linux_release_prtest/llvm/tools/llilc/include/Jit/LLILCJit.h:93:29: note: copy constructor of 'LLILCJitContext' is implicitly deleted because field 'NameToHandleMap' has a deleted copy constructor
16:32:11   llvm::StringMap<uint64_t> NameToHandleMap; ///< Map from global variable names
16:32:11                             ^
16:32:11 /mnt/j/workspace/dotnet_llilc_linux_release_prtest/llvm/include/llvm/ADT/StringMap.h:229:3: note: copy constructor is implicitly deleted because 'StringMap<unsigned long, llvm::MallocAllocator>' has a user-declared move constructor

JosephTremoulet:
I think you could just replace `LLILCJitContext Context = LLILCJitContext(PerThreadState);` with `LLILCJitContext Context(PerThreadState);` to construct it directly and avoid the copy. Your change looks correct too. You may want to rewrite to avoid the copy even on top of that (or instead).

erozenfeld:
Good point. I'll go with your suggestion.

erozenfeld:
retest

pgavlin:

PR #693, opened 2015-07-08
Cleanup and improve LLILC debug facilities

richardlford:
Modify LLILC options processing by adding the
following options:

* COMPlus_ALtJitExclude is a MethodSet. LLILC will only
  attempt to compile methods which are in the COMPlus_AltJit
  set, but not in the COMPlus_ALtJitExclude set.
* COMPlus_ALtJitBreak is a MethodSet. This is used
  when running LLILC under the control of a debugger.
  It causes a "break" message to be written for
  matching methods when the method is about to be
  jitted. The method is written from the
  LLILCJit::compileMethod method in file
  llilc\lib\Jit\LLILCJit.cpp. The user must put
  a breakpoint on the statement that writes
  out the message in order to actually get a
  break.
* COMPlus_ALtJitMSILDump is a MethodSet. The MSIL for
  methods in the set is dumped before JITTING the
  method.
* COMPlus_ALtJitLLVMDump is a MethodSet. The LLVM IR for
  methods in the set is dumped just after LLILC has
  finished reading the MSIL and converting it to LLVM IR.
* COMPlus_ALtJitCodeRangeDump is a MethodSet. For methods
  in the set, the starting and ending address of the method's
  code, and the code size, is printed after the method has
  been JITTED. This can be useful in identifying which
  method contains a given address.

In addition the llilc_runtest.py file was modified
to allow specification of running a single test.

In addition:

1. Delete LLILCJitContext::outputDebugMethodName and
   LLILCJitContext::outputSkippingMethodName and replace
   with simpler output using Context.MethodName.
2. ReaderBase::printMSIL(uint8_t, uint32_t, uint32_t) is made
   static so it can be called from outside ReaderBase.
   When it calls parseILOpcode, it passes nullptr as the
   Reader parameter.
3. parseILOpcode was modified so that if its Reader parameter
   is null, it uses dbPrint for error messages rather than
   terminating with a fatal error. This is friendlier as
   this only happens when called from printMSIL and we would
   like printMSIL to be robust as it is called from the
   debugger.

4. Provide printMSIL facility even for release mode.
   This fixes a build error for release mode.
   ReaderBase::printMSIL was being called but not being
   defined in release mode. I believe, at least during
   this bring-up period, that we want this facility
   even for release mode, so I removed some conditional
   compilation directives so that dumping of MSIL is
   available in any mode.

5. Added Documentation/Debugging.md file to
   document how to debug LLILC. More will be added to
   this file, but we have already added documentation
   of the environment variables used by LLILC
   (but not yet those used when LLILC is NGENing).

6. Added Python solution for the Test directory. This
   allows testing and single stepping in the
   llilc_runtest.py script (or others). The properties
   of the test.pyproj can be used to set the
   script to test and the parameters to pass.
   This required installing "Python Tools for Visual Studio"
   The checked-in test.pyproj has properties with
   template values that users should modify to fit their
   environment and desires.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -164,24 +164,6 @@ extern "C" void __stdcall sxsJitStartup(void *CcCallbacks) {
   // nothing to do
 }
 
-void LLILCJitContext::outputDebugMethodName() {
-  const char *DebugClassName = nullptr;
-  const char *DebugMethodName = nullptr;
-
-  DebugMethodName = JitInfo->getMethodName(MethodInfo->ftn, &DebugClassName);
-  dbgs() << format("INFO:  jitting method %s::%s using LLILCJit\n",
-                   DebugClassName, DebugMethodName);
-}
-
-void LLILCJitContext::outputSkippingMethodName() {
-  const char *DebugClassName = nullptr;
-  const char *DebugMethodName = nullptr;
-
-  DebugMethodName = JitInfo->getMethodName(MethodInfo->ftn, &DebugClassName);
-  dbgs() << format("INFO:  skipping jitting method %s::%s using LLILCJit\n",
-                   DebugClassName, DebugMethodName);
-}
-
 LLILCJitContext::LLILCJitContext(LLILCJitPerThreadState *PerThreadState)
     : State(PerThreadState), HasLoadedBitCode(false) {
   this->Next = State->JitContext;
@@ -239,8 +221,15 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
   // Flags and MethodInfo.
   JitOptions JitOptions(Context);
 
+  if (JitOptions.IsBreakMethod) {
+    dbgs() << "INFO:  Breaking for method " << Context.MethodName << "\n";
+  }
+  if (JitOptions.IsMSILDumpMethod) {
+    dbgs() << "INFO:  Dumping MSIL for method " << Context.MethodName << "\n";
+    ReaderBase::printMSIL(MethodInfo->ILCode, 0, MethodInfo->ILCodeSize);
+  }
   CorJitResult Result = CORJIT_INTERNALERROR;
-  if (JitOptions.IsAltJit) {
+  if (JitOptions.IsAltJit && !JitOptions.IsExcludeMethod) {
     Context.Options = &JitOptions;
 
     // Construct the TargetMachine that we will emit code for
@@ -273,12 +262,17 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
 
     // Now jit the method.
     if (Context.Options->DumpLevel == DumpLevel::VERBOSE) {
-      Context.outputDebugMethodName();
+      dbgs() << "INFO:  jitting method " << Context.MethodName
+             << " using LLILCJit\n";
     }
     bool HasMethod = this->readMethod(&Context);
 
     if (HasMethod) {
-
+      if (JitOptions.IsLLVMDumpMethod) {
+        dbgs() << "INFO:  Dumping LLVM for method " << Context.MethodName
+               << "\n";
+        Context.CurrentModule->dump();
+      }
       if (Context.Options->DoInsertStatepoints) {
         // If using Precise GC, run the GC-Safepoint insertion
         // and lowering passes before generating code.
@@ -307,7 +301,13 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
 
       // TODO: ColdCodeSize, or separated code, is not enabled or included.
       *NativeSizeOfCode = Context.HotCodeSize + Context.ReadOnlyDataSize;
-
+      if (JitOptions.IsCodeRangeMethod) {
+        errs() << "LLILC compiled: "
+               << ", Entry = " << *NativeEntry
+               << ", End = " << (*NativeEntry + *NativeSizeOfCode)
+               << ", size = " << *NativeSizeOfCode
+               << " method = " << Context.MethodName << '\n';
+      }
       GcInfoAllocator GcInfoAllocator;
       GCInfo GcInfo(&Context, MM.getStackMapSection(), MM.getHotCodeBlock(),
                     &GcInfoAllocator);
@@ -329,7 +329,8 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
   } else {
     // This method was not selected for jitting by LLILC.
     if (JitOptions.DumpLevel == DumpLevel::SUMMARY) {
-      Context.outputSkippingMethodName();
+      dbgs() << "INFO:  skipping jitting method " << Context.MethodName
+             << " using LLILCJit\n";
     }
   }
 
--- a/lib/Jit/jitoptions.cpp
+++ b/lib/Jit/jitoptions.cpp
@@ -36,6 +36,11 @@
 
 MethodSet JitOptions::AltJitMethodSet;
 MethodSet JitOptions::AltJitNgenMethodSet;
+MethodSet JitOptions::ExcludeMethodSet;
+MethodSet JitOptions::BreakMethodSet;
+MethodSet JitOptions::MSILMethodSet;
+MethodSet JitOptions::LLVMMethodSet;
+MethodSet JitOptions::CodeRangeMethodSet;
 
 template <typename UTF16CharT>
 char16_t *getStringConfigValue(ICorJitInfo *CorInfo, const UTF16CharT *Name) {
@@ -69,6 +74,12 @@ JitOptions::JitOptions(LLILCJitContext &Context) {
   DoTailCallOpt = queryDoTailCallOpt(Context);
 
   LogGcInfo = queryLogGcInfo(Context);
+
+  IsExcludeMethod = queryIsExcludeMethod(Context);
+  IsBreakMethod = queryIsBreakMethod(Context);
+  IsMSILDumpMethod = queryIsMSILDumpMethod(Context);
+  IsLLVMDumpMethod = queryIsLLVMDumpMethod(Context);
+  IsCodeRangeMethod = queryIsCodeRangeMethod(Context);
 
   // Validate Statepoint and Conservative GC state.
   assert(DoInsertStatepoints ||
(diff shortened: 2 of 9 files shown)


richardlford:
@pgavlin, PTAL. I believe you wrote the code that compares summary output from test 
runs, so perhaps you can point me to the places that need changed there.

lib/Reader/reader.cpp
```diff
@@ -336,7 +336,7 @@ uint32_t parseILOpcode(uint8_t *ILInput, uint32_t ILOffset, uint32_t ILSize,
   if (TheOpcode == ReaderBaseNS::CEE_ILLEGAL) {
     if (ReportErrors) {
       if (Reader == nullptr) {
-        ReaderBase::fatal(CORJIT_BADCODE);
+        dbPrint("parseILOpcode: Bad Opcode\n");
```
russellhadley: Does this have the same semantics as ReaderBase::fatal()?  I don't get how keeps throwing the exception.

lib/Reader/reader.cpp
```diff
@@ -363,7 +363,7 @@ uint32_t parseILOpcode(uint8_t *ILInput, uint32_t ILOffset, uint32_t ILSize,
 underflow:
   if (ReportErrors) {
     if (Reader == nullptr) {
-      ReaderBase::fatal(CORJIT_BADCODE);
+      dbPrint("parseILOpcode: Underflow\n");
```
russellhadley: Same question as above.

lib/Jit/LLILCJit.cpp
```diff
@@ -303,6 +283,16 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
...
+
+      if (Context.Options->DumpLevel >= ::DumpLevel::SUMMARY) {
+         errs() << "LLILC compiled: " 
+                << ", Entry = " << *NativeEntry 
+                << ", End = " << (*NativeEntry + *NativeSizeOfCode)
+                << ", size = " << *NativeSizeOfCode 
+                << " method = " << Context.MethodName 
+                << '\n';
+      }
+
```
russellhadley: This block is moved before the verification of the bitcode.  Is that the right thing to do?  I would have thought we wanted to run the IR check to ensure that we have a legal translation.

lib/Reader/reader.cpp
richardlford: No. But we do not want it do. Reader is only null when this method is called from
printMSIL. I don't want printMSIL to terminate execution if there is a problem.

lib/Reader/reader.cpp
richardlford: See response above.

lib/Jit/LLILCJit.cpp
richardlford: I don't understand your comment. This added code is at the end of compileMethod (except for some cleanup).
It is after the LLVM has been converted to machine code. 
One question I have is whether we can fail after reading the MSIL, but before reaching this
point. From the source code I don't see us trying to catch exceptions or checking
status of passes.

russellhadley: Ahh.  Ok.  I see it.

russellhadley: Touch base with @erozenfeld,  The semantics of this are now a little different than the old check.  We're not marking successful reading, but additionally the full compile.

pgavlin: Why not break this into two lines?

```
errs() << "Successfully read " << Context.MethodName << '\n';
errs() << "\tEntry: " << *NativeEntry << ", End: " << (*NativeEntry + *NativeSizeOfCode) << ", Size: " << *NativeSizeOfCode << '\n';
```

Keeping the original format in the first line would eliminate the need to update any of the scripts.

richardlford: The reason to keep on a single line, and with the address first is so that the file can be sorted
and used as a map from addresses to method name.
When debugging I often want to know if a location is in code we jitted and this
give that capability.

I have a fix for the llilc_checkpass.py file that I'll be pushing as soon as I've tested it.
By using a non-capturing group in the regular expression I'll be able to avoid having
to renumber the captured groups.

pgavlin: > When debugging I often want to know if a location is in code we jitted and this
give that capability.

OK. FWIW, the sos extension for WinDBG already gives you this capability via the `!ip2md` command.

I wonder if a new dump level--say DEBUG--would be useful here to avoid cluttering the logs in the typical case. No strong feelings, however.

erozenfeld: +1 on @pgavlin's suggestion. I would prefer a new dump level for these details.

richardlford:
@russellhadley, @AndyAyersMS, @pgavlin, @adiaaida Please take a look. This is a version that 
does not modify the output looked at by the test scripts.

richardlford:
I've pushed correction for building in release mode. I'm going to wait to squash all the 
commits until after review correction and the added documentation have been added.

richardlford:
Also keeping individual commits until the end makes it easier to see the incremental changes
and to review incrementally.

swaroop-sridhar:
@dotnet-bot test this please

russellhadley:
I'd like to see these split out into their own env variables rather than having a more complicated syntax for the passed variable.  Please review the set that are used by RyuJIT and see where the overlap is.

richardlford:
The RyuJit configuration variables are declared in the
coreclr\src\inc\clrconfigvalues.h file.

Here are the RyuJit options that seem closest to what I'm providing. In RyuJit the data type that provides a set of methods (on which our MethodSet is based) is ConfigMethodSet. I search sources for variables of that type. I found the following:

* JitImportBreak - This causes a break at the beginning of Jitting the selected methods. 
  They implement the break by causing a failed assertion. My preference is to have
  a place that prints a message and let the user set a debugger breakpoint on the 
  statement that prints the message. That way if they decide they don't want to
  break on all matching methods they can turn the breakpoint off.
* JitBreak - This is essentially the same as JitImportBreak (except for details that
  do not concern us here). 
* JitDebugBreak - sets a flag that does not appear to be used.
* JitHalt - Insert a nop and debug break instructions at the beginning of the generated
  code for selected methods so the debugger will be entered when the 
  method is executed. We could use the llvm.debugtrap intrinsic for this.
* JitExclude - Skips compiling the matching methods. 
* JitInclude - If specified, methods to be compiled must also match this.
  For us that function seems to be provided by AltJit itself. 
* JitDump* - There are a variety of JitDump options. 
  * JitDump - this dumps tree version of Ryujit IR. But I think the name is rather
    ambiguous so I've prefer options that specify what is being dumped. 
  * JitGCDump - dump garbage collection info
  * JitUnwindDump - dump unwind info
  * JitEHDump - dump exception handling info.
  * JitDisasm, JitLateDisasm - dump disassembly at two different locations

richardlford:
Given the above information about RyuJit options, I'd like to propose the following:
* JitExclude - to exclude methods from being compiled by LLILC
* JitBreak - to print a break message when LLILC encounters matching methods, that
  the user can set a breakpoint at. Or if desired we could also have a version that
  unconditionally enters the debugger. 
* JitMsilDump - dump MSIL for selected methods. 
* JitLlvmDump - dump llvm IR for selected methods. Do we want the option of dumping the 
  LLVM IR at different phases, e.g. at the beginning, and at the end?
* JitCodeRangeDump - dump the code address range of matching methods. 

For the future I'd suggest implementing the following:
* JitHalt - to insert debugger break at start of generated code for a method.
* JitDisasm - to get disassembly for methods
* Dumps for other information, .e.g EH, GC.

richardlford:
One concern might be that if we do not compile all the methods with LLILC,
then those compiled by RyuJit may be affected by setting of these options.
So should all our options start with AltJit? E.G. AltJitExclude?

richardlford:
I've pushed a new version which includes a start at documenting the LLILC 
debugging including environment variables. I've updated the PR description
above. @russellhadley, PTAL.

lib/Jit/jitoptions.cpp
```diff
@@ -50,11 +58,52 @@ void freeStringConfigValue(ICorJitInfo *CorInfo, UTF16CharT *Value) {
 }
 
 JitOptions::JitOptions(LLILCJitContext &Context) {
-  // Set 'IsAltJit' based on environment information.
-  IsAltJit = queryIsAltJit(Context);
+
+  if (!StaticOptionsInitialized) {
```
russellhadley: This block should be broken up into individual initializers for the options like the others are handled.

lib/Jit/jitoptions.cpp
```diff
@@ -50,11 +58,52 @@ void freeStringConfigValue(ICorJitInfo *CorInfo, UTF16CharT *Value) {
...
+
 
   // Set dump level for this JIT invocation.
-  DumpLevel = queryDumpLevel(Context);
+  DumpLevel = StaticDumpLevel;
+  const char *ClassName = nullptr;
+  const char *MethodName = 
+    Context.JitInfo->getMethodName(Context.MethodInfo->ftn, &ClassName);
+  PCCOR_SIGNATURE PCSig = Context.MethodInfo->args.pSig;
+  IsAltJit = AltJitPtr->contains(MethodName, ClassName, PCSig);
```
russellhadley: IsAltJit was already hoisted out into it's own initializer?  Why did this change?

lib/Jit/jitoptions.cpp
```diff
@@ -50,11 +58,52 @@ void freeStringConfigValue(ICorJitInfo *CorInfo, UTF16CharT *Value) {
...
+  IsExcludeMethod = ExcludeMethodSet.contains(MethodName, ClassName, PCSig);
+  IsBreakMethod = BreakMethodSet.contains(MethodName, ClassName, PCSig);
+  IsMSILDumpMethod = MSILMethodSet.contains(MethodName, ClassName, PCSig);
+  IsLLVMDumpMethod = LLVMMethodSet.contains(MethodName, ClassName, PCSig);
+  IsEntrySizeMethod = EntrySizeMethodSet.contains(MethodName, ClassName, PCSig);
```
russellhadley: I don't understand why this all changed.  Can you explain why this had to change from how I had it set up?

richardlford: The explanation for the change is that it is more efficient to check for the once-only 
initialization in just one place, rather than in multiple places. 

In addition, the previous methods of checking for initialization of MethodSets was to
use their "isInitialized" method, but for MethodSets that are not specified these
would never be initialized. In any case I think it is more confusing to have lots of
little "isInitialized" flags than to have just one, since all the options are
initialized at once (the first time LLILC is invoked).

The reason to change the code for AltJit is that it is a MethodSet so we should
handle it uniformly in the same way we handle the other MethodSets.

The refactoring is also more efficient each time JitOptions is constructed since
one call to getMethodName serves for all the calls to MethodSet.contains.

I didn't do a completely thorough job of converting all of the options to this approach.
That is why there still are some remaining separate query methods that remain.
I was going to complete the job in a subsequent check-in, though that is
not essential.

russellhadley: I don't particular like this approach nor do I buy the efficiency argument.  The separate routines were to allow for separate query for each that was easily understandable on it's own.  If you look at the RyuJIT handling you can see it's organically grown into the very large convolution of code.  Keeping each option query separate allows for much easier separation of the logic.  Having an isInitialized per methodset allows for the method set to be used more flexabily while preserving the (simple) thread safety.  My main complaint is that the options are hand inlined into the constructor and when the number of options them grows into the double or tipple digits this is going to make for a really confusing method.  We should probably catch up in person.

richardlford:
@russellhadley, I've made the refactoring changes that you suggested. PTAL.

richardlford:
There was a linux build failure. I pushed what I hope will fix that. There seem to be some win64 run failures.

richardlford:
Squashed commits and fixed formatting errors.

richardlford:
One more try at fixing formatting. Removed trailing whitespace, squished.

richardlford:
@russellhadley, it is now passing testing. Any further comments?

lib/Jit/jitoptions.cpp
```diff
@@ -161,6 +172,56 @@ ::DumpLevel JitOptions::queryDumpLevel(LLILCJitContext &Context) {
   return JitDumpLevel;
 }
 
+bool JitOptions::queryIsExcludeMethod(LLILCJitContext &JitContext) {
+  return queryMethodSet(JitContext, ExcludeMethodSet,
+                        (const char16_t *)UTF16("AltJitExclude"));
+}
+
+bool JitOptions::queryIsBreakMethod(LLILCJitContext &JitContext) {
```
russellhadley: Can we make BreakMethod's name more specific as to where it stops so it's easier to disambiguate from the JitHalt semantics (inserted dbgbreak in the generated code)?

russellhadley:
Other than the naming nit it looks fine.

richardlford:
I've changed AltJitBreak to AltJitBreakAtJitStart. After testing passes I'll merge.

PR #694, opened 2015-07-08
Add support for EH filters

JosephTremoulet:
Filters are outlined during initial IR generation (which includes
explicitly escaping any locals referenced in filters), to simplify
modeling their effects in the IR (invoking the outlined function must be
included as a possible side-effect of any invoke that may throw an
exception, since this must bottom out in a call to the throw helper [which
is external code] and the filter function has external linkage).

The ReaderBase class is unaware of the outlining.  GenIR generates the
filter code inline in the main function (except for an initial point
block) during its main passes, and then moves filter blocks to the correct
function in the readerPostVisit pass that this change adds between reading
MSIL and removing unreachable code.

The CurrentRegion tracked by the reader is used by GenIR to check whether
code currently being generated is destined for an outlined filter or not,
in the places where that is needed.  References to parameters and locals
check if the use is in a filter and insert escape code in the parent
function and recover code in the filter as necessary.

Any invoke in a filter-protected region needs a nominal edge to the filter
as well as to its landingpad (the landingpad doesn't have explicit flow to
the filter); this is achieved by having FlowGraphSuccessorEdgeList track
down and iterate through the entry blocks of any filters referenced in the
landingpad.

There's also some juggling in fgNodeGetNext to ensure that the point
blocks for filter entries are visited in their correct place in MSIL
order.

This change renames GenIR's Function field to RootFunction as reminder
that some methods will have multiple functions.  A few places are updated
to iterate through all functions in the module where previously they were
just operating on the one Function, including marking functions with
unmanaged call frames and removing unreachable code.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -2947,6 +2947,9 @@ public:
   // Called between building the flow graph and inserting the IR
   virtual void readerMiddlePass(void) = 0;
 
+  // Called after reading all MSIL, before removing unreachable blocks
+  virtual void readerPostVisit() = 0;
+
   // Called when reader has finished processing method.
   virtual void readerPostPass(bool IsImportOnly) = 0;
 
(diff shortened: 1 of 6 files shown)


JosephTremoulet:

@pgavlin PTAL.

pgavlin:
LGTM to me at a glance. I'll need a little bit more time to read through the specifics, but feel free to merge this in the meantime.

PR #697, opened 2015-07-08
Modify to use Intrinsic::localescape instead of Intrinsic::frameescape.

richardlford:
LLVM renamed Intrinsic::frameescape to be Intrinsic::localescape,
so use the new name.


richardlford:
This change is for an LLVM FI. Testing will not pass until the LLVM change is merged.

richardlford:
@AndyAyersMS, PTAL.

richardlford:
Matching LLVM and LLILC changed locally passed LLILC testing with a checked build.

AndyAyersMS:
LGTM. @JosephTremoulet FYI you will need to do likewise on your EH branch.

PR #706, opened 2015-07-09
Add ngen option to llilc_runtest

adiaaida:
To run tests with llilc ngen'd mscorlib, we need to not disable zap.
This change adds a switch to llilc_runtest, --ngen. When passed, we do
not set COMPLUS_ZapDisable.


adiaaida:
@erozenfeld PTAL

erozenfeld:
LGTM

PR #707, opened 2015-07-10
Enable ngen for il stubs.

### LINKED ISSUE #698: NGEN: il stub support
We are using constants instead of relocatable GlobalVariables for ThreadTrapAddress. That causes an AV in il stubs ngen-ed by llilc jit.

erozenfeld:
Use GlobalVariables instead of constants for computing thread trap addresses.

Fix read-only section alignment issue. It was causing an assert in allocateDataSection.

Closes #698.


erozenfeld:
@pgavlin @JosephTremoulet PTAL

JosephTremoulet:
LGTM

pgavlin:
LGTM as well.

PR #710, opened 2015-07-11
Don't use LLVMBuilder after removing unused blocks.

### LINKED ISSUE #709: Don't use LLVMBuilder after removing unused blocks
Removing unused blocks may invalidate LLVMBuilder's insert point. Currently we insert IR to keep generic context alive and IR for security object in readerPostPass, which runs after removal of unused blocks. That results in AV on some tests  with debug llilc jit.

The reason we haven't seen this in the lab is that we use ReleaseWithDebugInfo build. It uses msvcr120.dll (debug build uses msvcr120d.dll) and its implementation of operator delete doesn't overwrite deallocated memory. So we are accessing deallocated memory without an AV.

erozenfeld:
Removing unused blocks may invalidate LLVMBuilder's insert point.
Currently we insert IR to keep generic context alive and IR for security object in readerPostPass,
which runs after removal of unused blocks. That results in AV on some tests with debug llilc jit.

The reason we haven't seen this in the lab is that we use ReleaseWithDebugInfo build.
It uses msvcr120.dll (debug build uses msvcr120d.dll) and its implementation of operator delete
doesn't overwrite deallocated memory. So we are accessing deallocated memory without an AV.

@JosephTremoulet recently introduced readerPostVisit pass in his EH branch.
It runs after reading all MSIL and before removing unused blocks.
So it's a perfect place to move code that inserts IR to keep generic context alive
and IR for security object.

Closes #709.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -2929,6 +2929,9 @@ public:
   // Called between building the flow graph and inserting the IR
   virtual void readerMiddlePass(void) = 0;
 
+  // Called after reading all MSIL, before removing unreachable blocks
+  virtual void readerPostVisit() = 0;
+
   // Called when reader has finished processing method.
   virtual void readerPostPass(bool IsImportOnly) = 0;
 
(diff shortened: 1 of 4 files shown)


erozenfeld:
@JosephTremoulet PTAL

JosephTremoulet:
LGTM

erozenfeld:

PR #714, opened 2015-07-17
Changes for FI from master to MS

richardlford:
1. Set "no-frame-pointer-elim on Pinvoke stubs.


richardlford:
@pgavlin, this only includes your fix. I removed the changes I had made to llilc_runtest.py as I'm including them in a separate checking.

richardlford:
@pgavlin, PTAL. These changes passed local checked and release testing.

PR #715, opened 2015-07-23
Add an option to test with Precise-GC

swaroop-sridhar:
Add the --precise-gc option to llilc_runtest.py to test with precise GC.


adiaaida:
LGTM

PR #719, opened 2015-07-27
Change getDataLayout to createDataLayout

adiaaida:
The call to getting a DataLayout object changed in LLVM. Update LLILC to
use the new call.


adiaaida:
@pgavlin PTAL. This will fail in the lab until the llvm changes are committed.

pgavlin:
LGTM. Thanks for the fix.

PR #721, opened 2015-07-28
Fix formatting error.

richardlford:
A line was 81 characters long so needed to be broken.


richardlford:
http://dotnet-ci.cloudapp.net/job/dotnet_llilc_mergetest/8/console job passed except for 
formatting error that this corrects, so will proceed with merge.

PR #725, opened 2015-07-31
Implement translation of jmp msil instruction.

### LINKED ISSUE #192: Reader: implement jmp msil instruction

erozenfeld:
jmp is implemented as a call marked with musttail followed by a ret.

If the caller has a secret parameter we need to pass it to the jmp call
because the signature and the calling convention of the caller and
the musttail call must match. To accommodate that a parallel
LLVM change modifies CLR_SecretParameter calling convention
to make the secret parameter the first pointer-sized parameter
(still passed in R10) rather than a parameter passed with a
"CLR_SecretParameter" attribute.

All indirect parameters are passed as they come to the caller, without
copying them to the current frame.

Indirect result parameter is typed as a managed pointer in a jmp call
because we type indirect result parameter for the caller as a
managed pointer.

Closes #192.

### DIFF
--- a/lib/Reader/abisignature.cpp
+++ b/lib/Reader/abisignature.cpp
@@ -312,7 +312,7 @@ CallSite ABICallSignature::emitUnmanagedCall(GenIR &Reader, Value *Target,
 
 Value *ABICallSignature::emitCall(GenIR &Reader, Value *Target, bool MayThrow,
                                   ArrayRef<Value *> Args,
-                                  Value *IndirectionCell,
+                                  Value *IndirectionCell, bool IsJmp,
                                   Value **CallNode) const {
   assert(Target->getType()->isIntegerTy(Reader.TargetPointerSizeInBits));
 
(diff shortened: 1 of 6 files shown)


erozenfeld:
@pgavlin PTAL

lib/Reader/abisignature.cpp
```diff
@@ -312,7 +312,7 @@ CallSite ABICallSignature::emitUnmanagedCall(GenIR &Reader, Value *Target,
 
 Value *ABICallSignature::emitCall(GenIR &Reader, Value *Target, bool MayThrow,
                                   ArrayRef<Value *> Args,
-                                  Value *IndirectionCell,
+                                  Value *IndirectionCell, bool IsJmp,
```
pgavlin: Given that it looks like we'll need to do most (all?) of the transformations that are conditional on this parameter for tail calls as well, I wonder if it wouldn't be better to split this into `bool IsTailCall, bool MustTailCall` (or `bool IsTailCall, bool IsJmp`). Or maybe even something like:

```
enum CallKind
{
    Normal = 0,
    TailCall = 1,
    MustTailCall =2
}
```

lib/Reader/readerir.cpp
```diff
@@ -5100,6 +5110,49 @@ IRNode *GenIR::convertToBoxHelperArgumentType(IRNode *Opr,
...
+      NormalArg = (IRNode *)Arguments[I];
+    } else {
+      NormalArg = loadArg(I, IsJmp);
+    }
+    ReaderOperandStack->push(NormalArg);
+  }
+
+  // This will generate a call marked with musttail.
+  IRNode *CallNode = nullptr;
+  rdrCall(&Data, Opcode, &CallNode);
```
pgavlin: Will this push the result of the call on the stack? If so, I think we need to pop it (or update `genCall` to avoid generating a `ReturnNode` at all).

erozenfeld: No, it won't. For a normal call the result is pushed on the stack in ReaderBase::readBytesForFlowGraphNodeHelper.

lib/Reader/abisignature.cpp
erozenfeld: jmp is different from tail.call. For example, tail.call doesn't have to have the same signature as the caller so we don't have to pass a secret parameter. I would rather keep this as IsJmp for now to make it clear that this code has only been implemented and tested for jmp. When we get to tail calls we will probably do something like what you suggested.

lib/Reader/readerir.cpp
pgavlin: Okay.

lib/Reader/abisignature.cpp
pgavlin: Okay, that seems fine. You're right about the restrictions--I had misread the lang ref.

pgavlin:
LGTM.

erozenfeld:
llilc test failures are expected until corresponding llvm changes are in.

Issue #726, opened 2015-07-31
Turn on misc-unused-parameters check


swaroop-sridhar:
We have several failures if we turn on the [misc-unused-parameters] clang-tidy check.
Investigate if the parameters can be removed and if so, turn on the rule.

PR #727, opened 2015-07-31
Fix clang-tidy invocation in ccFormat

swaroop-sridhar:
1) Add -fExceptions argument so that functions with throw build OK with clang.
2) Disable the [misc-unused-parameters] temporarily since there
   are many failures for this rule.
   [https://github.com/dotnet/llilc/issues/726]

### DIFF
--- a/utils/ccformat.py
+++ b/utils/ccformat.py
@@ -57,7 +57,9 @@ def runTidy(args):
         "-target x86_64-pc-win32-msvc",
         "-fms-extensions",
         "-fms-compatibility",
-        "-fmsc-version=1800"
+        "-fmsc-version=1800",
+        "–fexceptions",
+        "-fcxx-exceptions"
     ]
     includes = [
         os.path.join(llilcLib, "Jit"),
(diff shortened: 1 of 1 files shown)


swaroop-sridhar:
@rhadley, please review

swaroop-sridhar:
I've verified that ccFormat works OK with the clang-tidy binaries provided by Michelle with this change.

swaroop-sridhar:
I'll merge the change to get lab testing green. @adiaaida please review post check-in.

adiaaida:
It looks like this fixed the tidy issue, but now there are some unrelated formatting changes that need to be made.

swaroop-sridhar:
Since the clang-format binary has changed, we should rerun with --fix.
I'm happy to do this if you can share the clang-format binary.

We should have a share available to everyone, where they can get the same binary tools used by this job.

russellhadley:
@adiaaida per our discussion about the binary, let's get the link to @swaroop-sridhar.

PR #728, opened 2015-08-01
Support for ngen-ing of methods with switches and floating point instructions.

### LINKED ISSUE #699: NGEN: support for methods with switches
LLVM generates jump tables for some switches and the code we generate to access them is incorrect in ngen scenario.

### LINKED ISSUE #700: NGEN: support for methods with floating point operations
LLVM uses constant pools for floating point operations and the code we generate to access them is incorrect in ngen scenario.
This affects conv.r4, conv.r8, conv.run, ldc.r4, ldc.r8 opcodes.

erozenfeld:
Process all relocations in ObjectLoadListener::recordRelocations.
Fix an issue with read-only section alignment: make sure Constant
pools have a 16 bytes alignment.

Create a copy of ObjectLinkingLayer (EEObjectLinkingLayer) so that
we can suppress relocation resolutiotion by LLVM. Lang Hames is
likely to re-design ObjectLinkingLayer in the future to allow more
control over relocations. When that happens we should be able to get rid
of our copy.

Closes #699.
Closes #700.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -22,6 +22,7 @@
 #include "readerir.h"
 #include "abi.h"
 #include "EEMemoryManager.h"
+#include "EEObjectLinkingLayer.h"
 #include "llvm/CodeGen/GCs.h"
 #include "llvm/Config/llvm-config.h"
 #include "llvm/DebugInfo/DIContext.h"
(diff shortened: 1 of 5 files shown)


erozenfeld:
@pgavlin PTAL

russellhadley:
LGTM.

lib/Jit/LLILCJit.cpp
```diff
@@ -552,41 +549,61 @@ void ObjectLoadListener::getDebugInfoForObject(
...
-          assert(NameOrError);
-          StringRef TargetName = NameOrError.get();
-          assert(Context->NameToHandleMap.count(TargetName) == 1);
-          uint64_t Target = Context->NameToHandleMap[TargetName];
-          Context->JitInfo->recordRelocation(MM->getCodeSection() + Offset,
-                                             (void *)Target,
-                                             translateRelocationType(RelType));
-        }
-        ++I;
+      if (!SectionName.compare(".pdata")) {
```
pgavlin: Why do we need to skip the `.pdata` section?

erozenfeld: Relocation types we get there are not supported by the jit interface's recordRelocation. We won't use the contents of the .pdata section so there is no need to resolve these relocations.

PR #736, opened 2015-08-04
Print command line arguments passed to ccformat

adiaaida:
In the lab, we aren't seeing any output from ccformat. To confirm that
it is being run, we want to print out the command line arguments that
were passed to ccformat.py.

### DIFF
--- a/utils/ccformat.py
+++ b/utils/ccformat.py
@@ -116,6 +116,7 @@ def runTidy(args):
     clangArgs = " ".join(["-p", expandPath(args.compile_commands)])
 
   returncode = 0
+  print("Running clang-tidy")
   for dirname,subdir,files in os.walk(llilcSrc):
     for filename in files:
       if filename.endswith(".c") or filename.endswith(".cpp"):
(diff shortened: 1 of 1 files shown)


adiaaida:
@swaroop-sridhar PTAL

This will print 

Running ccformat.py with: ['--fix']

if the command is
ccformat.py --fix

swaroop-sridhar:
I think we should print the command line for both clang-tidy and clang-format -- just printing one is confusing.

The other option is to not print anything from within ccFormat -- but change the lab-job to just print the invocation of ccFormat. (No further output means success).

adiaaida:
I tried to do the later. I could not get it to print out the invocation of ccformat.py. 

The only way I've been able to get it to run with printing ccformat's invocation is to do vcvarsall.bat && ccformat.py which I didn't want to do here in case we needed to do something after ccformat.

swaroop-sridhar:
OK, then let's print a message for the invocation of both clang-tidy and clang-format then. For clang-tidy, since we do several invocations, we could just print "Running clang-tidy ... " or something like that once. [To ensure that I won't be confused later that it is not running]. Thanks.

swaroop-sridhar:
Apart from the above comment, green for me. Thanks for fixing this.

adiaaida:
Fixed in place. We now print "Running clang-tidy" and "Running clang-format" before running those steps.

PR #738, opened 2015-08-04
Test with Statepoints: Add exclusions for failing tests tests

swaroop-sridhar:
Add exclusion for the four tests failing with Statepoints Inserted
when using conservative GC.

The condition in exclusion.targets is based off the
Complus_InsertStatepoints environment variable, taking advantage of
the fact that the lab job sets this variable on the command line
before invoking llilc_runtest.


swaroop-sridhar:
@adiaaida: Please review.

adiaaida:
LGTM

PR #739, opened 2015-08-05
Update ModuleDataLayout from TargetMachine

russellhadley:
Set DataLayout on newly created module so that there's agreement between the TargetMachine and the Module.  This check is forthcoming in the LLVM merge.

@erozenfeld @adiaaida


erozenfeld:
LGTM

adiaaida:
LGTM

Issue #33, opened 2015-02-17
GC: reporting for GC references from non-SSA references


AndyAyersMS:
Value type GC fields need reporting, but don't naturally fit into SSA. Find a way to report them.

swaroop-sridhar:
Here's a summary of the discussion with @AndyAyersMS,  @preames  and @sanjoy on this topic:

**Supporting aggregates containing GC-pointers**
Pointers stored within aggregates on the stack need to be reported explicitly (reporting the entire aggregate won’t work.

*Philip Reams wrote:*
Thinking about this, I see a couple of approaches.  They divide into two ways of thinking about the problem.

Option 1 - Extend the stackmap format to encode an object map for the on stack aggregate.  Stackmap contains only the pointer to the alloca and a pointer to the appropriate object map.  The runtime can translate the object map + alloca entry into specific stack offsets for individual fields if desired.  

Option 2 - Report each field within the object directly in the stackmap.  No format extensions needed.  

For option 2, we could extend the lowering code to distinguish a GEP from an alloca from a "normal" SSA GEP.  We already have this distinction for alloca pointers themselves.  We might want to introduce a new argument section (for explicit slots?), but this could be prototyped quickly today.  I do have some concerns about distinguishing values reliably - i.e. consider a transform which adds a 100 nop GEPs - but in practice, I suspect this would work fine.  

For option 1, we could reuse the metadata argument to gc.root.  We could interpret a pointer to a alloca with a gc_root marker as being an explicit stack slot (i.e. no mem2reg performed), and use the metadata field to describe the object map.  We would need to either define a metadata format for the object map or find a generic way to serialize a metadata encoding into the stack map.  

In either case, we probably need to extend Mem2Reg to not treat a statepoint gc arg use as an escaping reference.  We want allocas to still become SSA in most cases.  

Anyone else have another approach?  Or thoughts on these?  I mildly tempted to take option 2 out of simple expediency, but I'm mildly concerned some GC might expect to be able to get object maps for the structs on the stack.  

*Swaroop wrote:*
 
My preference is also for option 2 – reporting each GC-pointer field individually in the stackmap – since it is closer to the CLR stackmap format.
For CLR, if we generate the stackmap per option 1, the LLILC JIT layer will have to translate it to the individual-pointer format.

*Andy Ayers Wrote:* 
The option 1 approach doesn’t work with the current CoreCLR. That is the runtime cant’ figure it out.

[...]

PR #740, opened 2015-08-05
Statepoint-Insertion: Add additional check for Managed aggregates.

swaroop-sridhar:
In Reader's post-process phase, add an additional check to
detect allocation of managed aggregates on the stack (alloca).
Conservatively defer to the default Jit for functions containing
such instructions.

Without this check, LLVM silently generated bad code
(until Issue #33 is fixed).

This change also clarifies a comment in GcInfo.cpp.


swaroop-sridhar:
@pgavlin  please review.

pgavlin:
LGTM.

PR #741, opened 2015-08-06
Handle more relocations

JosephTremoulet:
The lowering of the blockaddress intrinsic (which EH lowering will make
use of) invovles relocations in .text with type ST_Debug.  Change the
exclusion in LLILC's relocation processing to check for section name
.debug rather than relocation type ST_Debug.

Also, LLVM expects the data in the image to be added to the symbol address
when a relocation is applied, but the EE's recordRelocation simply
overwrites the data in the image.  Compensate by looking up the data in
the image and adding it to the target value passed to recordRelocation.


JosephTremoulet:
@erozenfeld PTAL

erozenfeld:
LGTM. Thank you for fixing this.

PR #742, opened 2015-08-07
Add support for building Roslyn (and other apps) with LLILC.

richardlford:
test/llilc_run.py is a general utility for running applications
with LLILC as the JIT. The test/llilc_run.pyproj  and
test/llilc_run.sln are useful for testing it from within
Visual Studio (requires Python Tools for Visual Studio).

test/llilc_roslyn_stage.cmd is a batch script that
will create a bootstrap stage directory for Roslyn
from the existing Roslyn binaries. It then builds
Roslyn using those binaries.

### DIFF
--- a/test/llilc_roslyn_stage.cmd
+++ b/test/llilc_roslyn_stage.cmd
@@ -1,0 +1,27 @@
+REM Usage: dostage.cmd workspace buildsubdir stagedir
+REM 
+REM Will first delete (if present) and then create %workspace%/roslyn/%stagedir%
+REM directory and set it up as the directory
+REM from which Roslyn will run on CoreClr, with LLILC as JIT.
+
+setlocal
+set WORKSPACE=%1
+set buildsubdir=%2
+set stagedir=%3
+cd %WORKSPACE%
+xcopy /y %buildsubdir%\bin\RelWithDebInfo\llilcjit.dll coreclr\bin\Product\Windows_NT.x64.Debug
+call "%VS140COMNTOOLS%..\..\VC\vcvarsall.bat" x86
+echo on
+cd roslyn
+if exist %stagedir%/ rd /s /q %stagedir%
+mkdir %stagedir%
+xcopy /S /Q Binaries\Debug\core-clr\* %stagedir%
+rename %stagedir%\csc.exe csc.dll
+copy /y %WORKSPACE%\coreclr\bin\Product\Windows_NT.x64.Debug\CoreConsole.exe %stagedir%\csc.exe
+set command=C:\Python34\python %WORKSPACE%\llvm\tools\llilc\test\llilc_run.py  --llilc-coreclr-runtime-path %WORKSPACE%\coreclr\bin\Product\Windows_NT.x64.Debug  --llilc-app-path %WORKSPACE%\roslyn\%stagedir%\csc.exe %%*
+echo %command% > %stagedir%\runcsc.cmd
+echo exit /b %%ERRORLEVEL%% >> %stagedir%\runcsc.cmd
+
+rd /s /q Binaries
+msbuild /m /v:d /p:CSCTOOLPATH=%WORKSPACE%\roslyn\%stagedir% /p:CSCTOOLEXE=runcsc.cmd src/Compilers/CSharp/CscCore/CscCore.csproj
+endlocal
--- a/test/llilc_run.py
+++ b/test/llilc_run.py
@@ -1,0 +1,105 @@
+﻿#!/usr/bin/env python
+#
+# title           : llilcrun.py
+# description     : Run a managed application using LLILC jit.
+#   
+
+import argparse
+import os
+import subprocess
+import sys
+
+llilcverbose = False
+
+def QuoteArg(arg):
+    '''Strip off any enclosing single quotes and enclose in double quotes'''
+    arg = '"' + arg.strip("'").strip('"') + '"'
+    return arg
+
+def UnquoteArg(arg):
+    ''' Remove single and double quotes from front and back of arg'''
+    return arg.strip("'").strip('"')
+
+def RunCommand(command):
+    ''' Run a command and return its exit code, optionally echoing it.'''
+    global llilcverbose
+    if llilcverbose:
+        sys.stderr.write (command + '\n')
+        sys.stdout.write (command + '\n')
+    error_level = subprocess.call(command, shell=True)
+    return error_level
+
+def main(argv):
+    '''
+    main method of script. arguments are script path and remaining arguments.
+    '''
+    global llilcverbose
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--llilc-dump-level', type=str, choices={'summary', 'verbose'}, 
+                        help='the dump level: summary, or verbose')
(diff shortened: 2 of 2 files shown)


richardlford:
An example of a successful lab bootstrap build of Roslyn on CoreClr with LLILC is
http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_roslyn_test/7.

@adiaaida, @erozenfeld,  please take a look.

test/llilc_roslyn_stage.cmd
```diff
@@ -0,0 +1,25 @@
+REM Usage: dostage.cmd workspace buildsubdir stagedir
+REM 
+REM Will first delete (if present) and then create %workspace%/roslyn/%stagedir%
+REM directory and set it up as the directory
+REM from which Roslyn will run on CoreClr, with LLILC as JIT.
+
```
adiaaida: can we start with setlocal and end with endlocal, or is the goal to have the environment variables permanently set?

test/llilc_run.py
```diff
@@ -0,0 +1,105 @@
...
+    error_level = subprocess.call(command, shell=True)
+    return error_level
+
+def main(argv):
+    '''
+    main method of script. arguments are script path and remaining arguments.
+    '''
+    global llilcverbose
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--llilc-dump-level', type=str, choices={'summary', 'verbose'},
```
adiaaida: If all of these arguments are prepended with llilc, can we strip llilc off of them? So just --dump-level, --extra, --ngen, etc? Also, can we have short forms for them, like -d for dump-level, -e for extra, -n for ngen, etc?

test/llilc_run.py
```diff
@@ -0,0 +1,105 @@
...
+    global llilcverbose
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--llilc-dump-level', type=str, choices={'summary', 'verbose'}, 
+                        help='the dump level: summary, or verbose')
+    parser.add_argument('--llilc-extra', type=str, default=[], nargs='*',
+                        help='''list of extra COMPlus settings. Each item is Name=Value, where
+                                Name does not have the COMPlus_AltJit prefix.
+                             ''')
+    parser.add_argument('--llilc-ngen', help='use ngened mscorlib', default=False, action="store_true")
+    parser.add_argument('--llilc-run-verbose', help='echo commands', default=False, action="store_true")
```
adiaaida: I think --verbose/-v would be fine here

adiaaida:
In your jenkins job, there is actually a windows powershell option for build steps, so there's no need to call powershell from batch. You can just use the powershell build step and call the powershell script.

adiaaida:
Other than the few things I've mentioned, LGTM.

test/llilc_run.py
```diff
@@ -0,0 +1,105 @@
...
+        command += ' ' + arg
+    error_level = RunCommand(command)
+    if llilcverbose:
+        sys.stdout.write ('Exiting llilc_run.py with exit code ' + str(error_level) + '\n')
+        sys.stderr.write ('Exiting llilc_run.py with exit code ' + str(error_level) + '\n')
+    return error_level
+
+if __name__ == '__main__':
+    return_code = main(sys.argv[1:])
+    sys.exit(return_code)
```
russellhadley: Finish with a newline.  So we don't have github warning us with the little red glyph above.

test/llilc_run.pyproj
```diff
@@ -0,0 +1,50 @@
...
+  <Import Condition="!Exists($(PtvsTargetsFile))" Project="$(MSBuildToolsPath)\Microsoft.Common.targets" />
+  <!-- Uncomment the CoreCompile target to enable the Build command in
+       Visual Studio and specify your pre- and post-build commands in
+       the BeforeBuild and AfterBuild targets below. -->
+  <!--<Target Name="CoreCompile" />-->
+  <Target Name="BeforeBuild">
+  </Target>
+  <Target Name="AfterBuild">
+  </Target>
+</Project>
```
russellhadley: Same newline comment as above.

russellhadley:
I'm not sure how I feel about adding the project and solution files for the python script.  Seems to go against our "everything cross-platform" ethos for not a lot of benefit - since they're Windows/VS specific.  (Python already has a integrated interactive debugger module that runs cross platform) I'm not particularly passionate about not checking it in, but it does seem like avoidable cruft - how do you feel @richardlford about leaving these files out?

test/llilc_run.py
richardlford: I did that because the unrecognized options are passed on to the program that is being run
and I did not want to accidentally process the options belonging to that 
program. That is also why I excluded short form options as they are very
likely to class with the program being run.

test/llilc_roslyn_stage.cmd
richardlford: I'm fine with adding setlocal and endlocal. There is no intention to set environment variables persistently.

richardlford:
Regarding powershell, the one place where I use it is to invoke the NuGetRestore.ps1
script. The problem with making that whole build step powershell is that there
is not a simple way to invoke vcvarsall.bat from powershell. You can invoke
it, but it does not change the environment variables in the powershell script.
So for example msbuild will not be found on the path.

richardlford:
Regarding including the project and solution file for the python script, I don't feel strongly for my own benefit. The Python Tools for Visual Studio (PTVS) has a handy option to create a project and solution
from existing python code. So I can easily recreate them for my own benefit.

However I find using PTVS extremely useful and the using the existing python debugger manually
clunkly by comparison. So though our emphasis is on cross-platform, if a user is developing
on windows they will be missing a lot by not using PTVS. But we don't need to be the
ones to evangelize it, so I'm fine with not checking these in.

So I will remove them.

test/llilc_run.py
richardlford: Done.

test/llilc_run.pyproj
richardlford: OK, but I'm not going to be checking this in. I'm surprised the PTVS  does not include a trailing newline.

erozenfeld:
LGTM

test/llilc_run.py
adiaaida: I'm not sure I like that scheme. In fact, looking at the script, I didn't realize that was happening, suggesting that it is not really intuitive. I would be more comfortable if any arguments being passed on to the program being run were explicitly passed. Like having a --program-args or --app-args option that took a semicolon separated list of options and then parsing those args and sending them to the program being called.

richardlford:
I found an argparse feature that can be used to prevent option clash. If '--' is given on 
the option line then options past that are not processed by argparse (and so can be 
passed on to the client program). I'm going to revise llilc_run.py to use this feature,
so we can shorten the options and allow the short forms. But I will do that
in a subsequent check-in. So I'll merge this for now.

test/llilc_run.py
richardlford: I didn’t see this comment until after I merged, but as I said in a separate message, I think using ‘--' to separate the script args from the client args is the best option as it does not involve adding an extra parsing step. In any case, we can discuss that in the follow-on version.

PR #744, opened 2015-08-08
GcInfo: Fix Offsets for NGEN

### LINKED ISSUE #743: GC: Fix instruction offsets in GCTables  for NGEN
For NGEN and AOT scenarios, the Function entry address is relative to start of file (starts from zero). GcInfo's encodeLiveness() expects this as an offset from the allocated code block.

Fix the difference offset so that instruction offsets reported at  safepoints are correct.

swaroop-sridhar:
GcInfoEncoder needs to find the distance (offset) of the
Function entry point to the start of the code section.
For this, it calculated the diffrence between the HotCodeBlock
and the function address reported in __llvm_stackmaps.

However, in an Ngen/AOT build, the FunctionEntry address
recorded in LLVM's StackMap are post-relocation offsets.
Therefore, it cannot be compared to the Start of the CodeBlock
allocated during compilation.

This change fixes this bug by using non-relocated code-block
address for both function-entry and code-block-start.

Fixes #743

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -312,9 +312,11 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
                << ", size = " << *NativeSizeOfCode
                << " method = " << Context.MethodName << '\n';
       }
+
+      assert(*NativeEntry >= MM.getHotCodeBlock());
       GcInfoAllocator GcInfoAllocator;
-      GCInfo GcInfo(&Context, MM.getStackMapSection(), MM.getHotCodeBlock(),
-                    &GcInfoAllocator);
+      GCInfo GcInfo(&Context, MM.getStackMapSection(), &GcInfoAllocator,
+                    *NativeEntry - MM.getHotCodeBlock());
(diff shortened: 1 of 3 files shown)


swaroop-sridhar:
@pgavlin: Please review

lib/Jit/LLILCJit.cpp
```diff
@@ -313,8 +313,8 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
                << " method = " << Context.MethodName << '\n';
       }
       GcInfoAllocator GcInfoAllocator;
-      GCInfo GcInfo(&Context, MM.getStackMapSection(), MM.getHotCodeBlock(),
-                    &GcInfoAllocator);
+      GCInfo GcInfo(&Context, MM.getStackMapSection(), &GcInfoAllocator,
+                    *NativeEntry - MM.getHotCodeBlock());
```
russellhadley: Should there be some kind of assert that we always get a positive number for the offset?

russellhadley:
Other than the question about future proofing with an assert this LGTM.

pgavlin:
LGTM.

PR #752, opened 2015-08-10
Improve usability of llilc_run.py

richardlford:
llilc_run.py is used to run managed applications using LLILC
as the JIT. llilc_run.py takes command line arguments and the
managed application being run may also take command line
arguments. We must avoid conflicts between these command line
arguments.

The previous version of llilc_run.py attempted to avoid
conflicts between the arguments supplied to llilc_run.py and
the applications command line arguments by 
making llilc_run.py's command line arguments all start with
'llilc-' and by avoiding abbreviated forms. This made
the script less convenient to use and also did not provide
an absolute guarantee of no conflicts.

This change makes use of a common feature of open source
programs that invoke other programs. That feature is using
'--' to separate the parameters of the invoking program
from the parameters of the invokee. The Python argparse
library supports this feature. So we use '--' to separate
the llilc_run.py parameters from the parameters of the
managed program being run.

Along with the change the llilc_run.py parameters have
had the 'llilc-' prefix removed and one letter abbreviations
have been added. Some other small tweaks to the
script have been added.

- If the -j option is given, the LLILC jit at the given
  path is copied into the CoreClr binary directory.
  If not given we check that a LLILC jit has been previously
  copied to that location.

- Support for precise GC testing is added (-p option)

### DIFF
--- a/test/llilc_roslyn_stage.cmd
+++ b/test/llilc_roslyn_stage.cmd
@@ -1,8 +1,9 @@
-REM Usage: dostage.cmd workspace buildsubdir stagedir
+REM Usage: llilc_roslyn_stage.cmd workspace buildsubdir stagedir
 REM 
 REM Will first delete (if present) and then create %workspace%/roslyn/%stagedir%
 REM directory and set it up as the directory
 REM from which Roslyn will run on CoreClr, with LLILC as JIT.
+REM buildsubdir is the subdirectory of the workspace where LLVM (and LLILC) were built.
 
 setlocal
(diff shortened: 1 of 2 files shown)


richardlford:
@adiaaida, please take a look. Locally this has successfully run a Roslyn bootstrap. 
I've also started a lib job: 

http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_roslyn_test/11/

adiaaida:
LGTM, assuming testing is green.

PR #761, opened 2015-08-13
Allow int32 to int64 comparisons in branches

### LINKED ISSUE #637: Assertion: size mismatch between comparands provided to a conditional branch
This pops up during a large number of the new IL tests.

`Assertion failed: (Size1 == Size2) && "Size1 == Size2", file ..\..\..\..\..\tools\llilc\lib\Reader\readerir.cpp, line 6363`

richardlford:
LLILC was already allowing int32 to int64 comparisons
for the opcodes that produce a boolean result.
But the code that adjusted comparison arguments was
only partly duplicated for branch comparisons.

This change factors out the common code that
adjusts comparison arguments so that boolean result
comparison and branch comparison are treated consistently.

With this change test
JIT\Regression\CLR-x86-JIT\V1-M13-RTM\b91223,
which was failing with an assertion failure is now
able to compile the ILGEN_0x9c0b37ec.Method_0xa1f09b04
method. The test still fails later, because the tests
correct execution throws an overflow exception which
LLILC is not yet prepared to catch.

This change also includes two other small changes:

The MSIL dumps were being written directly to stderr
rather than going through dbgs() and I've observed
that this output does not appear when called from
the debugger. To fix this I've changed the
dbprint method to first format the message into a
string and then output that string to dbgs().
Currently we only use dbPrint to write MSIL
dumps. I allowed a buffer size of 200 characters,
but also used the safe string formatting method (vsnprintf).

In the llilc_run.py file, when verbose output is being
produced I modified the log function so that output to
stdout and stderr were tagged to avoid confusion,
as when they have the save destination the duplication
can be confusing.

Resolves #637.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -319,6 +319,21 @@ public:
   IRNode *ckFinite(IRNode *Arg1) override {
     throw NotYetImplementedException("ckFinite");
   };
+
+  /// \brief Modify comparison arguments, if needed, to have equal types.
+  ///
+  /// LLVM comparisons require that the operands have the same type
+  /// whereas MSIL operands may have different type, subject to
+  /// the conditions in the ECMA spec, partition III, table 4,
+  /// except we relax these conditions somewhat by allowing comparison
+  /// of int64 with int32 or native int. Promote the arguments
+  /// as needed to achieve type equality.
+  ///
+  /// \param [in,out] Arg1 Pointer to first argument.
+  /// \param [in,out] Arg2 Pointer to second argument.
+  /// \returns True if the comparison is a floating comparison.
+  bool prepareArgsForCompare(IRNode *&Arg1, IRNode *&Arg2);
+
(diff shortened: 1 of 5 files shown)


richardlford:
@erozenfeld, please take a look.

erozenfeld:
LGTM

richardlford:
I pushed a revision with a formatting fix and changed the issue number.

richardlford:

richardlford:
In job http://dotnet-ci.cloudapp.net/job/dotnet_llilc_prtest/967/console all but the formatting leg
passed. I corrected that by removing one trailing blank in a comment line.
In job http://dotnet-ci.cloudapp.net/job/dotnet_llilc_prtest/968/console the formatting leg passed
and all the others except for the 
http://dotnet-ci.cloudapp.net/job/dotnet_llilc_release_win32_prtest/867/
leg which failed because of an apparent network problem (clone failed). Since
that leg previously passed with sources the same except for a blank in comments,
I'm going to go ahead and merge this since I don't want to wait up for the
retest to complete.

PR #764, opened 2015-08-14
GCTables: Report StackBasePointer.

### LINKED ISSUE #758: GC: Fix StackBase reporting in GcTables
The FP/SP stack base should be reported in the header in addition to the slots representing liveness.

swaroop-sridhar:
CoreCLR requires that we identify whether a method's frame is
FP or SP based. This is inaddition to reporting the reference pointers
for the slots representing liveness. So, this change reports the
StackBasePointer in the GcTable header.

To facilitate the above, the GcInfo encoding functions are
restructured to be per-function routines.

Fixes #758.

### DIFF
--- a/include/GcInfo/Target.h
+++ b/include/GcInfo/Target.h
@@ -17,10 +17,19 @@
 #define GCINFO_TARGET_H
 
 #include "global.h"
+#include "corinfo.h"
 
 #if (defined(_TARGET_X86_) || defined(_TARGET_X64_) || defined(_TARGET_AMD64_))
 
-// Define DWARF encodings for registers
+// Identify the frame-pointer register number
+
+#if defined(_TARGET_X86_)
+#define REGNUM_FPBASE ICorDebugInfo::RegNum::REGNUM_EBP
+#else
+#define REGNUM_FPBASE ICorDebugInfo::RegNum::REGNUM_RBP
+#endif // defined(_TARGET_X86_)
+
+// Define encodings for DWARF registers
 // Size variants (ex: AL,AH,AX,EAX,RAX) all get the same Dwarf register number
 
 #define DW_RAX 0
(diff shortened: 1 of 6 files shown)


swaroop-sridhar:
@russellhadley: please review.

include/GcInfo/Target.h
```diff
@@ -20,7 +20,46 @@
 
 #if (defined(_TARGET_X86_) || defined(_TARGET_X64_) || defined(_TARGET_AMD64_))
 
-// Define DWARF encodings for registers
+// Define encodings for machine registers
+
```
russellhadley: Is this duplication of code that is in coreclr?  This looks an awful lot like a header that should be coming with the gcinfo lib.

russellhadley:
Other than the header question LTGM.

include/GcInfo/Target.h
swaroop-sridhar: OK, I'll change this to use the register numbers defined in ICorDebugInfo.

include/GcInfo/Target.h
```diff
@@ -17,10 +17,19 @@
 #define GCINFO_TARGET_H
 
 #include "global.h"
+#include "corinfo.h"
 
 #if (defined(_TARGET_X86_) || defined(_TARGET_X64_) || defined(_TARGET_AMD64_))
 
-// Define DWARF encodings for registers
+// Identify the frame-pointer register number
```
swaroop-sridhar: @russellhadley: This is the main change in the update -- use register number definitions from ICorDebugInfo.
The original definitions in Target.h were copied from RYU Jit. 
I checked that those register enumerations are the same as the ones in ICorDebugInfo.

swaroop-sridhar: It would be good to have a ICorTargetInfo, containing target specific definitions such as registers etc., instead of picking it up from the ICorDebugInfo.

russellhadley:
Open an issue to figure out out to reconcile the header issue with RyuJIT and go a head and merge.

swaroop-sridhar:
I opened #770 for the target-specific headers issue.

Issue #593, opened 2015-05-21
Debugger: Debug info for locals


adiaaida:
Add debug info support for locals

russellhadley:
@adiaaida Move this out to S87 and add any notes.

russellhadley:
@adiaaida can you add some notes on where we're at with this?

adiaaida:
* Have added logic in LLILC reader to create dbg.declare intrinsics for locals.
* declare intrinsics are currently being dropped in LLVM. tracking down where this is happening and why/what happens with clang.

adiaaida:
* Added logic to LLILCJit to extract locals info from the dwarf debug info
* Added logic to pass debug info to CLR EE.
* Opened in debugger and we are seeing accurate locals info.

* Discovered that stepping is broken. It appears this is unrelated to the locals support. I am opening an issue.

russellhadley:
@adiaaida this will be held in S87.  I'd like to get this in as soon as we can.

adiaaida:
We currently cannot run with my changes with zap disabled. Need to investigate.

PR #765, opened 2015-08-17
Add debug support for locals

adiaaida:
This change add support to LLILC to send local variable information to
LLVM and then on to the CLR EE to be used in debugging. This change
adds:

* Support to the Reader to capture local variables as they are created.
  We copy the logic from clang to specify if a local is a formal
  parameter or a regular local variable, though the CLR does not seem to
  care about this distinction. We assume that the local is in scope for
  the full function.
* Update convertType to deal with struct types. Since we don't use the
  type information for debugging, we can create a debug struct type with
  no real information about the struct. This change is really just there
  to prevent LLVM from attempting to reference a nullptr.
* Support to the ObjectLoadListener to extract DWARF information from
  the ObjectFile for local variables. Unlike line numbers, there is no
  built in support for extracting local variable information, so we have
  to add support to LLILC for this purpose. The extraction works much
  the same way as the DWARF info dumper: we extract the Unit
  DebugInfoEntry (DIE), find the Subprogram DIE -- DWARF's name for a
  function -- and then extract the locals that are in the subprogram.
  The DWARF format stores the frame_base register in the Subprogram
  DIE, so we must extract that and convert the DWARF register name to
  the value the CLR EE understands. Each of the locals can be found at
  an offset from the value in the frame_base register, and that offset
  is the last entry in the DW_AT_location attribute. For each local, we
  must then communicate back the EE the local number (assumed to be in
  order by creation in the DWARF info), the frame_base register and the
  offset. For now, we assume that all locals are found on the stack,
  though this may change, and will need future updates. If there were no
  locals in a function, we do not report anything back to the EE for
  that function.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -27,6 +27,7 @@
 #include "llvm/Config/llvm-config.h"
 #include "llvm/DebugInfo/DIContext.h"
 #include "llvm/DebugInfo/DWARF/DWARFContext.h"
+#include "llvm/DebugInfo/DWARF/DWARFFormValue.h"
 #include "llvm/ExecutionEngine/Orc/CompileUtils.h"
 #include "llvm/IR/DataLayout.h"
 #include "llvm/Support/DataTypes.h"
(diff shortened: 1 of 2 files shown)


adiaaida:
@russellhadley PTAL

adiaaida:
Please note: if this goes in before the LLVM FI, then we will need to update the reader to use createParameterVariable and createAutoVariable. If the FI goes in first, this change will need to be updated.

adiaaida:
failed formatting. Updated in place.

adiaaida:
updated to use the new functions for creating local variables for debug info.

russellhadley:
Green.

adiaaida:
This closes #593

PR #769, opened 2015-08-18
No longer exclude tests now passing as a result of #637 fix.

richardlford:
After the fix to #637 forty more tests now pass, so remove
these from the exclusion.targets. In additional one other
exclusion (JIT\Directed\coverage\oldtests\arrgetlen.cmd)
was for a non-existent test so also removed it.

Checked compiler testing passed locally with this updated
exclusion.targets list.


richardlford:
@russellhadley, PTAL.

russellhadley:
LGTM

PR #771, opened 2015-08-18
Add link to clang-format.exe to ccformat

adiaaida:
If formatting fails, we would like to have a link to the
clang-format.exe that is being used in the lab so users can easily find
what their code will be tested against.


adiaaida:
@russellhadley PTAL

PR #772, opened 2015-08-19
Additional LLILC debugging aids

richardlford:
The changes makes the following improvements to the
LLILC debugging facilities:

1. A make_sln.py script is added which can be used to create
a Visual Studio 2015 solution that can be used to debug
a managed application using LLILC. Optionally the
script can invoke Visual Studio on the created solution.
The handles setting up all the required environment
variables for making LLILC the alternate JIT.

2. A --windbg-and-args option was added to the
llilc_run.py script to enable debugging the
application being run using windbg.

3. In order to ensure that empty values for
LLILC configuration variables are equivalent to
unspecified environment variables, the
code that LLILC uses to check some configuration
variables was refactored to require that
the feature is only enabled if the variable
is both non-null and non-empty.

4. The Debugging.md documentation was updated
to describe these features.

5. The llilc_run.py and make_sln.py were modified so they
run successfully on either Python 2.7 or Python 3.4.

### DIFF
--- a/test/make_sln.py
+++ b/test/make_sln.py
@@ -1,0 +1,304 @@
+﻿#!/usr/bin/env python
+#
+# title           : make_sln.py
+# description     : Make a Visual Studio solution to debug a managed 
+#                   application using LLILC jit.
+#
+# This script has been tested running on both Python 2.7 and Python 3.4.
+#
+# usage: make_sln.py [-h] [-d {summary,verbose}] [-x [EXTRA [EXTRA ...]]] [-n]
+#                    [-p] [-v] [-r [CORERUN_ARGS [CORERUN_ARGS ...]]]
+#                    [-j JIT_PATH] [-s SLN_PATH] [-i] [-f] -a APP_PATH -c
+#                    CORECLR_RUNTIME_PATH
+# 
+# Make a solution to debug a managed application with LLILC as the JIT. If the
+# application has any arguments, they must be appended to the end of the command
+# line, preceded by "--".
+# 
+# optional arguments:
+#   -h, --help            show this help message and exit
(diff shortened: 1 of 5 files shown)


richardlford:
@adiaaida, PTAL.

test/make_sln.py
```diff
@@ -0,0 +1,294 @@
...
+        vspath = os.path.abspath(os.path.join(vstools, "..", "IDE", "devenv.exe"))
+        command = "start"
+        command += ' '
+        command += QuoteArg(vspath)
+        command += ' '
+        command += QuoteArg(sln_path)
+        error_level = RunCommand(command)
+
+    if llilcverbose:
+        log ('Exiting llilc_run.py with exit code ', error_level)
```
adiaaida: Is this supposed to be make_sln.py, not llilc_run.py?

adiaaida:
Other than the one nit, looks fine to me.

test/make_sln.py
richardlford: You are right. I'll fix that.

richardlford:
I squashed and pushed the log message.

adiaaida:
We are going to be switching to python 2.7 as the default version of Python for LLILC. Can you make sure this runs appropriately when run with 2.7?

richardlford:
I modified make_sln.py as well as llilc_run.py so they run successfully on both Python 2.7 and Python 3.4. I also added an additional documentation not about the debugger type in the solutions.

Since the previous version passed testing
(http://dotnet-ci.cloudapp.net/job/dotnet_llilc_prtest/990/)
and this change only changed python scripts that are not used lab tests, I'm
not going to wait for the lab job to complete before merging.

Issue #768, opened 2015-08-18
GC: Report GS Cookie


swaroop-sridhar:
Once #353 is implemented,
Report the GSCookie slot in the GcInfo encoding.
For such methods, the size of both prolog and epilog(s) must be reported as well.

Issue #767, opened 2015-08-18
GC: Report SecurityObjectStackSlot


swaroop-sridhar:
Report the stack slot of the security object when a security object is inserted by  `GenIR::insertIRForSecurityObject()`

For such methods, the prolog size must also be reported to the GcInfo encoder.

AndyAyersMS:
See also #301, which I left open for this.

Issue #766, opened 2015-08-18
GC: Report GenericsInstContextStackSlot


swaroop-sridhar:
Report the generic instantiation context stack slot to the GC Encoding, when we generate the slot in `GenIR::insertIRToKeepGenericContextAlive()`

For such methods, the prolog size must also be reported.

AndyAyersMS:
See also #40, which I left open for this.

Issue #353, opened 2015-03-24
Implement stack protection checks


AndyAyersMS:
Methods with "unsafe" value types on the stack or locallocs trigger the need for stack protection. We currently don't look for the former and haven't implemented the latter.

This is similar to the /GS used in native code. A cookie is placed on the stack between the locals and the return address. The locals are reordered to put vulnerable locals near the cookie. Then the cookie is initialized in the prolog and checked at the return points (and likely reported to the unwinder too).

PR #774, opened 2015-08-20
Precise GC: Add NYIs for functions with missing info

swaroop-sridhar:
GcInfo encoder doesn't yet report certain data in the
GCInfo headers.

1) Generic Context slot: #766
2) Security Object Slot: #767
3) GSCookie slot: #768

This change adds exclusions for functions containing (1) and (2)
when compiling for Precise GC.

There is no exclusion for (3) because stack protection checks are
yet to be implemented in LLILC #353


swaroop-sridhar:
@russellhadley please review

russellhadley:
Green.

PR #775, opened 2015-08-20
﻿Remove SafepointPoll from the Module after Safepoint placement.

swaroop-sridhar:
This change removes the SafepointPoll function from the module
once the Safepoint placement and rewriting Phases are run.
There will be no references to this function after the
PlaceSafepoints phase is run -- since this phase inlines the contents
of the function SafepointPoll() at all polling locations.

This change will improve throughput, since machine lowering sees
one less function in every module. We'll get this optimization for free
once we turn on the LLVM inliner via Inliner::removeDeadFunctions().

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -23,6 +23,7 @@
 #include "abi.h"
 #include "EEMemoryManager.h"
 #include "EEObjectLinkingLayer.h"
+#include "llvm/Analysis/CallGraph.h"
 #include "llvm/CodeGen/GCs.h"
 #include "llvm/Config/llvm-config.h"
 #include "llvm/DebugInfo/DIContext.h"
@@ -290,6 +291,18 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
 
         Passes.add(createRewriteStatepointsForGCPass(false));
         Passes.run(*M);
+
+        // PlaceSafepoints phase inlines the contents of gc.safepointPoll()
+        // at all polling locations. So, once the phase has run, remove the
+        // @gc.safepointPoll() function from the module.
+        Function *SafepointPoll = Context.SafepointPoll;
+#if !defined(NDEBUG)
+        // Ensure that there are no more callers for SafepointPoll function.
+        CallGraph CG(*Context.CurrentModule);
+        CallGraphNode *PollNode = CG[SafepointPoll];
+        assert(PollNode->getNumReferences() == 0);
+#endif
+        SafepointPoll->eraseFromParent();
       }
 
       // Use a custom resolver that will tell the dynamic linker to skip
--- a/include/Jit/LLILCJit.h
+++ b/include/Jit/LLILCJit.h
@@ -81,6 +81,7 @@ public:
   llvm::LLVMContext *LLVMContext; ///< LLVM context for types and similar.
   llvm::Module *CurrentModule;    ///< Module holding LLVM IR.
   llvm::TargetMachine *TM;        ///< Target characteristics
+  llvm::Function *SafepointPoll;  /// < The Safepoint Poll helper
   bool HasLoadedBitCode;          ///< Flag for side-loaded LLVM IR.
   llvm::StringMap<uint64_t> NameToHandleMap; ///< Map from global variable names
                                              ///< to the corresponding CLR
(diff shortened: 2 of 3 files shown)


swaroop-sridhar:
@russellhadley : please review

lib/Jit/LLILCJit.cpp
```diff
@@ -290,6 +290,16 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
...
         Passes.add(createRewriteStatepointsForGCPass(false));
         Passes.run(*M);
+
+        // Once PlaceSafepoints phase is run, there will be no references
+        // to the gc.safepointPoll() function. PlaceSafepoints phase inlines 
+        // the contents of gc.safepointPoll() at all polling locations.
+        //
+        // Therefore, remove the function from the Module. 
+        // We'll get this optimization for free once we turn on the 
+        // LLVM inliner.
```
russellhadley: Are we sure the checkers will catch if this isn't the case?  If there still is a reference to the safepointPoll?  I'm just wondering if we need an extra assert here.  Thoughts?

lib/Jit/LLILCJit.cpp
```diff
@@ -290,6 +291,18 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
 
         Passes.add(createRewriteStatepointsForGCPass(false));
         Passes.run(*M);
+
+        // PlaceSafepoints phase inlines the contents of gc.safepointPoll()
+        // at all polling locations. So, once the phase has run, remove the
+        // @gc.safepointPoll() function from the module.
+        Function *SafepointPoll = Context.SafepointPoll;
+#if !defined(NDEBUG)
```
swaroop-sridhar: @russellhadley: I added a check to verify that there are indeed no callers for the polling function.

swaroop-sridhar:
@dotnet-bot  test this please

swaroop-sridhar:
@dotnet-bot test this please

swaroop-sridhar:
All failure were "error MSB4175: The task factory "XamlTaskFactory" could not be loaded from the assembly  ... "

swaroop-sridhar:
@dotnet-bot test this please

swaroop-sridhar:
@dotnet-bot test this please

adiaaida:
@dotnet-bot test this please

lib/Jit/LLILCJit.cpp
swaroop-sridhar: This assertion (that no references to SafepointPoll remain after PlaceSafepoint phase) failed for some of the new GC tests (not yet added in the CoreCLR test tree). 

The correct way to remove the unreachable functions is to invoke the Inliner (Inliner::removeDeadFunctions()).
However, we shouldn't have to run this optimization in /Od builds.

In order to get the lab job (InsertStatepoints nightly) green, I'll add a workaround to remove trivially unreachable SafepointPolls.

AndyAyersMS:
@swaroop-sridhar I think there may also be tricks we can play to suppress processing the helper during codegen without needing to delete the function. Something like setting `IsMaterializable` false.

swaroop-sridhar:
@AndyAyersMS: I tried setting the isMaterializable attribute, but it did not help solve the problem. I also tried setting the AlwaysInline attribute, which also did not work.

So, I think we should directly fix the problems caused by the SafepointPoll function, rather than trying to remove it. The two known problems are:
1) Debug Locals (I requested @adiaaida  to put in a workaround for this): https://github.com/dotnet/llilc/pull/783 
2) Registering unwind tables -- I'm working on a fix for this.

JosephTremoulet:
@swaroop-sridhar, the EH branch has code for parsing the EH tables in a module with multiple functions.  You'd have to get rid of the bit that's expecting markers between entries and just use the logic that figures it's done when it reaches the end of the .xdata section.  Look at [getXdataSize[(https://github.com/dotnet/llilc/blob/10aa3cff6d21036ccd0a9337bdc893e14a2ef4cf/lib/Jit/EEMemoryManager.cpp#L107) and its [two](https://github.com/dotnet/llilc/blob/10aa3cff6d21036ccd0a9337bdc893e14a2ef4cf/lib/Jit/EEMemoryManager.cpp#L166) [callers](https://github.com/dotnet/llilc/blob/10aa3cff6d21036ccd0a9337bdc893e14a2ef4cf/lib/Jit/EEMemoryManager.cpp#L260).  The difference is that the xdata you'll see won't have those three dwords (`IsLast`, `StartOffset`, `EndOffset`) between them.

JosephTremoulet:
Ah, ok, just saw #784, that seems like a perfectly reasonable approach here.

swaroop-sridhar:
Thanks @JosephTremoulet. 
Yes,  #784 will help circumvent the problem for SafepointPolls, and we can let the multi-function version of registerEHFrames() merge in from EH branch.

PR #777, opened 2015-08-20
Don't assume start address is 0 for local debug info

adiaaida:
We should not assume that the function is starting at address 0. Since
we do grab that info, we should make it explicity that locals are in
scope from the starting address of the function through the starting
address + size.


adiaaida:
@russellhadley PTAL

russellhadley:
Green

PR #778, opened 2015-08-20
Update LLILC ORC Compiler

russellhadley:
ORC Compiler functor was using a API that had been deleted in LLVM.  Call is no longer needed and so is removed.


russellhadley:
@adiaaida take a look.  (but this is a trivial change)

adiaaida:
LGTM

Issue #13, opened 2015-02-13
Implement Exception Handling Support


JosephTremoulet:
What's currently implemented is a stub to allow testing programs that may statically include EH constructs, but don't dynamically raise exceptions.  We will of course need a functional implementation; this is an overall tracking issue for that.

The tree contains a [document](https://github.com/dotnet/llilc/blob/master/Documentation/llilc-jit-eh.md) describing the overall status/plan for exception handling.

Current status:
1. [x] Stub EH support
   - [x] Reader discards catch/filter/fault handlers
   - [x] Explicit throw becomes helper call
   - [x] Continuation passing for finally handlers invoked by `leave` (#260)
   - [x] Implicit exceptions expanded to explicit test/throw sequences
     - [x] Null dereference (#63)
     - [x] Divide by zero (#64)
     - [x] Arithmetic overflow (#65)
     - [x] Convert with overflow (#48)
     - [x] Array bounds checks (#187/#188)
     - [x] Array store checks (#188)
2. [ ] Handler bring-up in EH branch
   - [ ] Catch handler support
     - [x] In reader (includes updating throws to use `invoke` rather than `call` with EH edge to `landingpad`) (#66)
     - [ ] Funclet prolog/epilog generation, Previous Stack Pointer Symbol handshake implemented (#67)
     - [x] EH Clause generation (#68)
     - [x] Reporting funclets back to EE with parent functions in `.text` (#69)
   - [x] Support for `rethrow` (#70)
   - [ ] Finally handler support
     - [ ] In reader (includes `leave` processing, continuation selection) (#71)
     - [x] EH Clause generation (#73)
   - [ ] Filter handler support
     - [ ] In reader (includes early outlining) (#74)
     - [ ] EH Clause generation (#75)
   - [x] Fault handler support (#76)
3. [x] Migrate changes back into master branch
4. [ ] EH-specific optimizations
   - [ ] Finally cloning
   - [ ] Others TBD
5. [ ] Support for other target architectures
6. [ ] Support for ahead-of-time compilation

PR #788, opened 2015-08-25
For former #637 tests that throw exceptions, change to issue #13

richardlford:
Some of the failing excluded tests that formerly were excluded
because of issue #637 still fail even though issue #637 has
been fixed, because they throw exceptions. Change the issue
number for these to #13. I verified in the debugger that
these tests do in fact throw an exception (when compiled with
Ryujit on the CoreClr).

In addition this change deletes some exclusions for tests that
no longer exist. Also one test had "13(" as its issue, so I
fixed that so it is "13".


richardlford:
@AndyAyersMS, PTAL.

richardlford:
@dotnet-bot, retest this please.

richardlford:
This passed locally except for the known 6 failures. Russ approved this.

Issue #776, opened 2015-08-20
Fix Debug Locals with InsertStatepoints


adiaaida:
It appears that the checkin for local debug info broke the insert statepoints job in the lab:

http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_insert_statepoints_nightly/103/console

Assert is in CoreCLR:
Assert failure(PID 10444 [0x000028cc], Thread: 12668 [0x317c]): m_pNativeVarInfo == NULL && m_iNativeVarInfo == 0\r\n\r\nCORECLR! CEEJitInfo::setVars + 0x1A3 (0x00007ff8`27ac5adf)\r\nLLILCJIT1440061588! ObjectLoadListener::getDebugInfoForLocals + 0x213 (0x00007ff8`24a85423)\r\nLLILCJIT1440061588! ObjectLoadListener::getDebugInfoForObject + 0x2DB (0x00007ff8`24a857db)\r\nLLILCJIT1440061588!

adiaaida:
Insert statepoints inserts a stub function that we should not, but are, setting variables for the stub function. We need to identify that we are looking at the stub function so we do not set vars for that.

adiaaida:
closed with #790

PR #790, opened 2015-08-25
Make sure subprogram addr and function addr match

adiaaida:
For communicating debug locals to the CLR EE, we want to make sure that
the starting address for the current function matches that of the
Subprogram DIE. In the case where we have multiple function symbols in
one module (GC Statepoints, funclets), we want to make sure that we are
passing to the EE locals that belong to the actual function (ie, not the
gc.safepoint_poll function). This change extracts the Subprogram DIE's
low pc (start address) and compares it to the symbol's start address. If
the two are not the same, it returns without collecting local debug
info.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -693,6 +693,18 @@ void ObjectLoadListener::getDebugInfoForLocals(
       FrameBaseRegister = mapDwarfRegisterToRegNum(FormValues->back());
     }
 
+    if (SubprogramDIE->getAttributeValue(CU.get(), dwarf::DW_AT_low_pc,
+                                         FormValue)) {
+      Optional<uint64_t> FormAddress = FormValue.getAsAddress(CU.get());
+      
+      // If the Form address doesn't match the address for the function passed
+      // do not collect debug for locals since they do not go with the current
+      // function being processed
+      if (FormAddress.getValue() != Addr){
+        return;
+      }
+    }
+
     std::vector<uint64_t> Offsets;
     extractLocalLocations(CU.get(), SubprogramDIE, Offsets);
 


adiaaida:
@swaroop-sridhar @russellhadley PTAL

This should close #776.

swaroop-sridhar:
Green for the fix.

Is this change intended to be temporary (to fix #776), or permanant?
When we have funclets, how is the debug info expected to be collected at load time? 
Will getDebugInfoForLocals() be called with the address of each functlet individually?

swaroop-sridhar:
A few side comments -- which should be considered/addressed separate from this checkin:

1) The `ObjectLoadListner` code has become large enough that I think we want it to live in its own file. 
2) The target specific definitions for dwarf->reg should also be separated out. I think ObjectLoadListner and GcInfo structs can share this helper.

adiaaida:
This change is intended to be a permanent fix, as we should be checking for this anyway. Locals are only valid in the subprogram that we have their information for, so we should be doing this check anyway. This change is very similar to what the getLineInfoForAddressRange function does to get line information. The reason we do not see these conflicts for line info is because we are only producing debug information for the actual function that was read in (so not the safepoints function, and I imagine also not any funclets that will be generated, but I could be wrong there). So this should work for both gc and funclets, though I will need to wait until we actually have funclets to confirm.

Do you read in dwarf info in GcInfo? The dwarf->reg function is pretty specific to dwarf debug info.

swaroop-sridhar:
@adiaaida: We don't need debug info for the gc.safepointpoll() functions.
However, we'll need it for funclets, since they are a part of the original source function.
@JosephTremoulet  can clarify this.

swaroop-sridhar:
GCInfo encoder uses the Dwarf register encode/decoding to read the stack base-register from `__llvm.stackmap`. There's a minimalistic infrastructure for it in [include/GcInfo/Target.h](https://github.com/dotnet/llilc/blob/master/include/GcInfo/Target.h).

So, it would be ideal to move the register definition and conversion functions to `include/Target/register.h` and use it from both `lib/GcInfo/GcInfo.cpp` and `lib/DebugInfo/ObjectLoadListner.cpp`.

JosephTremoulet:
Right, funclets will have instructions that have line numbers that should be reported, and if we're reporting locals' liveranges and/or ranges where locals are enregistered then there will be births and deaths in funclets as well.  The home locations of locals will always be in `alloca`s in the root function, never funclets.

adiaaida:
Ok. When we get there, I will have to figure out how to make it work. I'm not sure how we'll communicate this to the EE if they are all part of the same Module. If they have their own subprogram sections in the dwarf section, other parts of this will also have to change. Until I have funclets, though, I don't know how they will be represented in the dwarf info, or how to handle them.

PR #793, opened 2015-08-26
Fix for llilc build out of llvm tree

kyulee1:
When using RelWithDebInfo instead of Debug/Release, we failed to find llvm-config file.
The fix is to simply add such case as well.


kyulee1:
@richardlford PTAL

russellhadley:
LGTM

PR #798, opened 2015-08-26
Fix for using relative path for llilc_runtest.py

kyulee1:
When specifying relative paths instead of absolute paths for arguments in llilc_runtest.py, the test failed.
The fix is to expand a full path for the given path.
Also added a check to verify it for the case of the required arguments.


kyulee1:
@swaroop-sridhar  PTAL

russellhadley:
LGTM

PR #800, opened 2015-08-26
Fix #789 and #797

### LINKED ISSUE #789: Make Stack State Merge conform to ECMA standard
The ECMA Common Language Infrastructure (CLI) standard, V5, Partition III, CIL Instruction Set
in section 1.8.1.3 "Merging stack states", tells how stack states should be merged. 
For the case where the types are assignable either way (e.g. as for int32 and native int),
the merging must keep the first type found.

Our current algorithm does not do this. This issue is to fix that.

### LINKED ISSUE #797: Ensure LLVM Switch instruction selector and values have same type.
LLVM requires that the selector and case values for a switch instruction must have the same type.
At the time that LLILC first makes a switch instruction it does not know the type of the 
selector so it makes case values that are i32. If then the selector is another type, 
e.g. i64, verification of the switch instruction will fail unless the case values are
fixed up to have the type of the selector.

richardlford:
Closes #789. Closes #797.

For #789, when merging stack states, if the types being merged
are int32 and native int, keeps the first type in order to
conform to the ECMA standard. This fixed 8 tests.

For #797, if a switch has a selector whose type is not i32,
fix up the types of the case values to match. This fixed
two tests.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -5993,8 +5993,40 @@ void GenIR::switchOpcode(IRNode *Opr) {
   BasicBlock *CurrBlock = LLVMBuilder->GetInsertBlock();
   TerminatorInst *TermInst = CurrBlock->getTerminator();
   SwitchInst *SwitchInstruction = cast<SwitchInst>(TermInst);
-
   SwitchInstruction->setCondition(Opr);
+
+  Type *OprType = Opr->getType();
+  IntegerType *OprIntType = cast<IntegerType>(OprType);
+  unsigned OprBitWidth = OprIntType->getBitWidth();
+
+  // LLVM requires that for a switch instruction the type of the selector and
+  // the type of the case values all be of the same integer type.
+  // The purpose of the following code is to adjust the type of the
+  // case values to match the type of the selector if they are different.
+
+  // If a small selector type was used, e.g. i8, the type might not
+  // be big enough to hold all the case values. But the value for the
+  // selector is popped off the evaluation stack, so it should be at
+  // least 32 bits. So check that.
+  assert((OprBitWidth >= 32) && "Selector bit width is less then 32");
+
+  // If we have to adjust the types of the operands we use a
+  // uint64_t to transfer the value, so we cannot handle larger selector.
+  assert((OprBitWidth <= 64) && "Selector bit width is greater than 64");
+
+  LLVMContext &LLVMContext = *JitContext->LLVMContext;
+  for (SwitchInst::CaseIt Case : SwitchInstruction->cases()) {
+    ConstantInt *OldValue = Case.getCaseValue();
+    unsigned OldBitWidth = OldValue->getBitWidth();
+    if (OldBitWidth != OprBitWidth) {
+      // Need to adjust type.
+      assert((OldBitWidth <= 64) && "Old value bit width is greater than 64");
+      uint64_t CaseValue = OldValue->getZExtValue();
+      ConstantInt *Value =
+          ConstantInt::get(LLVMContext, APInt(OprBitWidth, CaseValue, false));
+      Case.setValue(Value);
+    }
+  }
 }
 
 void GenIR::throwOpcode(IRNode *Arg1) {
@@ -7476,8 +7508,7 @@ void GenIR::addPHIOperand(PHINode *Phi, Value *NewOperand,
         Phi->setIncomingValue(I, NewOperand);
         FoundBlockOperand = true;
         break;
-      }
-      else {
+      } else {
         assert(CurrentOperand == NewOperand);
       }
     }
@@ -7514,12 +7545,12 @@ Type *GenIR::getStackMergeType(Type *Ty1, Type *Ty2, bool IsStruct1,
 
   LLVMContext &LLVMContext = *this->JitContext->LLVMContext;
 
-  // If we have nativeint and int32 the result is nativeint.
+  // If we have nativeint and int32 the result is the first type.
   Type *NativeIntTy = Type::getIntNTy(LLVMContext, TargetPointerSizeInBits);
   Type *Int32Ty = Type::getInt32Ty(LLVMContext);
   if (((Ty1 == NativeIntTy) && (Ty2 == Int32Ty)) ||
       ((Ty2 == NativeIntTy) && (Ty1 == Int32Ty))) {
-    return NativeIntTy;
+    return Ty1;
   }
 
   // If we have float and double the result is double.
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -1252,8 +1252,9 @@ private:
 
   /// Get the type of the result of the merge of two values from operand stacks
   /// of a block's predecessors. The allowed combinations are nativeint and
-  // int32 (resulting in nativeint), float and double (resulting in double),
-  /// and GC pointers (resulting in the closest common supertype).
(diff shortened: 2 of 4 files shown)


richardlford:
@erozenfeld, PTAL.

lib/Reader/readerir.cpp
```diff
@@ -5994,8 +5994,36 @@ void GenIR::switchOpcode(IRNode *Opr) {
   BasicBlock *CurrBlock = LLVMBuilder->GetInsertBlock();
   TerminatorInst *TermInst = CurrBlock->getTerminator();
   SwitchInst *SwitchInstruction = cast<SwitchInst>(TermInst);
-
   SwitchInstruction->setCondition(Opr);
+
+  llvm::Type* OprType = Opr->getType();
```
erozenfeld: We don't use llvm:: qualifier in .cpp files.

lib/Reader/readerir.cpp
```diff
@@ -5994,8 +5994,36 @@ void GenIR::switchOpcode(IRNode *Opr) {
...
+  LLVMContext &LLVMContext = *this->JitContext->LLVMContext;
+  Type *Int32Ty = Type::getInt32Ty(LLVMContext);
+  if (OprType != Int32Ty) {
+    // When we made the case values we assumed int32 as the type for the 
+    // values. But now we have an operand of a different type. LLVM 
+    // requires the values and the switch value to have the same type.
+    // So change the type of the values to be the type of the selector.
+    assert(OprType->isIntegerTy() && "Switch selector is not an integer type");
+    llvm::IntegerType* OprIntType = (llvm::IntegerType*)OprType;
+
```
erozenfeld: Please use 
IntegerType* OprIntType = cast<IntegerType>(OprType);
instead of the assert and () cast.

lib/Reader/readerir.cpp
```diff
@@ -5994,8 +5994,36 @@ void GenIR::switchOpcode(IRNode *Opr) {
...
+    // If a small selector type was used, e.g. i8, the type might not
+    // be big enough to hold all the values. But the values for the
```
erozenfeld: typo: values --> value

lib/Reader/readerir.cpp
```diff
@@ -5994,8 +5994,36 @@ void GenIR::switchOpcode(IRNode *Opr) {
...
+    // selector is popped off the evaluation stack, so it should be at
+    // least 32 bits. So check that.
+    unsigned OprBitWidth = OprIntType->getBitWidth();
+    assert((OprBitWidth > 32) && "Selector bit width is less then 32");
```
erozenfeld: Should it be 
`(OprBitWidth >= 32)`
?

lib/Reader/readerir.cpp
```diff
@@ -5994,8 +5994,36 @@ void GenIR::switchOpcode(IRNode *Opr) {
...
+
+    // Create new values of the correct type.
+    int NumCases = SwitchInstruction->getNumCases();
+    llvm::SwitchInst::CaseIt Case = SwitchInstruction->case_begin();
```
erozenfeld: Remove 
`llvm::`

lib/Reader/readerir.cpp
```diff
@@ -7515,12 +7543,12 @@ Type *GenIR::getStackMergeType(Type *Ty1, Type *Ty2, bool IsStruct1,
 
   LLVMContext &LLVMContext = *this->JitContext->LLVMContext;
 
-  // If we have nativeint and int32 the result is nativeint.
+  // If we have nativeint and int32 the result is the first type.
```
erozenfeld: Please update the header comment for this method in readerir.h file.

erozenfeld:
Looks good other than an incorrect assert and a few nits.

richardlford:
A local run of ccformat showed some formatting errors. I've pushed updated sources (and aborted the previous tests).

lib/Reader/readerir.cpp
richardlford: We already tested that OprType is not Int32Ty, so I don't think the = 32 case would apply.

richardlford:
Responded to the review comments, squashed and rebased, pushed.

lib/Reader/readerir.cpp
```diff
@@ -5994,8 +5994,35 @@ void GenIR::switchOpcode(IRNode *Opr) {
...
+    // least 32 bits. So check that.
+    unsigned OprBitWidth = OprIntType->getBitWidth();
+    assert((OprBitWidth > 32) && "Selector bit width is less then 32");
+
+    // Create new values of the correct type.
+    int NumCases = SwitchInstruction->getNumCases();
+    SwitchInst::CaseIt Case = SwitchInstruction->case_begin();
+    for (size_t I = 0; I < NumCases; I++) {
+      ConstantInt *Value = ConstantInt::get(
+          *JitContext->LLVMContext, APInt(OprBitWidth, (uint64_t)I, false));
```
JosephTremoulet: Should this maybe assert that the value we're replacing was in fact (32-bit) I?

richardlford: Currently we unconditionally make all the case values i32. The current code could fail if the selector was i32 and original case values had some type other than i32. We could instead check the type of the first case and compare that with the type of the selector. If they differed then we could do the fix. Adding an assert, as you suggest, would at least detect the problem.

JosephTremoulet: Sorry, I didn't mean this should assert something about the original type, I meant it should assert something about the original value.  Your example in e-mail was:
```
  switch i64 %23, label %24 [
    i32 6, label %109
    i32 1, label %26
    i32 2, label %41
    i32 3, label %48
    i32 4, label %52
    i32 5, label %64
  ]
```
which, as written, this code would extend to
```
  switch i64 %23, label %24 [
    i64 1, label %109
    i64 2, label %26
    i64 3, label %41
    i64 4, label %48
    i64 5, label %52
    i64 6, label %64
  ]
```
which would be incorrect.  I've looked at the code and I think it's true that given the way we generate these switches the value of the _n_ th case will always be _n_, but then again I don't know where the switch in the e-mail got permuted, an assertion would make sure no permuting has happened (or you could compute the new value based on the old value rather than based on the `for` loop iterator).

richardlford:

lib/Reader/readerir.cpp
richardlford: Now that you mention it, the switch printed out when the verification failed looks odd since it does not have a value for 0. I've gone back into the debugger and verified that the switch at the 
point where I modify the types is:

```
  switch i64 %23, label %24 [
    i32 0, label %24
    i32 1, label %25
    i32 2, label %27
    i32 3, label %29
    i32 4, label %30
    i32 5, label %31
    i32 6, label %36
  ]
```

so the switch gets modified later. However seeing this I do think the code should be written
to be more robust. It should have as a post-condition that the case values are preserved 
but that all their types are modified to match. I will make those changes.

JosephTremoulet: Not that it matters since we're agreed on the direction now, but I see that the switch at verification time is what you'd get if you took the switch at the point where you modify the types and called `removeCase` on the 0 case, which perhaps happened as part of the unreachable code removal pass if that target goes to an `unreachable`.

richardlford:
I've pushed a (squashed) update that makes the switch code more robust as discussed above. @erozenfeld  and @JosephTremoulet, PTAL.

JosephTremoulet:
LGTM, thanks

PR #802, opened 2015-08-27
Fix assert in box codegen.

### LINKED ISSUE #643: Assertion: unexpected CorInfoType passed to GenIR::size
This pops up in one of the new IL tests.

`Assertion failed: (0) && "UNREACHED", file ..\..\..\..\..\tools\llilc\lib\Reader\readerir.cpp, line 1957`

AndyAyersMS:
Closes #643.

When boxing, pass the size instead of the CorType into the argument helper, so if the box type is a value type we have the size available.

Enable the test that was failing with the assert.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -794,7 +794,7 @@ public:
   void callMonitorHelper(bool IsEnter);
 
   IRNode *convertToBoxHelperArgumentType(IRNode *Opr,
-                                         CorInfoType CorType) override;
+                                         uint32_t DestSize) override;
 
   IRNode *makeBoxDstOperand(CORINFO_CLASS_HANDLE Class) override;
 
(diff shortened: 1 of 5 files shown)


AndyAyersMS:
@erozenfeld PTAL.

include/Reader/readerir.h
```diff
@@ -794,7 +794,7 @@ class GenIR : public ReaderBase {
   void callMonitorHelper(bool IsEnter);
 
   IRNode *convertToBoxHelperArgumentType(IRNode *Opr,
-                                         CorInfoType CorType) override;
+                                         uint32_t DestSize) override;
```
erozenfeld: Nit: can you make the name of the last param the same in reader.h and readerir.h? They are Size and DestSize.

erozenfeld:
LGTM.

include/Reader/readerir.h
AndyAyersMS: Made both be `DestSize`.

Issue #633, opened 2015-06-11
Assert while importing indirect calls through function-pointer-typed locals/args


pgavlin:
`Assertion failed: Target->getType()->isIntegerTy(Reader.TargetPointerSizeInBits), file ..\..\..\..\..\tools\llilc\lib\Reader\abisignature.cpp, line 314`

The EE does not distinguish between normal pointers and function pointers, so the reader ends up blowing up when it tries to call through what looks like a pointer to a struct.

richardlford:
Closed by PR 805,

PR #805, opened 2015-08-28
Fix issue #633

richardlford:
This fix handles function pointers in calls by using
bitcast to convert to pointer to function type,
rather than inttoptr as is done if the target is a
pointer-sized integer.

Two of the tests were originally classified as #637
but after that was fixed became #633.

With this fix 11 tests pass. 5 tests were found
to no longer exist. 4 tests still fail with
thrown exceptions and are given issue #13.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -5182,6 +5182,18 @@ IRNode *GenIR::genCall(ReaderCallTargetData *CallTargetInfo, bool MayThrow,
                        std::vector<IRNode *> Args, IRNode **CallNode) {
   IRNode *Call = nullptr;
   IRNode *TargetNode = CallTargetInfo->getCallTargetNode();
+  if (TargetNode->getType()->isPointerTy()) {
+    // According to the ECMA CLI standard, II.14.5, the preferred
+    // representation of method pointers is a native int
+    // which is an int the size of a pointer.
+    // However the type returned by the CoreClr for a method pointer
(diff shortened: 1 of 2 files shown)


richardlford:
@pgavlin, @erozenfeld, PTAL.

lib/Reader/abisignature.cpp
```diff
@@ -448,8 +450,11 @@ Value *ABICallSignature::emitCall(GenIR &Reader, Value *Target, bool MayThrow,
   const bool IsVarArg = false;
   Type *FunctionTy = FunctionType::get(FuncResultType, ArgumentTypes, IsVarArg);
   Type *FunctionPtrTy = Reader.getUnmanagedPointerType(FunctionTy);
-
-  Target = Builder.CreateIntToPtr(Target, FunctionPtrTy);
+  if (Target->getType()->isPointerTy()) {
+    Target = Builder.CreateBitCast(Target, FunctionPtrTy);
```
pgavlin: I'd prefer that this transformation be done up the stack from `::emitCall`, as it deals with IL semantics (and this code doesn't, for the most part). In that case, the assert above would remain `assert(...isIntegerTy(...))`. It might also be good to note the parts of the spec that imply that this is necessary (which I think is just II.14.5 and III.3.20).

richardlford:
@pgavlin , here is a revised version that makes the change further up the stack
as you suggested. PTAL.

pgavlin:
LGTM. Thanks for doing this.

Issue #808, opened 2015-08-31
Fix Float to Integer Conversion constant folding


richardlford:
In CoreClr test JIT\Regression\CLR-x86-JIT\V1-M10\b05617\b05617
there is the following MSIL:

```
ldc.r4 2.564783e+007
conv.u1
```

The constant loaded has hex value 0x1875AD6 which should truncate to 0xD6 
when converting to u1. However LLILC, using LLVM utilities, is currently
returning 0. Either the LLVM utilities are incorrect, or we are misunderstanding their
defined semantics.

richardlford:
I've been reading the MSIL spec more carefully. In III.3.27, conv.<to type>,
it says:

```
If overflow occurs converting a floating-point type to an integer, or if the floating-point value being converted to an integer is a NaN, the value returned is unspecified.
```

So it may be that the test is not valid in that it is expecting a specified result
where the Spec says the result is unspecified. 

However, in order to give the same behavior as RyuJit it seems worthwhile
to first convert to i64 and then to the destination type.

Issue #52, opened 2015-02-18
Reader: enable vararg support


erozenfeld:

russellhadley:
@erozenfeld is this still open?  I'm guessing that we can close this because the underlying pinvoke issue is closed.

erozenfeld:
@jkotas Can you please confirm that CoreCLR doesn't support CORINFO_CALLCONV_VARARG calling convention?

jkotas:
Managed varargs are not supported in .NET Core profile. System.ArgIterator and related types that are required to use managed varargs are not part of .NET Core profile.

CoreCLR includes code to support varargs, but it is there just for the legacy scenarios. We are not doing work to make it work cross-platform, etc.

PR #809, opened 2015-08-31
Resolve Issue #645

### LINKED ISSUE #645: Investigate various non-obvious failures in the IL tests.
A number of the new IL tests fail without an assertion or a stack trace that implicates #13. These failures need to be investigated.

richardlford:
Closes #645.

Issue #645 was "Investigate various non-obvious failures in the IL tests".
Twenty-two tests were excluded because of issue #645.
The investigation did not reveal any new passing tests. The tests were
disposed as follows:

* Test found to no longer exist: 6
* Tests throwing exceptions (issue #13): 11
* Tests using unsupported variable-length argumenet lists (issue #52): 4
* Test failing because of float of integer conversin problem
  (new issue #808): 1

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -889,11 +889,8 @@
         <ExcludeList Include="$(XunitTestBinBase)\JIT\Regression\CLR-x86-JIT\V2.0-RTM\b598031\test.cmd" >
              <Issue>13</Issue>
         </ExcludeList>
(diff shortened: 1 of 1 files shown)


richardlford:
@AndyAyersMS, PTAL. Note this PR does not change the compiler, only the exclusions list.

AndyAyersMS:
LGTM.

PR #811, opened 2015-09-01
Resolve issue #634

### LINKED ISSUE #634: Assertion failure due to a type mismatch between a storage location and the value being stored
`Assertion failed: getOperand(0)->getType() == cast<PointerType>(getOperand(1)->getType())->getElementType() && "Ptr must be a pointer to Val type!", file ..\..\..\lib\IR\Instructions.cpp, line 1038`

richardlford:
Closes #634.

I reassessed the tests formerly excluded for issue #634.
Of these, two no longer exist and the remaining 8 pass.
So remove them from the exclusion list.

I also corrected one typo in a comment in the make_sln.py
script.


richardlford:
@AndyAyersMS, PTAL.

AndyAyersMS:
LGTM

PR #812, opened 2015-09-01
Handle cases where the finally lexically precedes the try.

### LINKED ISSUE #644: Assertion in LLVM when deleting a Value that is still in use
This pops up in one of the new IL tests.

`Assertion failed: use_empty() && "Uses remain when a value is destroyed!", file ..\..\..\lib\IR\Value.cpp, line 83`

AndyAyersMS:
Closes #644.

Create the finally dispatch IR a bit more flexibly -- either when the reader first sees the endfinally, or first sees a leave.  Otherwise we risk leaving the finally IR dangling outside of any basic block.

The test case now fails because of EH, so reclassify as blocked by issue #13.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -1655,6 +1655,16 @@ private:
 
   int getElementCountOfSIMDType(CORINFO_CLASS_HANDLE Class) override;
 
+  /// Create the IR for a finally dispatch.
+  ///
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@JosephTremoulet PTAL.

JosephTremoulet:
LGTM, thanks for fixing this.

PR #814, opened 2015-09-01
Resolve issue #635

### LINKED ISSUE #635: Assertion from LLVM: isa<> used on a null pointer
This pops up during several of the new IL tests.

`Assertion failed: Val && "isa<> used on a null pointer", file E:\dev\LLVM\include\llvm/Support/Casting.h, line 95`

richardlford:
Closes #635.

All four of the tests excluded for issue #635 throw
exceptions as part of their normal execution.
So change the issue number for these to #13
until LLILC implements catching exceptions.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -602,7 +602,7 @@
              <Issue>13</Issue>
         </ExcludeList>
         <ExcludeList Include="$(XunitTestBinBase)\JIT\Regression\Dev11\Dev11_468598\Test_HndIndex_10_Reordered.cmd" >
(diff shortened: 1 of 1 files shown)


richardlford:
@AndyAyersMS, PTAL.

AndyAyersMS:
LGTM.

richardlford:
The failed leg was one that does not run tests so could not have been affected by this change. So I'm going to go ahead and merge.

pgavlin:
I clarified with Rich offline that the original problem referenced in #635 no longer repros.

PR #815, opened 2015-09-02
Tolerate integers in mkrefany.

### LINKED ISSUE #641: Assertion: bad value passed to mkrefany
This pops up during one of the new IL tests.

`Assertion failed: isManagedPointerType(Object->getType()) && "wrong type for refany value", file ..\..\..\..\..\tools\llilc\lib\Reader\readerir.cpp, line 6712`

AndyAyersMS:
Closes #641.

Ecma-335 III.4.19 says the argument to `mkrefany` must be type `&` or `native int`, but it appears traditional that arbitrary ints are allowed. Tolerate this.

Enabled the one blocked test that now passes.


AndyAyersMS:
@richardlford PTAL.

richardlford:
LGTM.

Issue #543, opened 2015-05-05
Track last-known-good commits of repos LLILC depends on


JosephTremoulet:
It would be good to know what sync points of LLVM and CoreCLR are known/expected to work with a given LLILC commit, to avoid getting randomized by failures from accidental mismatched sync (and to avoid halting all progress when a change requiring LLILC update gets checked in to CoreCLR).

Strawman proposal:
1. Add a file to LLILC repo whose contents look something like
   
   > llvm: 16afeda234
   > coreclr: 3251342a
2. Lab scripts checkout the specified commits of each repo before building and running tests
3. When LLVM or CoreCLR has a change that breaks LLILC, when the fix goes in to LLILC, that change also includes updating the LKG for the relevant repo
4. We're already pulling changes from LLVM:master to LLVM:MS on a regular basis.  Whenever LLVM:MS gets updated in this way, LLILC would need a companion change updating the LKG stamp for LLVM
5. Branch owner (or a bot?) would daily test updating CoreCLR LKG to latest and check in the change to update if successful.  (And if unsuccessful, we could file an issue and fix it in parallel to ongoing work without bringing down LLILC testing in the meantime)
6. When we can start using packages/drops instead of building LLVM/CoreCLR, step 2 would change accordingly, but the rest of this should be pretty much the same.

I'm curious if others agree that something like this would be a good idea.

erozenfeld:
Yes, this is definitely a good idea. We should always know lkg timestamps for llvm and coreclr.

apskim:
Have you considered using `git subtree`, relying on a fixed commit of llvm/coreclr, and only update the subtrees occasionally to other fixed commits that have been tested to work?

JosephTremoulet:
@apskim, would that require housing the llvm and coreclr enlistments as subfolders inside the llilc enlistment?  I'm hesitant to start requiring that (currently we're flexible wrt how a dev wants to arrange these locally in relation to each other, and in fact putting llvm under llilc would be inverted from the setup that many of us use where llilc is homed under llvm/tools).  Also, where we'd really like to end up is being able to pull down prebuilt packages of llvm and coreclr instead of building them locally, and it seems like even if we used subtree now we'd have to switch to something like I outlined above once we get to packages, no?

JosephTremoulet:
@adiaaida, would this be easy to set up in the lab?

adiaaida:
[...]

PR #816, opened 2015-09-03
Fix a warning and enable warnings as errors.

AndyAyersMS:
Fix a C4189 that I introduced recently. Turn on -WX for Windows builds for the code we author.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,6 +261,15 @@ if (DEBUG)
   add_definitions( -DDEBUG -D_DEBUG )
 endif()
 
+# Add appropriate flags for MSVC
+if (MSVC)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -EHsc")
+  # Force -WX on even if it's not enabled for LLVM.
+  if (NOT LLVM_ENABLE_WERROR)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -WX")
+  endif()
+endif()
+
 # Add appropriate flags for GCC
 if (LLVM_COMPILER_IS_GCC_COMPATIBLE)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-common -Woverloaded-virtual -Wcast-qual -fno-strict-aliasing -fno-rtti")
--- a/lib/GcInfo/CMakeLists.txt
+++ b/lib/GcInfo/CMakeLists.txt
@@ -4,10 +4,6 @@ include_directories(${LLILC_INCLUDES}/clr
                     ${LLILC_INCLUDES}/GcInfo
                     ${LLILC_INCLUDES}/Jit
                     ${LLILC_INCLUDES}/Pal)
-
-if( WIN32 )
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -EHsc")
-endif()
 					
 add_definitions(-DSTANDALONE_BUILD)
 					
--- a/lib/Jit/CMakeLists.txt
+++ b/lib/Jit/CMakeLists.txt
@@ -23,8 +23,6 @@ set(LLVM_LINK_COMPONENTS
 set(LLILCJIT_LINK_LIBRARIES LLILCReader GcInfo)
 
 if (WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -EHsc")
-
   # Create .def file containing a list of exports preceeded by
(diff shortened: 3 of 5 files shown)


AndyAyersMS:
@JosephTremoulet I think you asked for this, so PTAL.

JosephTremoulet:
1. Looks from the lab job like we have some additional warnings in release builds. @erozenfeld, were you looking into those?
2. One of the release warnings is coming from the gcencoder files we import from CoreCLR.  I think @swaroop-sridhar has PRed changes there to get rid of warnings in LLILC before, and we can do the same again here, but if we're going to keep getting breaks like this I wonder if maybe it wouldn't be prudent to build with warnings-as-errors after all (until/unless we do something along the lines of #543)
3. I don't know what's the difference between setting LLVM_ENABLE_WERROR=ON versus directly appending -WX, but I'd imagine we may want the former.

AndyAyersMS:
I'm reluctant to turn on warning-as-error more broadly unless we're sure everyone who works on LLILC has the ability to repro the non-windows builds easily.

JosephTremoulet:
My comment about LLVM_ENABLE_WERROR was intended as a question about the mechanism, not the policy. I'd imagine there's some way in cmake to set LLVM_ENABLE_WERROR conditionally based on the host platform, no? Are we sure that setting -WX where you set it has the same effect, or is it possible that using LLVM_ENABLE_WERROR would also pull in some -Wds, or affect only a subset of the compiler invocations, or set some preprocessor symbol, or etc.?

AndyAyersMS:
Well, this question is answerable -- as of right now, (per LLVM sources)[https://github.com/Microsoft/llvm/blob/master/cmake/modules/HandleLLVMOptions.cmake], if warnings are enabled, `LLVM_ENABLE_WERROR` turns on `-Wx` for windows, and `-Werror` for "gcc compatible" compilers. So we could instead use `LLVM_ENABLE_WERROR` conditionally for windows to limit the scope of the error checks. It wouldn't change the behavior we see with just turning on `-Wx`.

AndyAyersMS:
@erozenfeld @JosephTremoulet I'll fix the errors in LLILC... did we ever settle on which `#defines` we want to use for debug vs release?

JosephTremoulet:
To me it seems like a kind of "layering violation" to use `-WX` directly, but feel free to ignore me if you think that's silly.  The only other thought I have about it is that I don't know if implementing it via `LLVM_ENABLE_WERROR` could make it easier for someone to override it on their cmake invocation, which could be desirable.

So LGTM modulo fixing the release warnings (or limiting this to just warn-as-error in debug builds), and fixing or excluding the imported CoreCLR sources.

JosephTremoulet:
> did we ever settle on which #defines we want to use for debug vs release?

I didn't know there was an issue there, what are the alternatives?  I thought `NDEBUG` #defined for release and undefined for debug was the only option.

AndyAyersMS:
I've switched over to using `LLVM_ENABLE_WERROR`. 

As far as conditional code: for "debug" only code we'd use `#ifndef NDEBUG`, right? Note we have a similar `NODEBUG` lurking about (as well as `_DEBUG`, etc).

JosephTremoulet:
Right, `#ifndef NDEBUG` for debug-only code.  At least that's what I've been using, based on what LLVM does.

AndyAyersMS:
I have what I hope is a full set of fixes locally, though annoyingly I can't repro some of the lab warnings in my local builds. It's obvious what needs to be fixed though, so this isn't blocking.

Waiting for related LLVM/CoreClr fixes to go in before pushing the LLILC part up.

AndyAyersMS:
Ok, @JosephTremoulet, take another look. I updated the way we enable warnings as errors, and fixed all the warnings (I hope).

JosephTremoulet:
LGTM, thanks for fixing this!

AndyAyersMS:
Argh -- auditing the build lab logs, `set(LLVM_ENABLE_WERROR true)` is not effective. Need to redo this with something like `set(LLVM_ENABLE_WERROR ON)`.

AndyAyersMS:
@JosephTremoulet had to backtrack to really get -WX through cmake. What I have now is similar to how we forcibly enable `-pedantic` for GCC, even if LLVM is not set up that way.

JosephTremoulet:
Thanks, sorry for the headache.  I do like how the MSVC flags are now conditional on MSVC and not on WIN32, at least.

Looks good to me.

AndyAyersMS:
Fixed a second-order effect....

JosephTremoulet:
Still green :)

erozenfeld:
LGTM too. Glad to see these treated as errors.

PR #819, opened 2015-09-03
Adding NYI if verifying to avoid degenerate Blocks

jashook:
If asked to verify we will currently check and break of out phase 2 of the
reader early. This will cause a BB to be processed later without a
terminator causing a crash.


russellhadley:
LGTM (assuming tests pass)

russellhadley:
@jashook go a head and merge. :)

jashook:
Merged

PR #823, opened 2015-09-04
Make float to int conversion more determined.

### LINKED ISSUE #808: Fix Float to Integer Conversion constant folding

richardlford:
The ECMA-335 spec says that for unchecked floating point to
integer conversion, that if there is overflow then the result
is unspecified. However for compatibility with other jits
we make the result more deterministic. If the target
bit size is less than 64 we first convert to i64 where overflow
is less likely. Then we truncate to the target bit size.

Closes #808 and allows test
\JIT\Regression\CLR-x86-JIT\V1-M10\b05617\b05617.cmd
to pass.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -5979,8 +5979,23 @@ IRNode *GenIR::conv(ReaderBaseNS::ConvOpcode Opcode, IRNode *Source) {
     Conversion = SourceIsSigned ? LLVMBuilder->CreateSIToFP(Source, TargetTy)
                                 : LLVMBuilder->CreateUIToFP(Source, TargetTy);
   } else if (SourceTy->isFloatingPointTy() && TargetTy->isIntegerTy()) {
-    Conversion = DestIsSigned ? LLVMBuilder->CreateFPToSI(Source, TargetTy)
-                              : LLVMBuilder->CreateFPToUI(Source, TargetTy);
+    // The ECMA-335 spec says that for unchecked floating point to
+    // integer conversion, if there is overflow then the result
+    // is unspecified. However for compatibility with other jits
+    // we want to first convert to i64 and then truncate.
+    IntegerType *TargetIntegerType = cast<IntegerType>(TargetTy);
+    int TargetBitWidth = TargetIntegerType->getBitWidth();
+    if (TargetBitWidth < 64) {
+      IntegerType *Int64Ty = Type::getInt64Ty(*JitContext->LLVMContext);
+      Conversion = DestIsSigned ? LLVMBuilder->CreateFPToSI(Source, Int64Ty)
+                                : LLVMBuilder->CreateFPToUI(Source, Int64Ty);
+      // Now do the final conversion to the target type.
+      Conversion =
+          LLVMBuilder->CreateIntCast(Conversion, TargetTy, DestIsSigned);
+    } else {
+      Conversion = DestIsSigned ? LLVMBuilder->CreateFPToSI(Source, TargetTy)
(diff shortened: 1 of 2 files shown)


richardlford:
@erozenfeld, PTAL.

lib/Reader/readerir.cpp
```diff
@@ -5679,8 +5679,26 @@ IRNode *GenIR::conv(ReaderBaseNS::ConvOpcode Opcode, IRNode *Source) {
     Conversion = SourceIsSigned ? LLVMBuilder->CreateSIToFP(Source, TargetTy)
                                 : LLVMBuilder->CreateUIToFP(Source, TargetTy);
   } else if (SourceTy->isFloatingPointTy() && TargetTy->isIntegerTy()) {
-    Conversion = DestIsSigned ? LLVMBuilder->CreateFPToSI(Source, TargetTy)
-                              : LLVMBuilder->CreateFPToUI(Source, TargetTy);
+    // The ECMA-335 spec says that for unchecked floating point to
+    // integer conversion, that if there is overflow then the result
```
erozenfeld: 'that' is repeated twice

lib/Reader/readerir.cpp
```diff
@@ -5679,8 +5679,26 @@ IRNode *GenIR::conv(ReaderBaseNS::ConvOpcode Opcode, IRNode *Source) {
...
+    IntegerType *TargetIntegerType = cast<IntegerType>(TargetTy);
+    int TargetBitWidth = TargetIntegerType->getBitWidth();
+    if (TargetBitWidth < 64) {
+      IntegerType *Int64Ty = Type::getInt64Ty(*JitContext->LLVMContext);
+      if (DestIsSigned) {
+        Conversion = LLVMBuilder->CreateFPToSI(Source, Int64Ty);
+      } else {
+        Conversion = LLVMBuilder->CreateFPToUI(Source, Int64Ty);
+      }
+      // Now do the final conversion to the target type.
```
erozenfeld: Can you use the ternary operator here for consistency with the 'else' clause below?

erozenfeld:
LGTM with a couple of nits.

lib/Reader/readerir.cpp
richardlford: Fixed.

lib/Reader/readerir.cpp
richardlford: Done.

richardlford:
@erozenfeld, I've pushed the (squashed) version with the changes you suggested.

richardlford:
P.S. When I ran the formatter locally it changed the indentation of a couple of
comment lines unrelated to my change.
This is using the latest formatter binaries so hopefully will check out OK in the lab.

erozenfeld:
LGTM

AndyAyersMS:
The formatting change around line 7520 makes no sense to me. Presumably what was there before checked out ok when I submitted it.

richardlford:

richardlford:
I rebased and pushed. The rebased version does not have the extraneous formatting change near 7520.

richardlford:

richardlford:
The failures were all infrastructure related except the release win64 which is a failing a new ready-to-run test. Since the checked win64 leg passed I'm going to go ahead and merge.

Issue #750, opened 2015-08-10
ReadyToRun: string literal


erozenfeld:
Make changes to LDSTR necessary to support ReadyToRun scenario.

erozenfeld:
No special code is needed for LDSTR.

Issue #749, opened 2015-08-10
ReadyToRun: static field access


erozenfeld:
Make changes to LDSFLD and STSFLD necessary to support ReadyToRun scenario.

erozenfeld:
Implemented in #824.

Issue #748, opened 2015-08-10
ReadyToRun: type casting


erozenfeld:
Make changes to ISINST and CASTCLASS necessary to support ReadyToRun scenario.

erozenfeld:
Implemented in #824.

Issue #747, opened 2015-08-10
ReadyToRun: object creation


erozenfeld:
Make changes to NEWOBJ and NEWARR necessary to support ReadyToRun scenario.

erozenfeld:
Implemented in #824.

PR #824, opened 2015-09-05
Implement llilc support for ReadyToRun.

### LINKED ISSUE #751: ReadyToRun: shared generics
Make changes to shared generics handling necessary to support ReadyToRun scenario.

erozenfeld:
These changes implement jit features for ReadyToRun compilation.
The ReadyToRun overview is located here:
https://github.com/dotnet/coreclr/blob/master/Documentation/botr/readytorun-overview.md
It's a format that provides version resiliency on top of ngen.

These changes mostly follow what's implemented in RyuJit.

Special handling is done for the following operations:

* Type context creation
* Struct copy with write barriers: CORINFO_HELP_ASSIGN_STRUCT is not available for ReadyToRun
and is not used by RyuJit at all (it's considered inefficient). I implemented inline copy expansion.
* Field access
* isinst and castclass
* Class constructor insertion
* Static field address calculation
* Direct call target computation
* Indirect virtual call target computation
* Function pointer loading
* Object creation
* Array creation

Some additional changes:

* OptLevel is set to Default for ngen and ReadyToRun compilation
regardless of whether we are compiling debug or release.
* CodeModel is set to Default instead of JITDefault for ngen
and ReadyToRun compilation. We get rel32 relocations, which matches
what RyuJit generates.

COMPLUS_AltJitNgen is now set to * for our test runs. That ensures that
the existing ReadyToRun tests will use llilc to generate ReadyToRun images.

I verified that llilc can produce ReadyToRun images of tiny Roslyn and
can use them to compile itself and the result can run.

Closes #751, #750, #749, #748, #747, #745, #746

This pull request is dependent on https://github.com/dotnet/coreclr/pull/1501 merging first.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -274,15 +274,28 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
     }
     TargetOptions Options;
     CodeGenOpt::Level OptLevel;
-    if (Context.Options->OptLevel != ::OptLevel::DEBUG_CODE) {
+    bool IsNgen = Context.Flags & CORJIT_FLG_PREJIT;
+    bool IsReadyToRun = Context.Flags & CORJIT_FLG_READYTORUN;
+
+    // Optimal code for ReadyToRun should have all calls in call [rel32] form to
+    // enable crossgen to use shared delay-load thunks. We can't guarantee that
+    // LLVM will always generate this form so we currently don't take advantage
+    // of that and use non-shared delay-load thunks. The plan is to have as many
+    // calls as possible in that form and use shared delay-load thunks when
+    // possible. Setting OptLevel to Default increases the chances of calls via
+    // memory and setting CodeModel to Default enables rel32 relocations.
+    if ((Context.Options->OptLevel != ::OptLevel::DEBUG_CODE) || IsNgen ||
+        IsReadyToRun) {
       OptLevel = CodeGenOpt::Level::Default;
     } else {
       OptLevel = CodeGenOpt::Level::None;
       // Options.NoFramePointerElim = true;
     }
-    TargetMachine *TM = TheTarget->createTargetMachine(
-        LLILC_TARGET_TRIPLE, "", "", Options, Reloc::Default,
-        CodeModel::JITDefault, OptLevel);
+    llvm::CodeModel::Model CodeModel =
+        (IsNgen || IsReadyToRun) ? CodeModel::Default : CodeModel::JITDefault;
+    TargetMachine *TM =
+        TheTarget->createTargetMachine(LLILC_TARGET_TRIPLE, "", "", Options,
+                                       Reloc::Default, CodeModel, OptLevel);
     Context.TM = TM;
 
     // Set target machine datalayout on the method module.
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -872,9 +872,9 @@ void GenIR::zeroInitBlock(Value *Address, Value *Size) {
                  (IRNode *)ZeroByte, (IRNode *)Size);
 }
 
-void GenIR::copyStruct(Type *StructTy, Value *DestinationAddress,
-                       Value *SourceAddress, bool IsVolatile,
-                       ReaderAlignType Alignment) {
+void GenIR::copyStructNoBarrier(Type *StructTy, Value *DestinationAddress,
+                                Value *SourceAddress, bool IsVolatile,
+                                ReaderAlignType Alignment) {
   // TODO: For small structs we may want to generate an integer StoreInst
   // instead of calling a helper.
   const DataLayout *DataLayout = &JitContext->CurrentModule->getDataLayout();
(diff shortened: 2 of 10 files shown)


erozenfeld:
@AndyAyersMS PTAL

lib/Reader/CMakeLists.txt
```diff
@@ -21,3 +21,5 @@ add_llilcjit_library(LLILCReader
 if( NOT LLILC_BUILT_STANDALONE )
   add_dependencies(LLILCReader intrinsics_gen)
 endif()
+
+add_definitions(-DFEATURE_READYTORUN_COMPILER)
```
AndyAyersMS: We set `FEATURE_CORECLR` up in the root CMakeLists.txt. For consistency perhaps we should set `FEATURE_READYTORUN_COMPILER` up there too? Then you don't need to duplicate it in each leaf cmake file.

lib/Jit/LLILCJit.cpp
```diff
@@ -274,15 +274,20 @@ CorJitResult LLILCJit::compileMethod(ICorJitInfo *JitInfo,
     }
     TargetOptions Options;
     CodeGenOpt::Level OptLevel;
-    if (Context.Options->OptLevel != ::OptLevel::DEBUG_CODE) {
+    bool IsNgen = Context.Flags & CORJIT_FLG_PREJIT;
+    bool IsReadyToRun = Context.Flags & CORJIT_FLG_READYTORUN;
+    if ((Context.Options->OptLevel != ::OptLevel::DEBUG_CODE) || IsNgen ||
```
AndyAyersMS: This feels like it will be a target specific thing. I'd say it at least deserves a comment saying why we are setting the opt levels like this for x64.

lib/Reader/readerir.cpp
```diff
@@ -887,6 +887,89 @@ void GenIR::copyStruct(Type *StructTy, Value *DestinationAddress,
         Alignment, IsVolatile);
 }
 
+void GenIR::copyStruct(CORINFO_CLASS_HANDLE Class, IRNode *Dst, IRNode *Src,
+                       ReaderAlignType Alignment, bool IsVolatile,
+                       bool IsUnchecked) {
+  if (classHasGCPointers(Class)) {
+    GCLayout *RuntimeGCInfo = getClassGCLayout(Class);
+    const uint32_t PointerSize = getPointerByteSize();
```
AndyAyersMS: We went to a fair amount of trouble to have LLVM types correctly describe the GC references in value types. Seems like we should try and use that here instead of parsing the runtime GC info, especially because you seem to rely on the LLVM type info a bit lower down anyways.

lib/Reader/readerir.cpp
```diff
@@ -1678,6 +1762,44 @@ Type *GenIR::getClassTypeWorker(
     }
   }
 
+  uint32_t EEClassSize = 0;
+  bool HaveClassSize = false;
+  if (!IsRefClass) {
+    try {
+      EEClassSize = getClassSize(ClassHandle);
+      HaveClassSize = true;
+    } catch (...) {
```
AndyAyersMS: Should we catch something more specific here than `...`?

lib/Reader/readerir.cpp
erozenfeld: I considered that but felt that the code would be more complicated with nested struct processing. If you strongly prefer that I not use GCLayout, I'd like to change that separately later.
Also, I think now that perhaps I can simplify the code I have here and not try to figure out the exact type of the gc pointer and just use object.

lib/Reader/readerir.cpp
erozenfeld: It's HRException with hr set to E_FAIL. We currently don't have access to the coreclr eh.h file that declares that exception. Since the exception is so generic I didn't feel it was much better than ...

AndyAyersMS:
I have taken an initial look but would like to do another pass of review.

lib/Reader/CMakeLists.txt
erozenfeld: Done.

lib/Jit/LLILCJit.cpp
erozenfeld: Added a comment.

erozenfeld:
@dotnet-bot test this please

erozenfeld:
@dotnet-bot test this please

lib/Reader/readerir.cpp
AndyAyersMS: Is there any way to query the class handle up front to see if asking the runtime for size would cause an exception?

AndyAyersMS:
No real comments of substance from my second pass, other than trying to avoid the exception around 'getClassSize' if there's some way to do that.

I'm kind of curious how this all would have looked if we tried to implement ReadyToRun via virtuals instead of a number of conditional checks in various methods. Given how you had to modify logic in both reader.cpp and readerir.cpp I suppose it would have been somewhat clunky. But in some sense it feels like the right way to implement this sort of variant code gen... maybe we can consider refactoring down the road.

lib/Reader/readerir.cpp
erozenfeld: No, there isn't. I asked @jkotas about whether such a way could be added and his response was
"I think it would be better if the JIT-EE interface method had an argument that would say whether the layout of struct should be recorded as method pre-requisite."
He later clarified that the way it would work was "You can pass “do not record the size as pre-requisite” when you are mirroring the type system, and then call the API again with “do record size as pre-requisite” as you actually used the information later.

Or you can always pass in “do not record the size as pre-requisite” - assuming that we will eventually get framework build-time tooling in place that marks valuetype layout changes as breaking changes."

I thought it wasn't straightforward for us to do what he suggested and just decided to go with catching the exception for now. I may revisit this later depending on how soon that build tooling is expected to be in place. (Jan did say it was the plan of record).

erozenfeld:
I followed RyuJit's approach with conditionals. Yeah, I think it's possible to do it with virtuals but will probably require another base class and will be clunky at places... Not sure it's worth it, the number of conditionals isn't that large. If more R2R features are added we may consider refactoring.

erozenfeld:
@AndyAyersMS Are you ok with my current changes or do you want something changed before I merge?

AndyAyersMS:
I'm ok with what you have now. LGTM.

PR #826, opened 2015-09-07
Fix cross-compile cmake error

JosephTremoulet:
When cmake is "cross-compiling" LLVM (which means that the build-host
which LLVM will be built on differs from the build-target/compile-host
that LLVM will be built to run on [regardless whether the built LLVM is a
cross-compiler or not]), it uses a recursive cmake invocation targeting
the build-host to build some tools that LLVM uses in its build process.
Check the LLVM_TARGET_IS_CROSSCOMPILE_HOST flag that is set in these
recursive invocations, and skip adding LLILC targets when it is set; this
is fine to do because LLILC is not one of the tools that LLVM uses in its
build process, and is necessary because the recursive invocations do not
pass the WITH_CORECLR setting that LLILC requires.

This will allow configuring with LLVM_OPTIMIZED_TABLEGEN=ON (which
internally uses cross-compiling), which can provide substantial build-time
savings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,536 +1,544 @@
-# If we are not building as a part of LLVM, build LLILCJit as an
-# standalone project, using LLVM as an external library:
-if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
-  cmake_minimum_required(VERSION 2.8.12)
+# Do nothing if building the host tools for a cross-compile; LLILC does not need to be
+# included in the host tools, and the recursive invocation that builds them does not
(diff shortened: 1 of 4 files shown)


JosephTremoulet:
This depends on Microsoft/llvm#81

JosephTremoulet:
Looks like the diff of CMakeLists.txt had trouble figuring this out, but the only change there is to wrap the file in if()/endif() and indent accordingly.

JosephTremoulet:
@AndyAyersMS PTAL

AndyAyersMS:
Can't wait to get this in to enable super-fast rebuilds of LLVM. We should enable in the lab too since this will save a lot of time.

LGTM.

PR #829, opened 2015-09-08
Removing unnecessary temporary.

jashook:
The temporary vector that was created is reallocated instead of resized.
Small change to resize the vector to avoid any possible memory leaks due
to unsafely reallocating the vector.

### DIFF
--- a/lib/Reader/reader.cpp
+++ b/lib/Reader/reader.cpp
@@ -5211,6 +5211,13 @@ ReaderBase::rdrCall(ReaderCallTargetData *Data, ReaderBaseNS::CallOpcode Opcode,
 
   std::vector<CallArgType> &ArgumentTypes =
       Data->CallTargetSignature.ArgumentTypes;
+
+  // Initialize ArgumentTypes for the first time
+  //
+  // CallTargetSignature's constructor has not been called at this
+  // point. Therefore, a placement new is required to initialize
+  // ArgumentTypes.
+
   new (&ArgumentTypes) std::vector<CallArgType>(TotalArgs);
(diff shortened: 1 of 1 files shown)


jashook:
@dotnet-bot test this please.

@erozenfeld PTAL

swaroop-sridhar:
@dotnet-bot  test this please

jashook:

lib/Reader/reader.cpp
```diff
@@ -5211,6 +5211,8 @@ ReaderBase::rdrCall(ReaderCallTargetData *Data, ReaderBaseNS::CallOpcode Opcode,
 
   std::vector<CallArgType> &ArgumentTypes =
       Data->CallTargetSignature.ArgumentTypes;
+
+  // Vector is being initialized for the first time as it was skipped in init.
```
erozenfeld: I don't understand this comment. How is this case different from the one where you changed code to use .resize?

jashook: Data->CallTargetSignature is created using _alloca and its constructor is never called. Therefore Data->CallTargetSignature.ArgumentTypes has not been constructed when we get to this point. The comment was meant to explain that the placement new is necessary here and .resize would cause a AV.

erozenfeld: Ok, that makes sense. I would expand the comment to include that info.

jashook:
Added the comment. @erozenfeld PTAL Thank you

erozenfeld:
LGTM

jashook:
Merged

PR #830, opened 2015-09-09
Rework flow graph edge iterators to reduce leaks

AndyAyersMS:
This change revises the flow graph edge iterator implementation to reduce memory leaks. The iterator is now a stack allocated object that frees its resources when destructed. I've used move-copy semantics to avoid needing to copy the iterator payload when the iterator itself is copied.

Unfortunately it's not possible to eliminate the allocations entirely without either edges having object identity (which they don't in LLVM) or a deeper reworking of the base/client contract, or by using the `Impl` variants directly in the reader code. For now I'm content to just not leak.

I ran this through a couple of the huge allocation cases but could not measure any significant impact.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -142,7 +142,6 @@ class GenIR;  // Compiler dependent IR production
 class IRNode; // Your compiler intermediate representation
 class ReaderStack;
 class FlowGraphNode;
-class FlowGraphEdgeList;
 class BranchList;
 class ReaderBitVector;
 struct EHRegion;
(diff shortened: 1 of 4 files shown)


AndyAyersMS:
@JosephTremoulet can you take a look?

AndyAyersMS:
readytorun._mainv2_._mainv2_cmd is looking like a flaky test.

include/Reader/reader.h
```diff
@@ -1020,19 +1019,92 @@ EHRegion *getFinallyRegion(EHRegion *TryRegion);
...
+
+  /// \brief Advance the iterato
+  virtual void moveNext() = 0;
+
+  /// \brief Obtain the sink for the current edge.
+  /// \returns The sink (aka target or destination) node of the current edge.
+  virtual FlowGraphNode *getSink() = 0;
+
+  /// \brief Obtain the source for the current edge.
+  /// \return The source (aka From) node of the current ed
```
JosephTremoulet: typo: s/ed/edge/

include/Reader/reader.h
```diff
@@ -1020,19 +1019,92 @@ EHRegion *getFinallyRegion(EHRegion *TryRegion);
...
+/// An iterator for flow graph edges that does not assume there is an actual
+/// object representing the edge. At construction time specify whether the
+/// iterator can traverse the successor or predecessor edges of a block.
+class FlowGraphEdgeIterator {
+public:
+  /// \brief Construct an iterator.
+  ///
+  /// \param Block         The basic block to use as edge source or sink.
+  /// \param IsSuccessor   If true, iterate the successor edges of the block.
+  FlowGraphEdgeIterator(FlowGraphNode *Block, bool IsSuccessor);
```
JosephTremoulet: Having this declared in the base reader.h but defined in the derived readerir.cpp threw me for a loop.  I'm fine with doing it that way, and I'm fine with the implication that we can't link two different derived reader types together, but I think it's worth a comment here explaining what's expected/required.

AndyAyersMS: Happy to add a note that the client must implement the constructor.

FWIW this was already an requirement -- note there are a ton of global `fg*` functions (not `ReaderBase` methods) that a client must provide. Presumably in a "multi-client" build all these methods would need similar treatment.

lib/Reader/reader.cpp
```diff
@@ -7865,12 +7862,7 @@ void ReaderBase::readBytesForFlowGraphNode(FlowGraphNode *Fg,
         false; // Blocks with errors can't have this verified
 
     // Delete all (non-EH reachability) flow edges that come from this block.
-    FlowGraphEdgeList *Arc, *ArcNext;
-    for (Arc = fgNodeGetSuccessorListActual(Fg); Arc != nullptr;
-         Arc = ArcNext) {
-      ArcNext = fgEdgeListGetNextSuccessorActual(Arc);
-      fgDeleteEdge(Arc);
-    }
+    fgRemoveAllActualSuccessorEdges(Fg);
```
JosephTremoulet: As you're freeing the base Reader class from its assumption that edges have objects, it might be nice to no longer need `DeleteEdge` (which is an awkward thing to try to implement on LLVM IR), mightn't it?  I notice that `ReaderBase::fgRemoveUnusedBlocks` is never called (only its override `ReaderIR::fgRemoveUnusedBlocks` is called), so you could remove that and then since it's the only caller of `fgDeleteBlockAndNodes` you could remove that and then since it in turn is the only caller of `fgDeleteNodesFromBlock` and `fgDeleteBlock` you could remove those as well.  Then if instead of adding `fgRemoveAllActualSuccessorEdges` (which bottoms out in NYI in `fgDeleteEdge`, you could move the NYI up here and remove `fgDeleteEdge` as well.
Feel free to defer/ignore if you don't want to scope-creep this change and/or don't agree that would be an improvement.

AndyAyersMS: I'm happy to come back to this -- would like to make a bit more progress on leaks first.

AndyAyersMS:
Updated comments per feedback.

include/Reader/reader.h
JosephTremoulet: It was my impression that those `fg*`  functions were done that way simply because they were (initially) stubbed out and so didn't need to be methods.  I've already had to move a few to be methods in the EH branch where I needed the reader object to give them real implementations.  I presume that if we wanted to support "multi-client" we'd just make them all virtual methods on `ReaderBase`.  And I guess the equivalent here would be to give `ReaderBase` a virtual factory method for FlowGraphEdgeIterators and make `Impl` settable.

JosephTremoulet:
Thanks for doing this, that code was rather embarrassing.  LGTM but with one further (optional) suggestion.

include/Reader/reader.h
AndyAyersMS: Yes, we could do that. We'd still have to sort out the typedefs of `FlowGraphNode` and the like; in our current approach the derivation runs the wrong way.

JosephTremoulet: Good point.  I wonder if we shouldn't just switch from forward-decls to typedefs (to be supplied by the client), then.  But of course that wouldn't be part of this change :).

JosephTremoulet:
Whoops, forgot one point:  would it be better to have the pointer from the iterator the the Impl be a unique_ptr?

AndyAyersMS:
Sure, that makes sense. Updated.

include/Reader/reader.h
```diff
@@ -1020,19 +1019,84 @@ EHRegion *getFinallyRegion(EHRegion *TryRegion);
...
+  ///
+  /// This method must be implemented by the client.
+  ///
+  /// \param Block         The basic block to use as edge source or sink.
+  /// \param IsSuccessor   If true, iterate the successor edges of the block.
+  FlowGraphEdgeIterator(FlowGraphNode *Block, bool IsSuccessor);
+
+  /// \brief Move-Construct an iterator from an rvalue reference.
+  /// \param Other   Temporary iterator to copy state from.
+  FlowGraphEdgeIterator(FlowGraphEdgeIterator &&Other) : Impl(nullptr) {
```
JosephTremoulet: I think you could use `= default` here, though of course it's fine as-is.

lib/Reader/readerir.cpp
```diff
@@ -3183,44 +3187,48 @@ FlowGraphNode *GenIR::fgNodeGetIDom(FlowGraphNode *FgNode) {
...
 
-FlowGraphEdgeList *fgNodeGetSuccessorList(FlowGraphNode *FgNode) {
-  FlowGraphEdgeList *FgEdge = new FlowGraphSuccessorEdgeList(FgNode);
-  if (fgEdgeListGetSink(FgEdge) == nullptr) {
-    return nullptr;
+FlowGraphEdgeIterator::FlowGraphEdgeIterator(FlowGraphNode *Fg,
+                                             bool IsSuccessor)
+    : Impl(nullptr) {
+  if (IsSuccessor) {
+    Impl = std::make_unique<FlowGraphSuccessorEdgeIteratorImpl>(Fg);
```
JosephTremoulet: I think you need `llvm::make_unique` here (`#include "llvm/STLExtras.h"`) to build with default gcc on Linux, since it's using c++11 and `std::make_unique` is a c++14 feature.  (I know llvm proper needs to do that, I'm not *sure* about LLILC).

JosephTremoulet:
LGTM modulo `make_unique` and one nit.

AndyAyersMS:
Updated to use `llvm::make_unique`. 

Sadly VS2013 [can't handle](https://msdn.microsoft.com/en-us/library/hh567368.aspx#defaultedanddeleted) a defaulted move-constructor yet.

JosephTremoulet:
Ah well.  Looks good, thanks for the updates.

AndyAyersMS:
I'm confident this change is fine and the lab failures are unrelated (one RtR issue and some git clone problems). So, I'll be merging.

PR #831, opened 2015-09-09
Generate new WinEH representation in reader

JosephTremoulet:

Change the reader to generate catchpad/cleanuppad instead of landingpad:
 - Catch handlers translate to catchpads with the token for the caught
   exception type as an argument.
 - Finally handlers translate to cleanuppads with no arguments.
 - Fault handlers translate to cleanuppads with a single dummy argument to
   distinguish them from finally handlers.
 - Filter expressions get outlined into separate filter functions, and
   their handlers get translated to catchpads with the filter function as
   an argument.

The IR generated for finally/fault handlers currently fails verification
because the uses of cleanuppads on cleanuprets/cleanupendpads aren't
dominated by those defs; we'll ultimately want a cleanupcall operator, and
the near-term plan is to duplicate out the exceptional path for each
finally in a post-pass, but this duplication is NYI.

The IR generated for the other handler types looks reasonable on some
small testcases.

None of this IR can be lowered to machine code yet, as LLVM's lowering
does not yet handle these opcodes.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -116,6 +116,10 @@ public:
   uint32_t EndMsilOffset;
   llvm::Instruction *HandlerEHPad;
   union {
+    llvm::CatchPadInst *CatchPad;     // used for catch and filter-handler
+    llvm::CleanupPadInst *CleanupPad; // Used for finally and fault
+  };
+  union {
     llvm::SwitchInst *EndFinallySwitch; // Only used for Finally regions
     mdToken CatchClassToken;            // Only used for Catch regions
     EHRegion *HandlerRegion;            // Only used for Filter regions
@@ -549,6 +553,10 @@ void GenIR::readerPostVisit() {
       PreviousBlock = MovingBlock;
     }
   }
+
+  // TODO: Clone cleanup bodies so that exceptional path has a separate copy
+  // and the cleanuppad def will dominate all its cleanupret/endcleanuppad
+  // uses.
 }
 
 void GenIR::readerPostPass(bool IsImportOnly) {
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
I posted [a gist](https://gist.github.com/JosephTremoulet/07c9c9d2aaef8e1ee17d) with some examples; [this C#](https://gist.github.com/JosephTremoulet/07c9c9d2aaef8e1ee17d#file-catchfinally-cs) (with a finally and a couple catches) turns into [this IR](https://gist.github.com/JosephTremoulet/07c9c9d2aaef8e1ee17d#file-catchfinally-ll), and [this IL](https://gist.github.com/JosephTremoulet/07c9c9d2aaef8e1ee17d#file-filter-il) (with a filter) turns into [this IR](https://gist.github.com/JosephTremoulet/07c9c9d2aaef8e1ee17d#file-filter-ll).

JosephTremoulet:
@pgavlin PTAL

pgavlin:
Still looking. The C#/IL and IR examples you've shown seem quite reasonable; just working through the code itself.

Also, FWIW, C# 6 supports exception filters:

```
var b = false;
try {
    CallSomeMethod();
    b = true;
    CallSomeOtherMethod();
} catch (Exception e) when (b) {
  // Swallow
}
```

lib/Reader/readerir.cpp
```diff
@@ -3240,232 +3272,171 @@ LandingPadInst *GenIR::createLandingPad(EHRegion *TryRegion,
...
+    // Create the filter function.
+    uint32_t FilterOffset = FilterRegion->StartMsilOffset;
+    Twine FunctionName = Twine("Filter_") + Twine::utohexstr(FilterOffset);
+    Function *FilterFunction =
+        Function::Create(FilterTy, Function::ExternalLinkage, FunctionName,
+                         JitContext->CurrentModule);
+    FilterFunction->addFnAttr("wineh-parent", RootFunction->getName());
+    assert(PersonalityFunction != nullptr);
+    FilterFunction->setPersonalityFn(PersonalityFunction);
+    FilterFunction->setGC("coreclr");
```
pgavlin: Nit: it might be better to write this as `FilterFunction->setGC(RootFunction->getGC())` or at least assert that the GC of the funclet matches that of its parent.

JosephTremoulet: Good idea, thanks.  Updated.

lib/Reader/readerir.cpp
```diff
@@ -3164,65 +3189,72 @@ void GenIR::beginFlowGraphNode(FlowGraphNode *Fg, uint32_t CurrOffset,
...
+  assert(!rgnIsOutsideParent(Region));
+
+  Instruction *EHPad = Parent->HandlerEHPad;
+
+  // When we enter a try region, eagerly create EH pads for its handler(s),
+  // in reverse order, because each needs to name the next as its successor
+  // when it is constructed, and any invokes inside the try need to name the
+  // first handler as their successsor.
+  SmallVector<EHRegion *, 4> Handlers;
+  for (EHRegionList *ChildNode = rgnGetChildList(Region); ChildNode;
```
pgavlin: Perhaps we should add an API that allows for reverse iteration of an `EHRegionList`.

pgavlin:
LGTM aside from a couple nits.

lib/Reader/readerir.cpp
JosephTremoulet: It's a singly-linked list, that seems awkward and I think this is the one place it would be used.  Am I missing something?

pgavlin: No, your rationale is fine: I didn't look at the representation of the list and assumed that we might have something that was easily reverse-iterable.

PR #834, opened 2015-09-09
Exclude readytorun mainv2 test

russellhadley:
New test case is intermitantly failing.  This change excludes the test in deference to the issue 833.
Test will be re-enabled after investigation.


JosephTremoulet:
LGTM but you may want to fix the typo in the commit message (s/Excluse/Exclude/).

russellhadley:
One failure was a infrastructure issue.  Rerun looked fine (per @adiaaida)

Issue #794, opened 2015-08-26
Test cases that trigger massive memory use by LLILC


AndyAyersMS:
```
   JIT\jit64\opt\cse\HugeArray.cmd -- allocates > 60GB, then dies
   JIT\jit64\opt\cse\HugeArray1.cmd -- allocates > 90GB, then dies
   JIT\jit64\opt\cse\hugeexpr1.cmd -- passes (needs: 100sec, 715MB, with RelWithDbg)
   JIT\jit64\opt\cse\hugeSimpleExpr1.cmd -- passes (needs: 100sec, 715MB, with RelWithDbg)
   JIT\jit64\opt\cse\HugeField1.cmd -- allocates > 90GB, then dies
   JIT\jit64\opt\cse\HugeField2.cmd -- allocates > 90GB, then dies
```

Note RyuJit's memory usage is relatively modest in all these cases.

It might make sense to start with `hugeexpr` because the others allocate GBs so quickly that they drive the machine into paging before one can effectively debug.

Would also be great to have some form of memory profiling or leak detection that we could rely on here.

PR #836, opened 2015-09-10
Fix various memory leaks in the jit

AndyAyersMS:
This is the initial step in fixing jit-related memory leaks. It partially addresses #794. Using the debug CRT (which entails various hacks too ugly to commit) this set of changes gets rid most of the direct leaks from the jit. Most simple methods now leak ~2K when jitted, down from ~50K. We can also now process the various huge cases that blocked us.

I plan to subsequently rework most of this to use automatic reclamation.

Free up EH region and related info when done reading. Make sure we track the all region list to enable this (the code was set up to do this at one time, but when we went to LLVM naming conventions we introduced a local shadow). Streamline processing of raw eh clauses a bit.

Free up "initial" reader stacks when done reading.

Free up node offset array and related when done reading.

Free up final "working" reader stack when done reading.

Free up the canonical exit offset buffer (if heap allocated) when done readong.

Free up "working" stacks when switching before reading IL for a block.

Free up node worklist after computing initial reachability set.

Free up runtime GC info after examining it.

Stack allocate some debug generation objects so they are cleaned up automatically. Explicitly delete others.

Also, reduce the default stack size. Since we won't be inlining in-reader, and most stacks are small, use 4 as the default size. This saves about 300MB of allocation on `HugeArray`.

Remove exclusions for three tests that now pass. The other three cases also generate correct code but fail at runtime because the generated code requires a huge stack frame (4.7MB in the case of `HugeArray`).

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -1445,7 +1445,6 @@ protected:
   uint32_t CurrentBranchDepth;
 
   // EH Info
-  CORINFO_EH_CLAUSE *EhClauseInfo; // raw eh clause info
   EHRegion *EhRegionTree;
   EHRegionList *AllRegionList;
 
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -119,7 +119,7 @@ private:
   ///
   /// \param DwarfContext Dwarf Context to extract debug info from
   /// \param Size Size of the function we are gathering info from
-  void getDebugInfoForLocals(DWARFContextInMemory *DwarfContext, uint64_t Addr,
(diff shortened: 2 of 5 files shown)


AndyAyersMS:
@erozenfeld PTAL.

erozenfeld:
LGTM.
>> Using the debug CRT (which entails various hacks too ugly to commit)
Do you have an alternative plan for a tool that will allow us to check for new leaks before committing new changes?

AndyAyersMS:
Unfortunately I don't have a good programmatic method for tracking down leaks yet. Hopefully we can find something.

AndyAyersMS:

PR #838, opened 2015-09-11
Clone finally bodies

### LINKED ISSUE #835: Clone non-exception paths out of finally cleanuppads
Eventually we'll want `cleanupcall` so that finally cloning can be handled as an optimization, but in the near-term the plan is to simply duplicate the code in a post-pass in the reader.

JosephTremoulet:
Note: this is a PR for code to go into the EH branch.

The main reader passes generate IR for finally handlers that reflects the
right semantics, but is ill-formed because the `cleanuppad` at the entry
of a finally doesn't dominate its uses at the exits from that finally
(because there is a side-entry for the non-exceptional paths through the
finally).

Add a post-pass (in `readerPostVisit`) that makes a separate clone for the
non-exception version of each finally handler, leaving the `cleanuppad`
def and uses well-formed on the exception-only path.

The long-term plan is to add a `cleanupcall` operator to LLVM that the
non-exceptional paths can use to target the `cleanuppad` directly (and
perform finally cloning as a subsequent optimization in many cases, but
without the need to introduce potentially-explosive cloning for
correctness).  That work is being deferred to simplify/speed-up the
initial WinEH bring-up.

Closes #835.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -1911,6 +1911,24 @@ private:
   ///          outlined.
   llvm::Function *getCurrentFilter();
 
+  /// Copy the IR in each finally so that the exceptional path does not share
+  /// code with the non-exceptional path.
+  void cloneFinallyBodies();
+
+  /// Copy the IR in the given finally so that the exceptional path does not
+  /// share code with the non-exceptional path.
+  ///
+  /// \param FinallyRegion  The region whose IR is to be cloned.
+  void cloneFinallyBody(EHRegion *FinallyRegion);
+
+  /// Remove the unwind edge from the given terminator, indicating that if it
+  /// faults it will unwind to this function's caller rather than a handler in
+  /// this function.  The terminator will be replaced with a new terminator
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
@pgavlin PTAL (or LMK if you won't have a chance).

I created this as a pull request against branch so that the right thing will happen once #831 gets merged, but the right delta to look at is actually to compare this against [that branch](https://github.com/JosephTremoulet/llilc/tree/NewRep), which conveniently is just the single commit 27d499260c12064c616b4b5a57f8bda7eb6f589c.

JosephTremoulet:
This again can't be validated downstream yet, so I posted an example of [C#](https://gist.github.com/JosephTremoulet/da7f683f9dc7faf0f2ef#file-cloning-cs) getting translated to [LLVM IR](https://gist.github.com/JosephTremoulet/da7f683f9dc7faf0f2ef#file-cloning-ll) after this change.

pgavlin:
LGTM. Thanks for pointing out the correct branch to diff against.

PR #839, opened 2015-09-11
Disable type caching across methods in ReadyToRun mode.

### LINKED ISSUE #833: ready to run mainv2 intermittently failing.
We're seeing failures in some legs with this new test.  It's been disabled so the fix is to remove the opt out and make it stable.

erozenfeld:
ReadyToRun mainv2 test has a series of subtests. One of them is TestGrowingStruct.
The scenario is that an app is ReadyToRun-compiled against a version of a struct in a lib.
Then two new fields are added to the struct and the lib is ReadyToRun recompiled.
Then we attempt to run the app without recompiling. What should happen in this case
is that the runtime should detect that the struct layout has changed and the pre-compiled
code for the method in the app is invalid and the method should be JIT-ed.

The reason that wasn't always happening was because we have per-thread caching of types.
Crossgen records structs that the JIT calls getClassSize for and that allows the runtime to do
the detection described above. In this case more than one method needed the info about
that struct but crossgen recorded just the method that happened to call getClassSize on it first.
So if a method was compiled on the same thread as TestGrowingStruct before it and called getClassSize,
crossgen wouldn't record the TestGrowingStruct's dependency on the struct layout.

The fix is not to do type caching across methods in ReadyToRun mode.

Closes #833.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -545,6 +545,19 @@ void GenIR::readerPostPass(bool IsImportOnly) {
     }
   }
 
+  // Crossgen in ReadyToRun mode records structs that method compilation
+  // depends on via getClassSize calls. Therefore, we shouldn't cache type info
+  // across methods.
+  if (JitContext->Flags & CORJIT_FLG_READYTORUN) {
+    // Clearing the cache is only safe if we're at the "top-level" context.
+    // We shouldn't have re-entrant jit requests in a prejit scenario.
+    assert(JitContext->Next == nullptr);
+    ClassTypeMap->clear();
+    ArrayTypeMap->clear();
+    ReverseClassTypeMap->clear();
+    BoxedTypeMap->clear();
+  }
+
   // Cleanup the memory we've been using.
   DBuilder->finalize();
   delete DBuilder;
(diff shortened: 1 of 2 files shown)


erozenfeld:
@russellhadley PTAL

erozenfeld:
@AndyAyersMS PTAL

lib/Reader/readerir.cpp
```diff
@@ -545,6 +545,16 @@ void GenIR::readerPostPass(bool IsImportOnly) {
     }
   }
 
+  // Crossgen in ReadyToRun mode records structs that method compilation
+  // depends on via getClassSize calls. Therefore, we shouldn't cache type info
+  // across methods.
+  if (JitContext->Flags & CORJIT_FLG_READYTORUN) {
```
AndyAyersMS: I think it's only safe to clear out the caches if we're at the "top-level" context. Presumably re-entrant jit requests are unlikely in a prejit scenario so this is probably an academic point, but can you add an assert anyways?

AndyAyersMS:
LGTM modulo one note.

lib/Reader/readerir.cpp
erozenfeld: Good point. Added an assert.

PR #840, opened 2015-09-12
InsertStatepoints: Add exclusion for a test that times out.

swaroop-sridhar:
Two tests time out because some loops take longer to run
with gc-probes inserted within loops.

We need to add an option to the CoreCLR test framework to
specify timeout factor (dotnet/coreclr#1541).

This change adds exclusion for the failing tests until the
above issue is fixed -- so that InsertStaetpoints job is green.


swaroop-sridhar:
@russellhadley  please review

PR #842, opened 2015-09-12
Enable delegate constructor optimization in ReadyToRun mode.

### LINKED ISSUE #841: ReadyToRun: Enable delegate constructor optimization

erozenfeld:
Delegate constructor optimization is implemented via
CORINFO_HELP_READYTORUN_DELEGATE_CTOR helper
and is only available for non-virtual function pointers
resolved at compile time.

Closes #841.

### DIFF
--- a/lib/Reader/reader.cpp
+++ b/lib/Reader/reader.cpp
@@ -5069,8 +5069,29 @@ ReaderBase::rdrCall(ReaderCallTargetData *Data, ReaderBaseNS::CallOpcode Opcode,
                      &ResolvedFtnToken);
 
         CORINFO_METHOD_HANDLE TargetMethod = ResolvedFtnToken.hMethod;
-
-        if (TargetMethod) {
+#ifdef FEATURE_CORECLR
+        {
+          // Do the CoreClr delegate transparency rule check before calling
+          // the delegate constructor
+          CORINFO_CLASS_HANDLE DelegateType = Data->getClassHandle();
+          CORINFO_METHOD_HANDLE CalleeMethod = TargetMethod;
+          mdToken TargetMethodToken = Data->getMethodToken();
+          rdrInsertCalloutForDelegate(DelegateType, CalleeMethod,
+                                      TargetMethodToken);
+        }
+#endif // FEATURE_CORECLR
+        if (Flags & CORJIT_FLG_READYTORUN) {
+          ReaderOperandStack->pop();
+          IRNode *TargetObject = ReaderOperandStack->pop();
+          assert((Data->getClassAttribs() & CORINFO_FLG_VALUECLASS) == 0);
+          IRNode *NewObjThisArg = rdrMakeNewObjThisArg(Data, CORINFO_TYPE_CLASS,
+                                                       Data->getClassHandle());
+          const bool MayThrow = false;
+          callReadyToRunHelper(CORINFO_HELP_READYTORUN_DELEGATE_CTOR, MayThrow,
+                               nullptr, &ResolvedFtnToken, NewObjThisArg,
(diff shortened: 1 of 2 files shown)


erozenfeld:
@AndyAyersMS PTAL

lib/Reader/reader.cpp
```diff
@@ -7158,6 +7161,12 @@ void ReaderBase::readBytesForFlowGraphNodeHelper(
       getCallInfo(&ResolvedToken, nullptr /*constraint*/,
                   CORINFO_CALLINFO_LDFTN, &CallInfo);
 
+      // ReadyToRun has delegate constructor optimization for non-virtual
+      // function pointers resolved at compile time only for now.
+      if ((Flags & CORJIT_FLG_READYTORUN) && (CallInfo.kind != CORINFO_CALL)) {
```
AndyAyersMS: Not sure I understand the impact of "for now" here.... do you mean this optimization is going away, or that in RtR it is currently limited in scope but might become more widely capable?

AndyAyersMS:
LGTM, just one question about a comment.

lib/Reader/reader.cpp
erozenfeld: It's the latter. I re-phrased the comments to make it more clear:

`// Currently ReadyToRun has delegate constructor optimization only for`
 `// non-virtual function pointers resolved at compile time.`

erozenfeld:
There was a bug in my changes: local var Flags in CEE_LDVIRTFTN was shadowing ReaderBase::Flags so the check for ReadyToRun mode was incorrect. Fixed that by renaming the local Flags to CallFlags.

erozenfeld:
@AndyAyersMS PTAL

AndyAyersMS:
Interesting -- this is the second shadowing issue to come up recently. I wonder if there's some frontend warning we can enable to avoid these?

Changes LGTM still.

erozenfeld:
I did a quick search and the closest thing I found is C6246: https://msdn.microsoft.com/en-US/library/eeb7y4c0.aspx
warning C6246: Local declaration of <variable> hides declaration of same name in outer scope. Additional Information: See previous declaration at <location>.

Unfortunately it only catches one local shadowing another local so it won't help in this case. Also, this is not a regular front end warning but one emitted by CodeAnalysis.

I ran CodeAnalysis on reader.cpp and we do get a few instances of C6246 as well as other kinds of warnings. We should probably look at them and see if any are worth fixing.

erozenfeld:
@dotnet-bot test this please

JosephTremoulet:
What about builds for other targets?  Looks like clang and gcc both support a -Wshadow that ought to catch that sort of thing.

erozenfeld:
From https://docs.freebsd.org/info/gcc/gcc.info.Warning_Options.html
```
`-Wshadow'
     Warn whenever a local variable shadows another local variable.
```
so it's similar to C6246 and won't flag a local shadowing a class field.

JosephTremoulet:
I was going by https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

> Warn whenever a local variable or type declaration shadows another variable, parameter, type, class member (in C++), or instance variable (in Objective-C) or whenever a built-in function is shadowed.

and clang sources:

```
  // Determine what kind of declaration we're shadowing.
  unsigned Kind;
  if (isa<RecordDecl>(OldDC)) {
    if (isa<FieldDecl>(ShadowedDecl))
      Kind = 3; // field
    else
      Kind = 2; // static data member
  } else if (OldDC->isFileContext())
    Kind = 1; // global
  else
    Kind = 0; // local
```

PR #845, opened 2015-09-15
Resolve issue #638 and update exclusion.targets

### LINKED ISSUE #638: Assertion: size of stack type should be at least as large as the size of the source type
This pops up during one or two of the new IL tests.

`Assertion failed: (Size <= DesiredSize) && "Size <= DesiredSize", file ..\..\..\..\..\tools\llilc\lib\Reader\readerir.cpp, line 2030`

richardlford:
Closes #638.

The test that was failing with issue #638 no longer fails.
Remote it from the exclusion.targets.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -1540,9 +1540,6 @@
         <ExcludeList Include="$(XunitTestBinBase)\JIT\Regression\CLR-x86-JIT\V1-M12-Beta2\b32613\b32613.cmd" >
              <Issue>13</Issue>
         </ExcludeList>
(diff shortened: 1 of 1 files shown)


richardlford:
@AndyAyersMS, PTAL.

AndyAyersMS:
LGTM.

richardlford:
The failed legs were infrastructure failures and in legs that could not possibly have been affected by this change (which merely removes a test from the exclusion list). Both of the win64 legs, which are the only ones that run tests, passed. So I'm going to go ahead and merge this change.

Issue #821, opened 2015-09-04
Verify we're not skipping methods in the enabled tests


AndyAyersMS:
I asked a couple people and nobody could say for sure if we were skipping methods in the enabled tests. Seems like these days a 100% jit rate is a reasonable expectation.

So, we should make sure we're doing exactly this in our lab tests. @richardlford @russellhadley can one of you take this?

richardlford:
@AndyAyersMS, I'll take this one.

erozenfeld:
We are skipping some methods:
Found 2490 results
2490 tests, 967092 methods, 966528 good, 564 bad (99% good)
[Tail call]                                     492
[ckFinite]                                      38
[Direct unmanaged call]                         26
[unexpected type in load/store primitive]       8

erozenfeld:
I'll take tail call since I already implemented jmp and it should be similar.

AndyAyersMS:
Can you open issues for the others?

erozenfeld:
I opened new issues and updated the existing ones:
[Tail call] #191
[ckFinite] #846 
[Direct unmanaged call] #499
[unexpected type in load/store primitive] #847

AndyAyersMS:
@richardlford @russellhadley would be nice to get this done

Issue #499, opened 2015-04-23
Reader: support direct unmanaged calls.


pgavlin:

erozenfeld:
We hit these in the following CoreCLR tests:

JIT\Methodical\varargs\callconv\gc_ctor_il_d
JIT\Methodical\varargs\callconv\gc_ctor_il_r
JIT\Methodical\varargs\callconv\val_ctor_il_d
JIT\Methodical\varargs\callconv\val_ctor_il_r
JIT\Methodical\varargs\misc\Dev10_615402
JIT\Methodical\varargs\misc\Dev10_615402
JIT\Regression\CLR-x86-EJIT\V1-M12-Beta2\b26323\b26323
JIT\Regression\CLR-x86-JIT\V1-M09.5-PDC\b16423\b16423
JIT\Regression\CLR-x86-JIT\V1-M09.5-PDC\b28901\b28901
JIT\Regression\CLR-x86-JIT\V1-M09.5-PDC\b30838\b30838
JIT\Regression\CLR-x86-JIT\V1-M09.5-PDC\b30864\b30864
JIT\Regression\CLR-x86-JIT\V1-M09.5-PDC\b32374\b32374
JIT\Regression\CLR-x86-JIT\V1-M11-Beta1\b35784\b35784
JIT\Regression\CLR-x86-JIT\V1-M11-Beta1\b36472\b36472
JIT\Regression\CLR-x86-JIT\V1-M11-Beta1\b37598\b37598
JIT\Regression\CLR-x86-JIT\V1-M11-Beta1\b46867\b46867
JIT\Regression\CLR-x86-JIT\V1-M12-Beta2\b31745\b31745
JIT\Regression\CLR-x86-JIT\V1-M12-Beta2\b31746\b31746
JIT\Regression\CLR-x86-JIT\V1-M12-Beta2\b37646\b37646
JIT\Regression\CLR-x86-JIT\V1-M12-Beta2\b41852\b41852
JIT\Regression\CLR-x86-JIT\V1-M13-RTM\b88793\b88793
JIT\Regression\CLR-x86-JIT\V1-M13-RTM\b91248\b91248
JIT\Regression\CLR-x86-JIT\V2.0-Beta2\b409748\b409748

AndyAyersMS:
I'll take a look at this one.

AndyAyersMS:
[...]

PR #851, opened 2015-09-17
Exclude tests with varargs calling convention

AndyAyersMS:
This addresses #821 and #499.

Exclude tests with varargs callees under dotnet/coreclr#1440. Also improve the NYI message to call out varargs differently than native conventions.


AndyAyersMS:
@erozenfeld PTAL

erozenfeld:
LGTM

PR #852, opened 2015-09-18
Run verifier before unreachable code removal.

### LINKED ISSUE #849: Reader: run IR verifier before unreachable code removal
Running the verifier before unreachable code removal would have caught #785.

erozenfeld:
Running the verifier before unreachable code removal would have caught
some of the bugs we had (e.g., #785).

Changes to pass the verifier:
* Add unreachable instruction in place of return during flow graph building.
   Remove the unreachable when processing return instruction in the second pass.
* Type dummy switch selector as int32 so that there is no mismatch with types
   of case values.
* Move DBuilder finalization to readerPostVisit to avoid invalid MDNodes.

Closes #849.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -509,6 +509,8 @@ void GenIR::readerPostVisit() {
   if (NeedsSecurityObject) {
     insertIRForSecurityObject();
   }
+
+  DBuilder->finalize();
 }
 
 void GenIR::readerPostPass(bool IsImportOnly) {
@@ -559,7 +561,6 @@ void GenIR::readerPostPass(bool IsImportOnly) {
   }
 
   // Cleanup the memory we've been using.
-  DBuilder->finalize();
   delete DBuilder;
   delete LLVMBuilder;
 }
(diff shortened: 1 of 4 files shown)


erozenfeld:
@JosephTremoulet PTAL

lib/Reader/readerir.cpp
```diff
@@ -2971,6 +2972,8 @@ FlowGraphNode *GenIR::fgSplitBlock(FlowGraphNode *Block, IRNode *Node) {
 }
 
 void GenIR::fgRemoveUnusedBlocks(FlowGraphNode *FgHead) {
+  bool IsOk = !verifyFunction(*Function, &dbgs());
+  assert(IsOk && "verification failed");
```
JosephTremoulet: Should any of this need to be under `#ifndef NDEBUG` / do you need a dummy use of IsOk to avoid release build warnings?

JosephTremoulet: Also, any reason not to verify the module (as opposed to just the function)?

JosephTremoulet:
Two nits, otherwise looks great, thanks for doing this!

lib/Reader/readerir.cpp
erozenfeld: Good point, added #ifndef NDEBUG.
I don't think the reader should make the assumption that we have one function in the module (and if there were more than one we wouldn't want to verify all functions every time). We still verify the module in LLILCJit::readMethod.

JosephTremoulet: > I don't think the reader should make the assumption that we have one function in the module (and if there were more than one we wouldn't want to verify all functions every time). We still verify the module in LLILCJit::readMethod.

Makes sense.  Green.

PR #853, opened 2015-09-18
Handle unmanaged pointer types passed to mkrefany

### LINKED ISSUE #592: Bad interaction between ldsflda and mkrefany

AndyAyersMS:
Closes #592.

Allow an unmanaged pointer as the pointer value in a mkrefany, by casting it to a managed pointer.


AndyAyersMS:
@erozen PTAL

erozenfeld:
LGTM

PR #854, opened 2015-09-18
Translate filters to catches in reader

JosephTremoulet:

Lowering for filters (as filters) is going to lag behind lowering for the other handler types.  This is a change to have filters masquerade as catches in the meantime -- the catch encompasses both the filter and the filter-handler, with an explicit rethrows if the `filter portion returns 0.  This is not fully "as-is", because it effectively delays the evaluation of filter expressions from the first pass of unwinding to the second.  Aside from that, however, it is a faithful representation of filter semantics.

I've created a [gist](https://gist.github.com/JosephTremoulet/bd635de873e90de9b710) to show the effect of the transform.  It includes a [C# input](https://gist.github.com/JosephTremoulet/bd635de873e90de9b710#file-input-cs), the [IR generated with filter outlining](https://gist.github.com/JosephTremoulet/bd635de873e90de9b710#file-outputwithfilteroutlining-ll) (as before this change), and the [IR generated with filters translated to catches](https://gist.github.com/JosephTremoulet/bd635de873e90de9b710#file-outputwithfiltertransform-ll) (as with this change).

The change is split into four commits:
 - the first two (97fe976 and f7316c0) are actually small bugfixes to the filter outlining that were needed to create the 'before' part of the gist
 - the next (7fa22f5) is the meat of the change to produce catches instead of outlining when encountering a filter
 - the last (d7cdeb3)  is a rote change almost entirely composed by reverting chunks to match master, renaming `RootFunction` back to `Function` and the like.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -1531,10 +1531,6 @@ private:
   ///        the thread pointer.
   void insertIRForUnmanagedCallFrame();
 
-  /// \brief Apply any function-level attributes needed to indicate an
-  ///        unmanaged call, on root function and any outlined filters.
-  void setAttributesForUnmanagedCallFrame(llvm::Function *F);
-
   /// \brief Create the @gc.safepoint_poll() method
   /// Creates the @gc.safepoint_poll() method and insertes it into the
   /// current module. This helper is required by the LLVM GC-Statepoint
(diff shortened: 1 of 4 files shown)


JosephTremoulet:
@pgavlin PTAL

pgavlin:
Looking at the sample now.

pgavlin:
LGTM.

PR #855, opened 2015-09-18
Remove vestigial EH/Region-related code

JosephTremoulet:
Delete a bunch of code left over from a previous EH tracking regime, which
we won't be using going forward.


JosephTremoulet:
@AndyAyersMS PTAL

AndyAyersMS:
Nice to see this getting cleaned up.

In `ReaderBase`, I think you've removed the last significant use of `FgGetRegionCanonicalExitOffsetBuff` so can you finish off removing this class member entirely?

JosephTremoulet:
> In ReaderBase, I think you've removed the last significant use of FgGetRegionCanonicalExitOffsetBuff so can you finish off removing this class member entirely?

I thought [I did](https://github.com/dotnet/llilc/pull/855/files#diff-ae787b5689435936b84c345640611102L1950), what am I missing?

AndyAyersMS:
You did -- I got confused looking at diffs and didn't see where you had removed the remaining references.

LGTM.

PR #859, opened 2015-09-21
Some retargeting support and warnings cleanup

AndyAyersMS:
For headers in "fixed" locations like the ones we get from the CLR, suppress CMake's desire to add cross-targeting prefixes.

Add some Arm64 register mappings for the GC info.

Fix warnings from Clang that were not caused by `-Wnon-virtual-dtor` or `-Wunknown-pragma`.


AndyAyersMS:
@JosephTremoulet PTAL

JosephTremoulet:
LGTM

AndyAyersMS:
Looks like the failure is unrelated. Going to go ahead and merge.

Issue #860, opened 2015-09-22
New interop test causes assertion in the runtime


russellhadley:
Under the Win64 check build we're hitting an exception with the newly added interop test.  This issue is to track investigation.  To unblock testing I'm going to add an exclusion that references this issue.

Test path is: tests\src\Interop\NativeCallable\NativeCallableTest.cs

<snip>
[...]

PR #861, opened 2015-09-22
Exclude new Interop test causing an assert in llilc testing

russellhadley:
Until we have a chance to investigate exclude this new test from our testing.
Issue is tracked as #860


russellhadley:
@adiaaida PTAL.  Didn't do local testing so it's gated on the lab going back green.

russellhadley:
Lab is green.  Going ahead with merge to unblock lab.

PR #863, opened 2015-09-22
Add hooks for loadfield via helper and storefield via helper

adiaaida:
* Change rdrCallFieldHelper to return an IRNode: return Destination for
  struct fields and the call instruction for others field types for
  load. Return the call for store, but it is not used by store field.
* Add support for LoadField via helper: For struct fields, we want to
  create a temporary IRNode, for all other types, we want to get a
  nullValue of the field type.
* Add call to rdrCallFieldHelper for store field via helper.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -2345,7 +2345,7 @@ private:
                                      IRNode **ThisPtr);
 
 public:
-  void rdrCallFieldHelper(
+  IRNode *rdrCallFieldHelper(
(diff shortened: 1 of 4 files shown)


adiaaida:
@russellhadley or @erozenfeld PTAL

adiaaida:
@dotnet-bot test this please

adiaaida:
@dotnet-bot test this please

Issue #865, opened 2015-09-23
relocation processing broken for ELF


AndyAyersMS:
The code in `ObjectLoadListener::recordRelocations` needs to handle ELF `.rela.text` sections where the addend is part of the relocation and not read from the relocation target.

This appears to require that we use an `elf_section_iterator` to enumerate the symbols, so likely this code needs to be conditional on what type of object file is produced.

AndyAyersMS:
The current code crashes doing the addend lookup. This happens because we're trying to read the backing bytes from the original section, not its relocated image; the former does not get loaded, so the loaded address lookup returns 0.

The fix is to use `RelocatedSection` for the load address computation, instead of `SI`. I need to look at the coff path and find out if this works ok there too. Since the addend backing field is readable and tends to be zero (as does the addend in the relocation) this works for now without any special handling for addends.

Beyond that there are other special section names that must be skipped. I tried coding this up based on section attributes (basically we only report text and read-only data directly to the EE, so no other section relocation processing should be needed) but that didn't seem to work. So instead I've extended the section name logic. Seems like there should be a better way.

Beyond that we get running code and crash handling System.Buffer.Memmove.

PR #867, opened 2015-09-24
Some fixes for ELF relocation processing

AndyAyersMS:
Addresses #865.

Make sure to use the relocated section for determining final addresses.

Skip some elf-specific sections that contain debut and unwind info when processing relocations.


AndyAyersMS:
@erozenfeld or @JosephTremoulet PTAL

erozenfeld:
LGTM

PR #868, opened 2015-09-24
Change casing of options text to be more consistent and match CoreCLR

### LINKED ISSUE #866: look into case-sensitivity of environment variables
When running on Linux it took me a while to figure out the proper casing for the various ComPLUS environment variables. Worse than that, LLILC internally uses different casing. So one must for example do
```
export COMPlus_gcConservative=1          // for coreclr
export COMPlus_GCCONSERVATIVE=1          // for llilc
```
On windows, none of this matters since environment variables are not case sensitive. We should standardize.

AndyAyersMS:
Closes #866.

Casing of environment variables doesn't matter on Windows but does on Linux. Make sure we're consistently using `COMPlus` as the prefix and that the suffix is generally camel-cased. There are two classes of exceptions:
* `COMPlus_gcConservative` is specified that way by the CoreCLR.
* LLILC specific options (ones that the CoreCLR doesn't look for or care about) are generally all uppercased, eg `COMPlus_DUMPLLVMIR`.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -155,7 +155,7 @@ ICorJitCompiler *__stdcall getJit() {
     sys::AddSignalHandler(&LLILCJit::signalHandler, LLILCJit::TheJit);
 
     // Allow LLVM to pick up options via the environment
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@rhadley PTAL

russellhadley:
Green.

PR #870, opened 2015-09-24
Removing return refany NYI.

jashook:
Refany wil be handled like a normal struct, returned byref. While
__arglist will now throw an NYI because CoreCLR does not support it.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -270,8 +270,8 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
   CORINFO_SIG_INFO Sig;
   getMethodSig(MethodHandle, &Sig);
 
-  if (Sig.retType == CORINFO_TYPE_REFANY) {
-    throw NotYetImplementedException("Return refany");
+  if (this->MethodSignature.isVarArg()) {
+    throw NotYetImplementedException("Vararg method");
   }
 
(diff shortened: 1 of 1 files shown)


jashook:
@AndyAyersMS @erozenfeld PTAL

lib/Reader/readerir.cpp
```diff
@@ -5630,11 +5626,11 @@ IRNode *GenIR::genCall(ReaderCallTargetData *CallTargetInfo, bool MayThrow,
     throw NotYetImplementedException("Direct unmanaged call");
   }
 
-  CallArgType ResultType = Signature.getResultType();
-  if (ResultType.CorType == CORINFO_TYPE_REFANY) {
-    throw NotYetImplementedException("Return refany");
+  if (this->MethodSignature.isVarArg()) {
+    throw NotYetImplementedException("Vararg method");
```
AndyAyersMS: You should put this check up around line 273 (didn't realize there were 2 places we looked for return refany).

Basically. we want to reject requests to jit varargs methods as soon as we know we have one, rather than spending time jitting parts of it only to bail out once we see it making a call. Aside from being a bit more efficient (not that that matters much if we're bailing out) there's the off chance that a varargs method doesn't make any calls at all.

jashook:
Makes sense, added.

AndyAyersMS:
Now that you have the check running early at line 273, you don't need the one later on around line 5629.

jashook:
Updated

AndyAyersMS:
LGTM

russellhadley:

russellhadley:

PR #881, opened 2015-10-02
Update EH table parsing to new format

JosephTremoulet:
The extra information passed in xdata is arranged differently now that
funclets aren't outlined into separate functions in LLVM IR.


JosephTremoulet:

Tests with EH fail in some debug-info-related assertion.  This code, if paired with a work-around for that and the LLVM changes in Microsoft/llvm#95, can successfuly transfer control into and out of a catch body, so this is forward progress and I see no reason not to go ahead and merge this into the EH branch while sorting out the other issues.

(Note that various things like referencing local variables or running a garbage collection while an exception handler is running don't work yet).

JosephTremoulet:
@AndyAyersMS and/or @pgavlin, PTAL

AndyAyersMS:
LGTM

PR #882, opened 2015-10-02
Reorder varargs NYI check

JosephTremoulet:
Delay it until after we've determined if the method is varargs.


JosephTremoulet:
@AndyAyersMS PTAL.  Given that the bug this fixes arbitrarily punts methods back to RyuJit, I'd think I should check in w/o waiting for the lab job, agree?

AndyAyersMS:
LGTM, go ahead and merge if you like.

PR #883, opened 2015-10-03
Skip relocation for personality routine in xdata

JosephTremoulet:
The runtime_function field in the unwind_info in xdata gets created with a
relocation for the personality function.  The personality function we
specify is a synthetic declaration because this works well with the string
matching on the name that LLVM does and the EE doesn't actually use the
pointer in the xdata (it overwrites it with a pointer to the correct
routine before passing the xdata to the OS).  This change recognizes that
relocation and passes on processing it.


JosephTremoulet:
Note: this is a PR for a change into the EH branch.

@AndyAyersMS or @erozenfeld PTAL

erozenfeld:
LGTM

PR #884, opened 2015-10-06
Update debug info emission to handle funclets

JosephTremoulet:
The object gets a function symbol for each funclet, so we need to sum
their sizes rather than trying to report each in turn.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -543,8 +543,13 @@ void ObjectLoadListener::getDebugInfoForObject(
   // want to also be able to eventually extract WinCodeView information as well
   DWARFContextInMemory DwarfContext(DebugObj);
 
-  // Use symbol info to iterate functions in the object.
-  // TODO: This may have to change when we have funclets
+  // Use symbol info to find the function size.
+  // If there are funclets, they will each have separate symbols, so we need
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
Note: this is a PR for a change into the EH branch.  With this change and Microsoft/llvm#98, we can successfully execute simple try/catches using the EH branch with no workarounds.

@AndyAyersMS or @adiaaida PTAL.

lib/Jit/LLILCJit.cpp
```diff
@@ -560,69 +565,74 @@ void ObjectLoadListener::getDebugInfoForObject(
...
+      // The main function is always laid out first
+      Addr = SingleAddr;
+    }
+    Size += SingleSize;
+  }
 
-    uint32_t LastDebugOffset = (uint32_t)-1;
-    uint32_t NumDebugRanges = 0;
-    ICorDebugInfo::OffsetMapping *OM;
+  uint32_t LastDebugOffset = (uint32_t)-1;
```
JosephTremoulet: I don't know why the diff isn't figuring this out, but fyi everything from here down is just an indentation change (and removing the closing brace where there used to be a loop).

AndyAyersMS:
LGTM

adiaaida:
LGTM too

PR #885, opened 2015-10-07
Set debug info version module flag

JosephTremoulet:
Explicitly set the module's debug info version to the current one.
Without this, writing out bitcode from LLILC and reading it back in will
strip the debug info, because the LLVM parsers check for the debug info
version and strip debug metadata if it does not match the version that the
parser was built with.


JosephTremoulet:
@adiaaida PTAL

russellhadley:
Green.

adiaaida:
LGTM

JosephTremoulet:

PR #886, opened 2015-10-09
Update llilc_roslyn_stage script

richardlford:
These changes enable Roslyn to build its CscCore project
while being jitted by llilc on CoreClr. This currently needs to
use the richardlford/llilc_branch version of Roslyn in order
to avoid throwing an exception when looking for
System.IO.MemoryMappedFiles.

Previously was using c:/Python34, but that is no longer
installed on lab machines. The script that it invokes
will work with either Pythone 3.4 or 2.7 so just use
the Python on the path.

Add buildtype as an argument to llilc_roslyn_stage.cmd
so it will work with debug, checked, or release builds.

Revise for new Roslyn strategy for running on coreclr.

The Roslyn analyzers sometimes throw task cancellation
exceptions. Since LLILC cannot handle these yet,
inhibit their use.

### DIFF
--- a/test/llilc_roslyn_stage.cmd
+++ b/test/llilc_roslyn_stage.cmd
@@ -8,21 +8,21 @@ REM buildsubdir is the subdirectory of the workspace where LLVM (and LLILC) were
 setlocal
 set WORKSPACE=%1
 set buildsubdir=%2
-set stagedir=%3
+set buildtype=%3
+set stagedir=%4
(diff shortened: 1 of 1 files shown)


richardlford:
@russellhadley, please review.

richardlford:
I pushed one other correction to the script.

richardlford:
There is an updated version on a different branch, so I'm closing this and will submit a new PR.

PR #887, opened 2015-10-09
Fix or suppress warnings building on Linux.

AndyAyersMS:
Suppress warnings about virtual dtors since the CLR header patterns are not clean.
Disable warnings about unknown pragmas since we use #pragma region.
Disable warnings about covered switch defaults since we use this to detect upstream enum changes.

Add a missing dtor to ReaderStack.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,9 +291,31 @@ if ( NOT LLVM_TARGET_IS_CROSSCOMPILE_HOST )
     endif ()
 
     check_cxx_compiler_flag("-Werror -Wnested-anon-types" CXX_SUPPORTS_NO_NESTED_ANON_TYPES_FLAG)
-    if( CXX_SUPPORTS_NO_NESTED_ANON_TYPES_FLAG )
+    if (CXX_SUPPORTS_NO_NESTED_ANON_TYPES_FLAG)
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-nested-anon-types" )
     endif()
+
+    # remove -Wnon-virtual-dtor for now, since CLR headers are not clean
+    string(REGEX REPLACE "-Wnon-virtual-dtor" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+    # if supported, add -Wno-unknown-pragmas, since we use pragma region in LLILC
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@JosephTremoulet PTAL

CMakeLists.txt
```diff
@@ -291,9 +291,30 @@ if ( NOT LLVM_TARGET_IS_CROSSCOMPILE_HOST )
...
+    # remove -Wnon-virtual-dtor for now, since CLR headers are not clean
+    string(REGEX REPLACE "-Wnon-virtual-dtor" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+    # if supported, add -Wno-unknown-pragmas, since we use pragma region in LLILC
+    check_cxx_compiler_flag("-Werror -Wno-unknown-pragmas" CXX_SUPPORTS_NO_UNKNOWN_PRAGMAS_FLAG)
+    if (CXX_SUPPORTS_NO_UNKNOWN_PRAGMAS_FLAG)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-pragmas")
+    endif()
+
+    # if supported, add -Wcovered-switch-default, since we use default NYI's to catch upstream contract changes
```
JosephTremoulet: Should this say either 'add -Wno-covered-switch-deafult' or 'remove -Wcovered-switch-default'?

JosephTremoulet:
The LLVM coding guidelines [suggest](http://llvm.org/docs/CodingStandards.html#don-t-use-default-labels-in-fully-covered-switches-over-enumerations) having -Wcovered-switch-default ON because they also suggest having -Wswitch on, and -Wswitch will let you catch upstream enum changes when building the compiler (as opposed the the NYI catching them while running the compiler) so long as you don't muzzle it with a default case.  So arguably if you're regularly building with clang or gcc you'll have a nicer experience if you don't suppress this.  I say arguably because things get awkward if you wanted to put `llvm_unreachable` in the defualt case.  Either way you want to go here is fine with me, just wanted to point out the counter-argument since I don't think it's an obvious one.

LGTM, either as-is or with covered-switch unsuppress and the violations "fixed".

CMakeLists.txt
AndyAyersMS: Changed to be `-Wno ...`

AndyAyersMS:
`-Wswitch` looks like a nice option. We could enable MSVC's C4602 to get similar checking.

However, I'll just go with this for now.

PR #889, opened 2015-10-14
Update llilc_roslyn_stage.cmd and llilc_run.py scripts

richardlford:
With these changes the lab job to run Roslyn building Roslyn
using LLILC succeeds.

llilc_roslyn_stage.cmd changes:

- Previously llilc_roslyn_stage.cmd was using c:/Python34,
  but that is no longer installed on lab machines. The script
  that it invokes will work with either Pythone 3.4 or 2.7 so
  just use the Python on the path.
- Add buildtype as an argument to llilc_roslyn_stage.cmd
  so it will work with debug, checked, or release builds.
- Revise for new Roslyn strategy for running on coreclr.
- The Roslyn analyzers sometimes throw task cancellation
  exceptions. Since LLILC cannot handle these yet,
  inhibit their use.

llilc_run.py changes:

Use lists to accumulate arguments for the command.
rather than strings. Using strings seems to cause
problems on some machines (observed for Windows Server 2012 R2)
(a quoting problem) whereas using lists the arguments
are converted to a string by the python subprocess module.
This avoided the problem.

### DIFF
--- a/test/llilc_roslyn_stage.cmd
+++ b/test/llilc_roslyn_stage.cmd
@@ -8,21 +8,21 @@ REM buildsubdir is the subdirectory of the workspace where LLVM (and LLILC) were
 setlocal
 set WORKSPACE=%1
 set buildsubdir=%2
-set stagedir=%3
+set buildtype=%3
+set stagedir=%4
 cd %WORKSPACE%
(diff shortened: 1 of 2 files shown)


richardlford:
@russellhadley, PTAL.

russellhadley:
LGTM  Just to be clear the .cmd file enables the stage build for windows in the lab but if we want to run this cross-platform we will need to make a parallel .sh version right?

richardlford:
@russellhadley, correct.

PR #891, opened 2015-10-15
Account for try regions that follow their handlers

JosephTremoulet:
It's legal MSIL for a protected region's handler(s) to precede it, but
fgSwitchRegion wasn't expecting this.  Add a property of try regions that
indicates which of the try region itself or its handlers is lexically
first (called its "EntryRegion"), computed during EH Region Tree
construction, which fgSwitchRegion can use to make sure it calls
fgEnterRegion exactly once for each try, and before switching to any
handler of the try.

### DIFF
--- a/include/Reader/reader.h
+++ b/include/Reader/reader.h
@@ -975,6 +975,10 @@ void rgnSetFilterHandlerRegion(EHRegion *EhRegion, EHRegion *Handler);
 EHRegion *rgnGetFinallyTryRegion(EHRegion *FinallyRegion);
 mdToken rgnGetCatchClassToken(EHRegion *CatchRegion);
 void rgnSetCatchClassToken(EHRegion *CatchRegion, mdToken Token);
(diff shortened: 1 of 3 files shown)


JosephTremoulet:
@AndyAyersMS or @pgavlin PTAL.  This fixes several Directed/leave tests in the EH branch (and since the relevant bits are already in master I decided to PR it straight to master).

AndyAyersMS:
LGTM.

Wonder if we could use this to simplify/check the logic I added back in #812 ?

JosephTremoulet:
> Wonder if we could use this to simplify/check the logic I added back in #812 ?

This change makes sure that we'll fgEnterRegion the try before we process the body of the finally, but #812 fixed an issue with assuming that we'd see the leave instructions within the try body before we process the body of the finally -- that assumption is still invalid, since fgEnterRegion doesn't visit the body of the try, just will set up the EH pads.  So I think we still need your change from #812.

Issue #29, opened 2015-02-17
GC:Handle pinned pointers


AndyAyersMS:
The CoreCLR GC requires that pinned pointers be reported specially.

swaroop-sridhar:
To support pinning, we’ll need to add some Pinning intrinsics to LLVM, demarcate the pinning boundaries:
i.	Have GCCast like controlled casts that allow for integer <-> pointer conversion
ii.	Ensure that the pinned pointers do not move beyond the safepoints outside the pinned zone
iii.	Note the slots containing the pinned pointers, and report them to the GC.
Details of this design are still to be worked out

JosephTremoulet:
> Have GCCast like controlled casts that allow for integer <-> pointer conversion

I would have thought that:
 1. we could use addrspacecast for unmanaged pointer <-> gc pointer conversions
 2. we could use inttoptr/ptrtoint for int <-> unmanaged pointer conversions
 3. we could decompose int <-> gc pointer conversions as int <-> unmanaged pointer <-> gc pointer conversions

Does that not work?

JosephTremoulet:
> add some Pinning intrinsics to LLVM, demarcate the pinning boundaries

This assumes we can identify the "pinning boundaries", which are stores to locals with the `pinned` attribute.  It seems [the ECMA spec](http://www.ecma-international.org/publications/standards/Ecma-335.htm) specifically allows pinned locals of type `O` (II.23.2.9 paragraph 2), and presumably such locals could have their address taken and a conservative implementation would have to assume that any `stind` via managed pointer to object reference might modify a pinned local.

I would be ok with assuming this doesn't happen (or at least isn't guaranteed to have optimizations allow for it to happen), and I think C# only ever generates pinned locals of type `&` (which can't have their address taken), and in turn assuming that modifications of pinned locals only happen at `stloc`s that directly reference them, but I think it's worth make a conscious choice about that.

It also might be worth checking what other JITs have done, I wouldn't be surprised if they effectively make the same assumption.

JosephTremoulet:
> Ensure that the pinned pointers do not move beyond the safepoints outside the pinned zone

Likewise for any dereferences of unmanaged pointers that we can't prove don't point to the pinned objects.  I'd imagine we'll be assuming conservative aliasing for such unmanaged pointer dereferences and so can constrain these dereferences by modeling the pin operation with wide enough side-effects, but it'll be worth making sure.

swaroop-sridhar:
[...]

PR #893, opened 2015-10-16
Precise GC: Handle GcAggregates

### LINKED ISSUE #33: GC: reporting for GC references from non-SSA references

swaroop-sridhar:
Report GcInfo for GcAggregates and Pinned pointers

Issue #33
The stackmap generated via Statepoints only handles SSA values,
and therefore does not track pointers contained within GC-aggregates.
To circumvent this problem,
(1) Gc-aggregate stack allocations are marked frame-escaped to
    keep the allocation live throughout the function.
(2) We report all pointers within GC-aggregates as untracked.

Issue #29
Pinned pointers are handled similarly.

Issues  #766, #767, #768
GS Cookie, Security Object, and Generics Context are tracked in the GcInfo
and escaped. They are not reported in the encoder because we need some 
additional information.

Testing:

NGEN of MsCorlib pass when built for precise GC.

All CoreCLR tests passed when built for Precise GC
(Statepoints inserted and GcInfo generated) but run with Conservative GC.
http://dotnet-ci.cloudapp.net/job/dotnet_llilc_release_win64_prtest/1111/

Previously passing CoreCLR tests pass with Precise GC (108 tests failed).
http://dotnet-ci.cloudapp.net/job/dotnet_llilc_release_win64_prtest/1112/

Notes:

1) GcInfo, Emitter and Recorder should be separated to different files.
   Will do in a separate checkin (to help view diffs for review)
2) DoInsertStatepoints can be enabled by default (while running with
   Conservative GC). Will do this in a separate change.

Fixes #33, #29


swaroop-sridhar:
@AndyAyersMS  please review.

lib/Reader/readerir.cpp
```diff
@@ -283,6 +282,10 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
       ABIMethodSignature(MethodSignature, *this, *JitContext->TheABIInfo);
   Function = ABIMethodSig.createFunction(*this, *JitContext->CurrentModule);
 
+  if (JitContext->Options->DoInsertStatepoints) {
+    GcFuncInfo = JitContext->GcInfo->newGcInfo(Function);
+  }
+
```
AndyAyersMS: It seems to me it would be cleaner to just always create a GcFuncInfo. Then you could remove a bunch of conditional logic below....

lib/Reader/readerir.cpp
```diff
@@ -460,36 +463,21 @@ void GenIR::readerPostVisit() {
...
-
-    for (const BasicBlock &BB : *Function) {
-      for (const Instruction &Instr : BB) {
-        if (isManagedAggregateType(Instr.getType())) {
-          throw NotYetImplementedException(
-              "NYI: Precice GC for Managed-Aggregate values");
-        }
+  if (JitContext->Options->DoInsertStatepoints) {
+    SmallVector<Value *, 4> EscapingLocs;
+    GcFuncInfo->getEscapingLocations(EscapingLocs);
```
AndyAyersMS: I think we want a general capability to handle multiple escapes, no matter what the GC situation is.

lib/Reader/readerir.cpp
```diff
@@ -538,29 +526,26 @@ void GenIR::insertIRToKeepGenericContextAlive() {
...
+    LLVMBuilder->SetInsertPoint(InsertPoint->getNextNode());
+    Value *FrameEscape = Intrinsic::getDeclaration(JitContext->CurrentModule,
+                                                   Intrinsic::localescape);
+    Value *Args[] = {ContextLocalAddress};
+    const bool MayThrow = false;
+    makeCall(FrameEscape, MayThrow, Args);
   }
+
+  // Don't move TempInsertionPoint up since what we added was not an alloca
+  LLVMBuilder->restoreIP(SavedInsertPoint);
```
AndyAyersMS: This is the kind of code that could be streamlined if there was always a GcFuncInfo around.

AndyAyersMS:
I did an initial pass and left a few comments. I want to go back over it again in more detail, but may not get to it right away.

lib/Reader/readerir.cpp
```diff
@@ -777,9 +750,13 @@ void GenIR::createSym(uint32_t Num, bool IsAuto, CorInfoType CorType,
...
-  AllocaInst *AllocaInst = LLVMBuilder->CreateAlloca(
-      LLVMType, nullptr,
-      UseNumber ? Twine(SymName) + Twine(Number) : Twine(SymName));
+  AllocaInst *AllocaInst =
+      createAlloca(LLVMType, nullptr,
+                   UseNumber ? Twine(SymName) + Twine(Number) : Twine(SymName));
+
+  if (IsPinned && JitContext->Options->DoInsertStatepoints) {
+    GcFuncInfo->recordPinnedSlot(AllocaInst);
+  }
```
AndyAyersMS: Could this be just:
```
if (IsPinned) {
  GcFuncInfo->recordPinnedSlot(AllocaInst);
}
```

lib/Reader/readerir.cpp
```diff
@@ -723,6 +690,16 @@ void GenIR::insertIRForUnmanagedCallFrame() {
 //
 //===----------------------------------------------------------------------===//
 
+AllocaInst *GenIR::createAlloca(Type *T, Value *ArraySize, const Twine &Name) {
+  AllocaInst *AllocaInst = LLVMBuilder->CreateAlloca(T, ArraySize, Name);
+
+  if (JitContext->Options->DoInsertStatepoints && GcInfo::isGcAggregate(T)) {
+    GcFuncInfo->recordGcAggregate(AllocaInst);
+  }
```
AndyAyersMS: Could this be just
```
if (GcInfo::isGcAggregate(T)) {
  GcFuncInfo->recordGcAggregate(AllocaInst);
}
```

AndyAyersMS:
Took another look and left two small notes. Otherwise LTGM. 

Very nice to see this....!

lib/Reader/readerir.cpp
swaroop-sridhar: I left the DoInsertStatepoints check here because even if the information is recorded on GcFuncInfo, it will not be reported for conservative GC (GcInfoEmitter::needsPointerReporting() will be false).

Conservative GC should not need this reported, because it must treat all those values as pointers already.

But I can remove the check, if you think its better to simplify the reader.

AndyAyersMS: I'm hoping one day soon that `DoInsertStatepoints` becomes the default, so I think I'd rather not have it be prominent in the readerir.cpp code.

PR #894, opened 2015-10-16
Stop passing a DIFile to createSubroutineType

JosephTremoulet:
This parameter was unused and so was removed from the method in r250374.

This requires a paired change to the MS branch of LLVM; http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1188/console has the test with both changes together.

@adiaaida or @russellhadley PTAL.


JosephTremoulet:
Microsoft/llvm#103 is the paired LLVM change

JosephTremoulet:
Lab job passed.

adiaaida:
LGTM

PR #895, opened 2015-10-19
Fix some typos and update the gc struct section to reflect the new plan

AndyAyersMS:
@swaroop-sridhar PTAL


swaroop-sridhar:
Green, thanks.

AndyAyersMS:
No need to wait for CI, since this is just a doc change.

PR #896, opened 2015-10-21
Insert failfast calls at exception handler entries

JosephTremoulet:
This will make it easier to triage failures due to unsupported EH
behavior.

Also add a setting that overrides this behavior to facilitate EH testing.


JosephTremoulet:
@russellhadley PTAL.

russellhadley:
LGTM.  Bring it on.  :)

PR #897, opened 2015-10-21
EH RI

JosephTremoulet:
Merge the EH branch back into master.

At a high level, the EH status is still "code with EH constructs compiles cleanly, but attempts to catch exceptions may not work correctly at run-time".  The differences are:
 - Code "compiling cleanly" now includes generating IR for all handlers, pushing it through the entire compiler pipeline, generating EH tables and passing them to the runtime.  Previously it meant stripping the handler code in the reader.
 - Failfast calls are now inserted into handler entries, so diagnosing failures due to lack of EH support becomes easier.
 - Things are stable enough now that it makes sense to do the rest of the EH bring-up with incremental changes off master, rather than in a separate long-lived branch.


JosephTremoulet:
The individual changes were all PRed when they went into the EH branch.

@AndyAyersMS PTAL.

JosephTremoulet:
Forgot to mention:
 - NGEN and InsertStatepoints tests still pass, except that JIT\Regression\VS-ia64-JIT\V1.2-M02\b28158\b28158 with its 16,732 catch clauses starts timing out.
 - Precise GC tests have some additional failures, but @swaroop-sridhar and I have looked at them and believe them to be due to issues already existing in the mainline.

JosephTremoulet:
This didn't kick off CI tests... The code is identical to the code from the last PR into the EH branch, and [its test](http://dotnet-ci.cloudapp.net/job/dotnet_llilc_prtest/1201/) passed.

swaroop-sridhar:
Can you please add an exclusion for b28158 under the `$(COMPLUS_INSERTSTATEPOINTS)` condition in `exclusion.targets`?

JosephTremoulet:
> Can you please add an exclusion for b28158 under the `$(COMPLUS_INSERTSTATEPOINTS) condition` in exclusion.targets? to keep the InsertStatepoints job green.

Sure; updated.

AndyAyersMS:
I'm quite happy to see this merge. 

Looked through the code a bit but not in detail. Am willing to trust that suitable rigor was applied along the way.

Do you think there is value in preserving branch history here, or should you rebase and squash?

JosephTremoulet:
I'd prefer to have the history as-is;
 - I think it will make any future history searches more likely to find the relevant details
 - I believe we'll never find ourselves trying to bisect through these changes
 - even if we would do that, reconstructing a bisectable rebased history would be an effort that doesn't seem worth the opportunity cost
 - neither squashing into a single monolithic commit not parceling out a non-bisectable rebased history by splitting out certain functions at a time gives you any information that isn't just as easily found in the diff against previous master, the header comments, and/or the non-rebased commits that git blame will find this way.

For Microsoft:LLVM, on the other hand, while the EH and MS branches now have identical sources, I considered doing a merge there just to keep the history from the EH branch, but after looking at the specific commits that would preserve, decided there wasn't anything in there worth preserving.

AndyAyersMS:
I don't feel strongly either way, just wanted your perspective.

LGTM.

test/llilc_run.py
```diff
@@ -103,6 +104,7 @@ def main(argv):
     parser.add_argument('-n', '--ngen', help='use ngened mscorlib', default=False, action="store_true")
     parser.add_argument('-p', '--precise-gc', help='test with precise gc', default=False, action="store_true")
     parser.add_argument('-v', '--verbose', help='echo commands', default=False, action="store_true")
+    parser.add_argument('-e', '--eh', help='enable exception handlers to run (as opposed to failfast)', default=False, action="store_true")
```
swaroop-sridhar: Shouldn't the llilc_runtest.py have the --eh option as well?

JosephTremoulet: Oops.  Yeah, good catch, thanks.  Will PR an update.

PR #899, opened 2015-10-22
Add --eh flag to llilc_runtest.py

JosephTremoulet:
This mirrors the --eh flag of llilc_run.py; it stops LLILC from inserting
failfasts in handler entries, so that the generated handler code can
execute.


JosephTremoulet:
@swaroop-sridhar PTAL

swaroop-sridhar:
Green, thanks.

PR #902, opened 2015-10-24
GCInfo: Report Stack Alloca'ed GC Pointers

### LINKED ISSUE #900: GC: Report Alloca's GC Pointers as Untracked
We need to report as untracked all any gc-pointers we’d alloca on the stack -- not just the pointers within GC-aggregates.

Since the alloca’ed value is treated as memory and not SSA, it goes unreported via statepoints.
For example, the “this” pointer goes unreported leading to weird failures.

swaroop-sridhar:
This change modifies LLILC's GcInfo tracking to
report all stack allocation of GC type (not in SSA)
as untracked. This including GC pointers allocated
on the stack frame via Alloca instructions, and not just
GC aggregates.

This change also has some additional fixes:
(1) Zero Initialize all GC value allocations, and not just
    the set of locals in the Reader post-pass.
(2) Report Generics Context slot correctly -- only report
    the slot as a gc-pointer if it is a GENERIC_CONTEXTPARAM_THIS
(3) Supports pinned aggregates -- where all contained pointers
    are reported as pinned.

Implementation wise, there are some improvements to GcInfo:
GcFuncInfo now has a Alloca->AllocaInfo map, instead of
Alloca->Offset mapping. This helps simplify some interfaces
(like reporting special slots in the reader) and the implementation.
All alloca->offset mappings are stored within the AllocaMap,
instead of separate fields for GsCookie, SecurityObject, etc.

Testing:
(1) All CoreCLR tests pass with Precise GC:
http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1219
The two timeouts in b68361 and 113574 are expected -- caused by
insertion of gc-polls on loop-back edges.
(2) NGEN of MsCorlib passes with Precise GC
(3) CoreCLR tests continue to pass with Conservative GC:
http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1218/

Fixes #900


swaroop-sridhar:
@AndyAyersMS please review.

swaroop-sridhar:

include/GcInfo/GcInfo.h
```diff
@@ -29,46 +29,103 @@
...
+private:
+  // Record a Stack Allocation in the FuncInfo, with appropriate
+  // Flags based on Type of allocation.
+  void recordAlloca(const llvm::AllocaInst *Alloca);
+
+  // Mark additional annotations on a recorded GC Value
+  void markGcAlloca(const llvm::AllocaInst *Alloca, const AllocaFlags Flags);
+
+  // Mark additional annotations on a recorded non-GC-Value
+  void markIntAlloca(const llvm::AllocaInst *Alloca, const AllocaFlags Flags);
```
AndyAyersMS: Why isn't this `recordNonGcAlloca` or similar?

Also, using `const` for by-value parameters looks a bit odd.

include/GcInfo/GcInfo.h
```diff
@@ -124,20 +185,21 @@ class GcInfoEmitter {
...
-  GcSlotId getSlot(int32_t Offset, GcSlotFlags Flags);
-  GcSlotId getTrackedSlot(int32_t Offset);
-  GcSlotId getAggregateSlot(int32_t Offset);
-  GcSlotId getPinnedSlot(int32_t Offset);
+  bool hasSlot(int32_t Offset) { return SlotMap.find(Offset) != SlotMap.end(); }
+  bool isTrackedSlot(const GcSlotId SlotID);
+  GcSlotId getSlot(const int32_t Offset, const GcSlotFlags Flags);
+  GcSlotId getTrackedSlot(const int32_t Offset);
+  GcSlotId getUntrackedSlot(const int32_t Offset, bool IsPinned = false,
+                            bool IsObjectRef = false);
```
AndyAyersMS: Ditto in the methods declared here -- for by-value parameters `const` just adds semantic noise....

lib/GcInfo/GcInfo.cpp
```diff
@@ -57,6 +58,15 @@ bool GcInfo::isGcAggregate(const Type *AggType) {
...
 
+bool GcInfo::isGcAllocation(const llvm::Value *Value) {
+  const AllocaInst *Alloca = dyn_cast<const AllocaInst>(Value);
+  if ((Alloca != nullptr) && GcInfo::isGcType(Alloca->getAllocatedType())) {
+    return true;
+  }
+
+  return false;
+}
+
```
AndyAyersMS: Seems like the latter bit of code could simply be

`return (Alloca != nullptr) && GcInfo::isGCType(...);`

include/GcInfo/GcInfo.h
```diff
@@ -29,46 +29,103 @@
...
+
+enum AllocaFlags {
+  Alloca_None = 0x0,
+  Alloca_GcPointer = 0x1,
+  Alloca_GcAggregate = 0x2,
+  Alloca_Pinned = 0x4,
+  Alloca_GsCookie = 0x8,
+  Alloca_SecurityObject = 0x10,
+  Alloca_GenericsContext = 0x20
+};
```
AndyAyersMS: Since there's already `Alloca` in the enum name, do we need to repeat it in each enumerate?

I think `AllocaFlags::Pinned` reads a bit better.

include/GcInfo/GcInfo.h
```diff
@@ -29,46 +29,103 @@
...
+
+struct AllocaInfo {
+  int32_t Offset;
+  AllocaFlags Flags; // Combination of AllocaFlags, to be precise.
+  const char *getAllocTypeString() const;
+  bool isGcPointer() const { return ((Flags & Alloca_GcPointer) != 0); }
+  bool isGcAggregate() const { return ((Flags & Alloca_GcAggregate) != 0); }
+  bool isGcValue() const {
+    return ((Flags & (Alloca_GcPointer | Alloca_GcAggregate)) != 0);
+  }
```
AndyAyersMS: Since you have a flags enumeration you could add `GcValue = GcPointer | GcAggregate` there.

include/GcInfo/GcInfo.h
```diff
@@ -29,46 +29,103 @@
...
 
-  void recordPinnedSlot(llvm::AllocaInst *Alloca);
-  void recordGcAggregate(llvm::AllocaInst *Alloca);
+  void recordGcAlloca(const llvm::AllocaInst *Alloca);
+  void recordPinned(const llvm::AllocaInst *Alloca);
+  void recordSecurityObject(const llvm::AllocaInst *Alloca);
+  void recordGsCookie(const llvm::AllocaInst *Alloca,
+                      const uint32_t ValidRangeStart,
+                      const uint32_t ValidRangeEnd);
+  void recordGenericsContext(const llvm::AllocaInst *Alloca,
```
AndyAyersMS: `const` on by-value parameters

lib/GcInfo/GcInfo.cpp
```diff
@@ -145,47 +155,130 @@ GcFuncInfo *GcInfo::getGcInfo(const llvm::Function *F) {
...
+      return "GenericsContextThis";
+    } else {
+      return "GenericsContextTypeArg";
+    }
   }
 
-  for (auto Pin : PinnedSlots) {
-    EscapingLocs.push_back((Value *)Pin.first);
+  if (Flags & AllocaFlags::Alloca_SecurityObject) {
+    return "SecurityObject ";
```
AndyAyersMS: Extra trailing space in `"SecurityObject "`

lib/GcInfo/GcInfo.cpp
```diff
@@ -545,85 +625,71 @@ void GcInfoEmitter::encodeTrackedPointers(const GcFuncInfo *GcFuncInfo) {
...
-      dbgs() << "  GSCookieStackSlot @" << GcFuncInfo->GsCookieOffset << " ["
-             << GcFuncInfo->GsCkValidRangeStart << " - "
-             << GcFuncInfo->GsCkValidRangeEnd << "]  Not Reported\n";
+    if (AllocaInfo.Flags & AllocaFlags::Alloca_GenericsContext) {
+      // TODO: GC: Report GenericsInstContextStackSlot
+      // https://github.com/dotnet/llilc/issues/766
+
+      // Encoder.SetGenericsInstContextStackSlot(
+      //    GcFuncInfo->GenericsContextOffset,
+      //    GcFuncInfo->GenericsContextParamType);
```
AndyAyersMS: This one you can implement now, right?

lib/GcInfo/GcInfo.cpp
```diff
@@ -545,85 +625,71 @@ void GcInfoEmitter::encodeTrackedPointers(const GcFuncInfo *GcFuncInfo) {
...
+void GcInfoEmitter::encodeGcAggregate(const AllocaInst *Alloca,
+                                      const AllocaInfo &AllocaInfo) {
+  int32_t AggregateOffset = AllocaInfo.Offset;
+  Type *Type = Alloca->getAllocatedType();
+  StructType *StructTy = cast<StructType>(Type);
+  SmallVector<uint32_t, 4> GcPtrOffsets;
+  const DataLayout &DataLayout = JitContext->CurrentModule->getDataLayout();
+  GcInfo::getGcPointers(StructTy, DataLayout, GcPtrOffsets);
+  bool IsPinned = AllocaInfo.isPinned();
+  bool IsObjectReferece = true; // Only ObjectReferences can be stored in
```
AndyAyersMS: Typo `Referece`

AndyAyersMS:
Overall this looks good, a few typos and style issues to address.

include/GcInfo/GcInfo.h
swaroop-sridhar: Do you mean rename combine the recordAlloca() and markIntAlloca() functionality to recordNonGcAlloca() ? I can do that.

swaroop-sridhar: I added const on the parameters to ensure const-ness in the implementataion (for example, flags aren't modified by the method). Sure, it doesn't have meaning in the interface. But I don't think C++ allows us to hide the const-ness only in the interface.

include/GcInfo/GcInfo.h
swaroop-sridhar: OK will change it, thanks.

lib/GcInfo/GcInfo.cpp
swaroop-sridhar: I can't set this until we set GsCookie valid range correctly. Otherwise, GcInfoEncoder::Build() fails this assertion:
https://github.com/dotnet/coreclr/blob/master/src/gcinfo/gcinfoencoder.cpp#L700

include/GcInfo/GcInfo.h
AndyAyersMS: I just meant it seems like there are two cases, recording a `gc` and recording something that is not `gc` (we don't really care what it is, just that it does not need gc reporting), so the naming might reflect that more directly.

For the `const` case: if you want to ensure that the parameter value is not modified by the method, you can put the `const` only on the definition and leave it off the declaration. Yes, it evidently works. It might seem odd but more directly captures your intent.

lib/GcInfo/GcInfo.cpp
AndyAyersMS: The diff may be off here -- by "this one" I meant the generic context reporting.

swaroop-sridhar: Sorry, I wasn't clear. The GcInfo encoder has an assert check for validity of GSCookieValidRangeStart whenever we set the GenericsContextParamType:

```
    if (hasSecurityObject || hasContextParamType)
    {
        ...
        _ASSERTE(m_GSCookieValidRangeStart > 0 && m_GSCookieValidRangeStart < m_CodeLength);
        ...
    }
```

So, I cannot set the generics context slot yet.

swaroop-sridhar:
@AndyAyersMS : Addressed all the above comments. Thanks.

AndyAyersMS:
LGTM.

PR #904, opened 2015-10-27
Update exclusion file to handle new paths for tests

### LINKED ISSUE #903: Test exclusion mechanism is broken
dotnet/coreclr#1857 broke the exclusion mechanism LLILC uses to skip coreclr tests it currently can't handle. We need to fix this by updating the paths mentioned in the exclusion file.

We should also think about how this generalizes to testing cross-platform.

AndyAyersMS:
Closes #903.

This is a bit of a hack, but it should get us back up off the ground.
Once EH lands we won't need most of these anyways.


pgavlin:
LGTM.

AndyAyersMS:
@pgavlin was about to ask you to PTAL, see you already did.... thanks.

PR #905, opened 2015-10-27
Fix for build break in Linux

kyulee1:
This fixes Linux build when using gcc. Some of name space qualifier are explicitly added.


kyulee1:
@AndyAyersMS PTAL

russellhadley:
LGTM

pgavlin:
Hm, I did not notice this PR before filing #907. #907 has some additional fixes that are necessary when building with clang; I'll wait for this PR to go through and rebase.

PR #906, opened 2015-10-27
Object Writer

kyulee1:
This is a standalone object writer (dll/so) that can wrap code/relocation using LLVM MC layer.
So this tool can be plugged into other either native or managed tool easily.
The following APIs are provided.
InitObjWriter
FinishObjWriter
SwitchSection
EmitAlignment
EmitBlob
EmitIntValue
EmitSymbolDef
EmitSymbolRef

Since the tool itself is very simple, adding an explicit test is skipped.
But we still validate the build of this tool.
Packaging this tool work will follow.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -1,3 +1,5 @@
 add_subdirectory(Reader)
 add_subdirectory(Jit)
 add_subdirectory(GcInfo)
+add_subdirectory(ObjWriter)
+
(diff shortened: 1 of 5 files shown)


kyulee1:
@AndyAyersMS It's not different than what you've seen before. Only small changes in Cmake/descriptions are made. PTAL.

AndyAyersMS:
LGTM.

kyulee1:

PR #907, opened 2015-10-27
Fix some Linux build breaks.

pgavlin:
- _DEBUG was not being defined when CMake was invoked without an explicit
  build type. The build now defaults to debug unless otherwise specified.
- A couple of GC-related types needed global namespace qualifiers to avoid
  name collisions.
- There were various const-cast related issues that needed resolution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,12 +260,12 @@ if ( NOT LLVM_TARGET_IS_CROSSCOMPILE_HOST )
     set(LLILC_HAS_VERSION_PATCHLEVEL 0)
   endif()
 
-  if (CMAKE_BUILD_TYPE STREQUAL "DEBUG")
-    set(DEBUG)
+  if ((CMAKE_BUILD_TYPE STREQUAL "DEBUG") OR (CMAKE_BUILD_TYPE STREQUAL ""))
+    set(DEBUG 1)
   endif()
 
   if (DEBUG)
-    add_definitions( -DDEBUG -D_DEBUG )
+    add_definitions( -D_DEBUG )
   endif()
 
   # Add appropriate flags for MSVC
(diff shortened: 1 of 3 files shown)


pgavlin:
@AndyAyersMS @JosephTremoulet PTAL

JosephTremoulet:
How bad would it be to make the code const-correct instead of inserting const_cast?  E.g. could those be vectors of `const Value*`s?

pgavlin:
@JosephTremoulet The vector of values is eventually used an argument to `IRBuilder::CreateCall`, which expects an `ArrayRef<Value*>`, so there's going to be a const cast somewhere. We can either:
a) change the types of `PinnedSlots` and `GcAggregates` to remove the `const` modifier in the first place
b) cast inside of `GcFuncInfo::getEscapingLocations` as in this change, or
c) cast at the point that the vector is passed to `ArrayRef<Value*>`.

I don't have a terribly strong opinion, but I am leaning towards either (a) or (c). This is a contains-style problem, really; (a) might in fact be the best choice.

JosephTremoulet:
Why are we creating an IR call in a MachineFunctionPass?  Given that we are, I don't have a strong opinion either, whatever's expedient.

pgavlin:
We're not: the call is created during `GenIR::readerPostPass`.

JosephTremoulet:
Ok, I see now.  We populate the map with keys but a sentinel value in the reader where we have non-const `Value *`s, then update the values in the map in the MachineFunctionPass where we have `const Value *`s.  Updating the values using const keys makes the map key type `const Value *`, but `getEscapingLocations` needs to make a vector and the handy thing to iterate is the keys of the map.

So then:
 1) That map iteration in `getEscapingLocations` is non-deterministic, we probably want to accumulate the escaping locations in a separate vector anyway just for determinism (i.e. have the reader build up a vector of non-const `Value *`s in parallel to calling the `record` methods on the GCInfo object).
 2) I'm uneasy about "this alloca is a key in my map" being a property we need transformations to preserve for correctness.  I'd rather rely on the `localescape` call being preserved, and use the alloca's indices in the argument list as the cross-pass persistent thing rather than pointers to the allocas.  So e.g. record the index of the GsCookie etc., and the number of Pinned things and then in `runOnMachineFunction` build up the map by walking the `localescape` call.
 3) No need to tie getting the linux build working to that.  Fine with me to check in what you have and file an issue for the rest.

PR #912, opened 2015-10-29
Insert gc.sagepoint_poll() after the real Method

swaroop-sridhar:
The EE DebugInfoManager requires the Jitted method to start at
the beginning of allocated Code block.
Therefore, this change moves gc.safepoint_poll() from the
beginning to the end of the module.


swaroop-sridhar:
@AndyAyersMS please review

AndyAyersMS:
LGTM

Issue #914, opened 2015-10-31
GC: Fix CodeGenPrepare GcOpts


swaroop-sridhar:
There's an optimization in CodegenPrepare that turns gc.relocate for derived pointers to GEP from the relocated base-pointer.

```
// Turns this:

%base = ...
%ptr = gep %base + 15
%tok = statepoint (%fun, i32 0, i32 0, i32 0, %base, %ptr)
%base' = relocate(%tok, i32 4, i32 4)
%ptr' = relocate(%tok, i32 4, i32 5)
%val = load %ptr'

// into this:

%base = ...
%ptr = gep %base + 15
%tok = statepoint (%fun, i32 0, i32 0, i32 0, %base, %ptr)
%base' = gc.relocate(%tok, i32 4, i32 4)
%ptr' = gep %base' + 15
%val = load %ptr'
```

However, this optimization seems to assume that the base-pointer's gc.relocate always appears before the derived pointer's gc.relocate. However, RewriteStatepointsForGC phase does not enforce this.

So, we can end up with incorrect transformation such as the follows:

After Reader

```
RuntimeHandleHelperCall:                          ; preds = %49
  %97 = call i64 bitcast (i64* @"AnyJITHelper::JitHelper.2" to i64 (i64, i64)*)(i64 %50, i64 ptrtoint (i64* @"Signature(TK_11000000)" to i64))
  br label %59
```

After PlaceSafepoints

```
RuntimeHandleHelperCall:                          ; preds = %49
  %safepoint_token.44 = call i32 (i64, i32, i64 (i64, i64)*, i32, i32, ...) @llvm.experimental.gc.statepoint.p0f_i64i64i64f(i64 2882400000, i32 0, i64 (i64, i64)* bitcast (i64* @"AnyJITHelper::JitHelper.2" to i64 (i64, i64)*), i32 2, i32 0, i64 %50
, i64 ptrtoint (i64* @"Signature(TK_11000000)" to i64), i32 0, i32 0)
  %97 = call i64 @llvm.experimental.gc.result.i64(i32 %safepoint_token.44)
  br label %59
```

After RewriteStatepoints

```
RuntimeHandleHelperCall:                          ; preds = %49
[...]

PR #915, opened 2015-10-31
Disable CodeGen prepare GcOpts

### LINKED ISSUE #911: GC: Assertion failure in Roslyn with statepoints inserted
Roslyn run stage 1 with statepoints inserted causes an assertion failure in the SelectionDAG
Assertion failed: !N.getNode() && "Already set a value for this node!"

swaroop-sridhar:
CodegenPrepare's gc-opt turns gc.relocate for derived pointers
into GEP from the relocated base-pointer.

However, this optimization seems to assume that the base-pointer's
gc.relocate always appears before the derived pointer's gc.relocate.
However, RewriteStatepointsForGC phase does not enforce this, which
leads to an incorrect transformation.

So, disable the optimization until this issue is fixed.
Further details: Issue #914

Fixes #911

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -162,10 +162,20 @@ ICorJitCompiler *__stdcall getJit() {
     // Allow LLVM to pick up options via the environment
     cl::ParseEnvironmentOptions("LLILCJit", "COMPlus_AltJitOptions");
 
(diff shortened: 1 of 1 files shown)


swaroop-sridhar:
@AndyAyersMS  please review

AndyAyersMS:
LGTM

PR #920, opened 2015-11-04
Fix SuppressExceptionHandlers processing

JosephTremoulet:
The logic to actually suppress handlers got dropped somewhere; re-enable
it by using calls instead of invokes everywhere when the flag is set.

Move the special logic for leaves inside of handlers under this flag (it
is correct only for the case that handlers are being suppressed).


JosephTremoulet:
@pgavlin PTAL.  You'll probably want to set SuppressExceptionHandlers to true on non-Windows for now.

JosephTremoulet:

russellhadley:
LGTM

PR #921, opened 2015-11-04
Exclude a couple of tests that use vararg methods.

richardlford:
This change also modifies the form of exclusions in the exclusion.targets
file.  Because of a recent change to the testing infrastructure each
test project will have its binary and cmd in a distinct directory even
if its source was in the same directory as other test projects. Thus
to exclude a test it is sufficient to give the path to the directory
contaning its cmd file, followed by "\*". The root name of the cmd
file will be the same as the subdirectory name so there is no need to
repeat it.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -1,2373 +1,2379 @@
 <Project DefaultTargets = "GetListOfTestCmds"
     xmlns="http://schemas.microsoft.com/developer/msbuild/2003" >
     <ItemGroup Condition="'$(XunitTestBinBase)' != ''">
-        <ExcludeList Include="$(XunitTestBinBase)\Interop\ICastable\**\Castable.cmd" >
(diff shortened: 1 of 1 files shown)


richardlford:
@AndyAyersMS, PTAL.

AndyAyersMS:
Since you edited the entire file I can't see diffs easily.

Was there some reason you needed to modify everything? What we were doing before seemed to work ok.

richardlford:
The added exclusions (two of them) are at the very end. 

Before we were using a ** wildcard which is wild over directories. It would eventually find the cmd file to exclude but might have to search several directories. The changed version specifies the correct directory so may be more efficient. And I think it is clearer than having the **.

richardlford:

AndyAyersMS:
Ok, thanks.

LGTM

richardlford:
Failure was for win32 leg that does not run tests, so proceeding with merge.

PR #924, opened 2015-11-05
Make b28158 test exclusion unconditional

JosephTremoulet:
This can time out with non-statepoints runs too; for example, it held up
the last FI from LLVM.


JosephTremoulet:
@russellhadley or @AndyAyersMS PTAL.

AndyAyersMS:
Nice test case you have there...

LGTM

PR #925, opened 2015-11-06
Updates for LLVM FI signature changes

JosephTremoulet:
Three changes in the next FI require changes to LLILC sources, which are bundled up here.

This will need to be committed in tandem with Microsoft/llvm#113.

@AndyAyersMS PTAL.

### DIFF
--- a/include/Jit/EEObjectLinkingLayer.h
+++ b/include/Jit/EEObjectLinkingLayer.h
@@ -34,9 +34,10 @@ private:
   class ConcreteLinkedObjectSet : public LinkedObjectSet {
   public:
     ConcreteLinkedObjectSet(MemoryManagerPtrT MemMgr,
-                            SymbolResolverPtrT Resolver)
-        : LinkedObjectSet(*MemMgr, *Resolver), MemMgr(std::move(MemMgr)),
-          Resolver(std::move(Resolver)) {}
+                            SymbolResolverPtrT Resolver,
+                            bool ProcessAllSections)
+        : LinkedObjectSet(*MemMgr, *Resolver, ProcessAllSections),
+          MemMgr(std::move(MemMgr)), Resolver(std::move(Resolver)) {}
 
     void finalize() override {
       State = Finalizing;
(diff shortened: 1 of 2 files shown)


include/Jit/EEObjectLinkingLayer.h
```diff
@@ -73,7 +75,18 @@ class EEObjectLinkingLayer : public ObjectLinkingLayerBase {
...
+        NotifyFinalized(std::move(NotifyFinalized)), ProcessAllSections(false) {
+  }
+
+  /// @brief Set the 'ProcessAllSections' flag.
+  ///
+  /// If set to true, all sections in each object file will be allocated using
+  /// the memory manager, rather than just the sections required for execution.
+  ///
+  /// This is kludgy, and may be removed in the future.
+  void setProcessAllSections(bool ProcessAllSections) {
```
AndyAyersMS: Maybe add `override` here? I was puzzled at first how this ever would get called.

AndyAyersMS:
LGTM, just one nit.

include/Jit/EEObjectLinkingLayer.h
JosephTremoulet: It's not an override, it will never be called.  Our `EEObjectLinkingLayer` is a copy of the stock Orc `ObjectLinkingLayer` but with the relocation processing stripped out.  The stock Orc `ObjectLinkingLayer` picked up a `setProcessAllSections` method so that it can be called by `OrcMCJitReplacement` which has a method of the same name that is an override.  It would be correct to remove this method and the new field from our copy and propagate all uses of it away to the 'false' value they always have, but I was leery to do so since it would introduce unnecessary diffs vs the stock copy which would complicate future merges of this sort (and also because I'm not entirely sure that we won't find a need for something similar when we point to whatever target the stock Orc is using it for).

include/Jit/EEObjectLinkingLayer.h
```diff
@@ -34,15 +34,15 @@ class EEObjectLinkingLayer : public ObjectLinkingLayerBase {
   class ConcreteLinkedObjectSet : public LinkedObjectSet {
   public:
     ConcreteLinkedObjectSet(MemoryManagerPtrT MemMgr,
-                            SymbolResolverPtrT Resolver)
-        : LinkedObjectSet(*MemMgr, *Resolver), MemMgr(std::move(MemMgr)),
-          Resolver(std::move(Resolver)) {}
+                            SymbolResolverPtrT Resolver,
+                            bool ProcessAllSections)
+        : LinkedObjectSet(*MemMgr, *Resolver, ProcessAllSections),
```
JosephTremoulet: This line is the critical one for the `ProcessAllSections` stuff -- the base `LinkedObjectSet` class is not a llilc-specific copy and its constructor now takes a `ProcessAllSections` parameter.  I believe just passing `false` here would have been correct, but copied the full change for the reasons I mentioned in my previous comment.

AndyAyersMS:
Ah, ok. Seems like we should avoid having copies like this and work on having the built-in stuff to be flexible enough to meet our needs.

Changes LGTM.

JosephTremoulet:
> we should avoid having copies like this and work on having the built-in stuff to be flexible enough to meet our needs

No argument here.

PR #926, opened 2015-11-07
Conditionalize 691 exclusions on ExecuteHandlers

JosephTremoulet:
A number of the EH-related tests can now pass (with conservative GC).  The
ExecuteHandlers configuration is still off-by-default since it's not yet
compatible with precise GC (and there are known errors with nested funclet
frame setup, and unknown errors in the 58 exclusions left unconditional),
but this change makes the exclusions for EH tests that work with precise
GC conditional on the ExecuteHandlers config flag so that we can start
tracking the progress of that configuration in the lab.


JosephTremoulet:
Whoops, didn't mean to commit this w/o review.  @AndyAyersMS / @russellhadley, PTAL.

AndyAyersMS:
LGTM

PR #929, opened 2015-11-07
Remove implicit list iterator conversions

JosephTremoulet:
LLVM change r252372 (git commit 84cb8974b) removed the implicit
conversions between ilist/iplist iterators and pointers to the iterated
nodes (to force people to consider whether they might be trying to
illegally cast an end iterator to a node).  Update LLILC sources to match.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -257,7 +257,7 @@ static Argument *functionArgAt(Function *F, uint32_t I) {
   for (; I > 0; I--) {
     ++Args;
   }
-  return Args;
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
This will be needed to get the FI at Microsoft/llvm#114 to succeed.

JosephTremoulet:
@AndyAyersMS PTAL

AndyAyersMS:
LGTM.

JosephTremoulet:
Failure is just the known unrelated issue with the new StructABI test; merging

PR #930, opened 2015-11-08
Make two test exclusions unconditional

JosephTremoulet:
JIT/Methodical/cast/SEH/_speed_dbgthrow and _speed_relthrow still
sometimes fail under ExecuteHandlers; exclude them unconditionally.


JosephTremoulet:
@AndyAyersMS PTAL.  I haven't yet dug into why these two tests passed before and don't know; could be nondeteministic failure, could be interaction with other changes (such as the proper RA fix that's in LLVM now vs the workaround I was testing on top of previously) that shifted us from getting lucky to not.  Regardless, they still belong on the unconditional exclusion list.

AndyAyersMS:
LGTM

PR #931, opened 2015-11-10
Exclude gettypetypeofmatrix test if eh is not enabled.

richardlford:
In preparation for the addition of a new test,
JIT\Directed\gettypetypeof\gettypetypeofmatrix\gettypetypeofmatrix.exe,
add an exclusion for it. The test fails if LLILC's
COMPlus_ExecuteHandlers is not set, but passes if it is, so
add the exclusion to the part where it is not set.


richardlford:
@pgavlin, PTAL.

pgavlin:
LGTM

JosephTremoulet:
> The test fails if LLILC's COMPlus_ExecuteHandlers is not set, but passes if it is, so add the exclusion to the part where it is not set.

Thanks for checking!

PR #933, opened 2015-11-10
Emit frame info in Object writer

kyulee1:
This supports for emitting frame info for Windows.
Basically pdata/xdata is constructed like C++ to allow call stack dumps.
EH table itself is not populated yet since EH model is not determined.


kyulee1:
@russellhadley @AndyAyersMS PTAL

AndyAyersMS:
LGTM.

PR #934, opened 2015-11-11
Remove the eager tail recursion detection code

AndyAyersMS:
Also remove as much of the related support code as I could find.
Also fix an unused local warning.


AndyAyersMS:
@JosephTremoulet PTAL

JosephTremoulet:
LGTM

PR #935, opened 2015-11-12
Add support for the SysV AMD64 ABI.

pgavlin:
Most of the work here involves adding support for aggregate arguments
that are represented as multiple primitive and aggregate arguments
that are represented as LLVM 'byvalue' arguments (i.e. LLVM arguments
with the 'byvalue' attribute). The former are used for IL by-value
struct arguments that must be passed in multiple registers; the latter
are used for by-value struct arguments that must be passed on the
stack. All arguments are classified as do enregister/do not
enregister using the interface provided by CoreCLR.

### DIFF
--- a/lib/Reader/abi.cpp
+++ b/lib/Reader/abi.cpp
@@ -32,8 +32,12 @@ using namespace llvm;
 class X86_64_Win64 {
 private:
   X86_64_Win64() {}
+
   static ABIArgInfo classify(const ABIType Ty, const DataLayout &DL,
                              bool IsManagedCallingConv);
+
+  static ABIArgInfo classifyResult(const ABIType Ty, const DataLayout &DL,
+                                   bool IsManagedCallingConv);
 
 public:
   static void computeSignatureInfo(bool IsManagedCallingConv,
@@ -48,8 +52,24 @@ class X86_64_SysV {
 private:
   X86_64_SysV() {}
 
+  static ABIArgInfo classify(LLILCJitContext &JitContext, const ABIType Ty,
+                             const DataLayout &DL, bool IsManagedCallingConv,
+                             bool IsResult, uint32_t &RequiredIntRegs,
+                             uint32_t &RequiredSSERegs);
+
+  static ABIArgInfo classifyArgument(LLILCJitContext &JitContext,
+                                     const ABIType Ty, const DataLayout &DL,
+                                     bool IsManagedCallingConv,
+                                     uint32_t &RequiredIntRegs,
+                                     uint32_t &RequiredSSERegs);
+
+  static ABIArgInfo classifyResult(LLILCJitContext &JitContext,
+                                   const ABIType Ty, const DataLayout &DL,
+                                   bool IsManagedCallingConv);
+
 public:
-  static void computeSignatureInfo(bool IsManagedCallingConv,
+  static void computeSignatureInfo(LLILCJitContext &JitContext,
+                                   bool IsManagedCallingConv,
                                    ABIType ResultType,
                                    ArrayRef<ABIType> ArgTypes,
                                    const DataLayout &DL, ABIArgInfo &ResultInfo,
@@ -64,15 +84,16 @@ private:
 public:
   X86_64ABIInfo(Triple TargetTriple, const DataLayout &DL);
 
-  void computeSignatureInfo(CallingConv::ID CC, bool IsManagedCallingConv,
-                            ABIType ResultType, ArrayRef<ABIType> ArgTypes,
-                            ABIArgInfo &ResultInfo,
+  void computeSignatureInfo(LLILCJitContext &JitContext, CallingConv::ID CC,
+                            bool IsManagedCallingConv, ABIType ResultType,
+                            ArrayRef<ABIType> ArgTypes, ABIArgInfo &ResultInfo,
                             std::vector<ABIArgInfo> &ArgInfos) const override;
 };
 
 ABIArgInfo X86_64_Win64::classify(const ABIType ABITy, const DataLayout &DL,
                                   bool IsManagedCallingConv) {
   Type *Ty = ABITy.getType();
+  assert(!Ty->isVoidTy());
 
   if (Ty->isAggregateType() || Ty->isVectorTy()) {
     // If the aggregate's size in bytes is a power of 2 that is less than or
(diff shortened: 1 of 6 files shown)


pgavlin:
@JosephTremoulet @russellhadley PTAL

lib/Reader/abi.cpp
```diff
@@ -125,17 +129,136 @@ void X86_64_Win64::computeSignatureInfo(bool IsManagedCallingConv,
...
+  Type *Ty = ABITy.getType();
+
+  if (Ty->isAggregateType() || Ty->isVectorTy()) {
+    SYSTEMV_AMD64_CORINFO_STRUCT_REG_PASSING_DESCRIPTOR Desc;
+
+    bool Ok = Context.JitInfo->getSystemVAmd64PassStructInRegisterDescriptor(
+        ABITy.getClass(), &Desc);
+    assert(Ok);
+
+    if (!Desc.passedInRegisters) {
```
JosephTremoulet: This could use some comments.  I take it the IsResult path is indicating that the caller will allocate a hidden buffer and pass a pointer in one of the argument registers (and that RequiredFooRegisters is tracking the passing of the arguments, not the returns), and the !IsResult path is indicating that the argument will be passed by value on the stack?

lib/Reader/abi.cpp
```diff
@@ -125,17 +129,136 @@ void X86_64_Win64::computeSignatureInfo(bool IsManagedCallingConv,
...
+
+  if (Ty->isFloatingPointTy()) {
+    assert(Ty->isFloatTy() || Ty->isDoubleTy());
+    RequiredIntRegs = 0;
+    RequiredSSERegs = 1;
+    return ABIArgInfo::getDirect(Ty);
+  }
+
+  assert(Ty->isPointerTy() || Ty->isVoidTy());
+  RequiredIntRegs = 1;
```
JosephTremoulet: Why does VoidTy require an int reg?

lib/Reader/abi.cpp
```diff
@@ -125,17 +129,136 @@ void X86_64_Win64::computeSignatureInfo(bool IsManagedCallingConv,
...
+    ABIArgInfo Info = classify(Context, Arg, DL, IsManagedCallingConv, false,
+                               RequiredIntRegs, RequiredSSERegs);
+
+    // As per the ABI, if any eightbyte of a struct value does not fit in the
+    // number of available registers, the entire struct value is passed on the
+    // stack.
+    if (Info.getKind() == ABIArgInfo::Expand &&
+        (FreeIntRegs < RequiredIntRegs || FreeSSERegs < RequiredSSERegs)) {
+      RequiredIntRegs = 1;
+      Info = ABIArgInfo::getDirect(Arg.getType());
```
JosephTremoulet: What's the int reg required for?

lib/Reader/abisignature.cpp
```diff
@@ -553,23 +708,41 @@ Function *ABIMethodSignature::createFunction(GenIR &Reader, Module &M) {
...
+
+    case ABIArgInfo::SignExtend:
+      ArgAttrs.addAttribute(Attribute::SExt);
+      goto direct;
+
+    case ABIArgInfo::Direct: {
+    direct:
+      Type *ArgTy = Arg.getType();
+      if (ArgTy->isStructTy()) {
+        ArgTy = Reader.getManagedPointerType(ArgTy);
```
JosephTremoulet: why managed?  I'd have expected that
 1. this will always be something allocated in the caller's frame, so outside the GC heap, and
 2. we won't be inserting write-barriers for stores into it

if that's right, then unmanaged would seem to make more sense.

lib/Reader/readerir.cpp
```diff
@@ -366,17 +379,47 @@ void GenIR::readerPrePass(uint8_t *Buffer, uint32_t NumBytes) {
...
+      Type *BytePtrType = Type::getInt8PtrTy(LLVMContext, 0);
+      Value *Home = LLVMBuilder->CreatePointerCast(Arguments[J], BytePtrType);
+
+      ArrayRef<ABIArgInfo::Expansion> Expansions = ArgInfo.getExpansions();
+      uint32_t ExpansionCount = Expansions.size();
+      for (uint32_t E = 0; E < ExpansionCount - 1; ++E, ++I, ++ArgI) {
+        ABISignature::collapse(*this, Expansions[E], &*ArgI, Home);
+      }
+      ABISignature::collapse(*this, Expansions[ExpansionCount - 1], &*ArgI,
+                             Home);
```
JosephTremoulet: why can't this last one be folded up into the loop?

JosephTremoulet:
Looks good, as near as I can tell, aside from some minor questions.

lib/Reader/abi.cpp
pgavlin: Yes. I've added comments to that effect.

lib/Reader/abi.cpp
pgavlin: It doesn't; good catch.

lib/Reader/abi.cpp
pgavlin: It isn't: indeed, this should be setting `RequiredIntRegs` and `RequiredSSERegs` to `0`.

pgavlin: (or skipping the free register count adjustment entirely)

lib/Reader/abisignature.cpp
pgavlin: Yes, you're correct. Good point.

lib/Reader/readerir.cpp
pgavlin: The containing loop (i.e. the loop over the argument infos) will also increment `I` and `ArgI`. This last `collapse` is outside the inner loop to account for that increment.

AndyAyersMS:
I'm probably confused, but haven't we established that all the by-val to by-ref conversions in the ABI require modelling the resulting argument as a managed pointer, since some cases (reflection?) might call these methods in unusual ways? 

Or is this only the case for return values, governed by `isStructRequiringStackAllocRetBuf`?

pgavlin:
> I'm probably confused, but haven't we established that all the by-val to by-ref conversions in the ABI require modelling the resulting argument as a managed pointer, since some cases (reflection?) might call these methods in unusual ways?

Normally, yes, but the conversion in this case is not really a by-val to by-ref conversion. In the case of the AM64 ABI, tagging the pointer-typed argument with the `byval` attribute instructs the backend to place the provided argument in the outgoing args area of the stack. In other words, the parameter is not actually by-ref as far as the ABI is concerned, LLVM merely models it as such.

lib/Reader/abi.cpp
```diff
@@ -107,35 +128,193 @@ ABIArgInfo X86_64_Win64::classify(const ABIType ABITy, const DataLayout &DL,
...
+                                 uint32_t &RequiredSSERegs) {
+  Type *Ty = ABITy.getType();
+  assert(!Ty->isVoidTy());
+
+  if (Ty->isAggregateType() || Ty->isVectorTy()) {
+    SYSTEMV_AMD64_CORINFO_STRUCT_REG_PASSING_DESCRIPTOR Desc;
+
+    bool Ok = Context.JitInfo->getSystemVAmd64PassStructInRegisterDescriptor(
+        ABITy.getClass(), &Desc);
+    assert(Ok);
```
JosephTremoulet: Looks like this would give a warning in release builds that Ok is unused, no?

PR #936, opened 2015-11-13
Emit DebugLocInfo/DebugFileInfo in ObjectWriter for Windows

kyulee1:
This basically emits two tables.
One is for file name tables (along with index table).
The other is for native offset to source location info.
Adapted CodeView emission logic from LLVM mainline. Ideally this code should be refactored to be shared, which I will follow-up.

Nuget package version (and name for linux) is updated.

### DIFF
--- a/lib/ObjWriter/.nuget/Microsoft.Dotnet.ObjectWriter.nuspec
+++ b/lib/ObjWriter/.nuget/Microsoft.Dotnet.ObjectWriter.nuspec
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>Microsoft.DotNet.ObjectWriter</id>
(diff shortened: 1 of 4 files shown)


kyulee1:
@AndyAyersMS @russellhadley  PTAL

russellhadley:
LGTM

I'm thinking though as this get's more sophisticated we need to be able to add some kind of unit tests.  We should just be able to feed dummy data to the lib and check output.  How hard would this be in xunit?

kyulee1:
Updated the source to reflect the feedback from the consumer side. The current form is more appropriate when we emit loc for dwarf which will be emitted for each native offset.

kyulee1:
@russellhadley Will think about how to harness testing it locally.

PR #937, opened 2015-11-13
Make GcInfo lib depend on intrinsics_gen

JosephTremoulet:
LLVM change r252796 (git hash f805775e) reworked llvm/IR/Attributes.h so
that it now #includes a generated .inc file.  GCInfo.cpp transitively
 #includes Attributes.h, so make it depend on the intrinsics_gen target
(which includes the generation of the .inc file)


JosephTremoulet:
@russellhadley or @pgavlin PTAL.  This issue is preventing FIs into Microsoft/llvm:MS.

pgavlin:
LGTM. Thanks for doing this.

russellhadley:
LGTM

PR #940, opened 2015-11-14
Enable LLILC exception handling when compiling Roslyn

richardlford:
Previously we had to use a workaround in Roslyn when processing it
with LLILC because its default execution path threw an exception and
LLILC could not handle that. However now LLILC can handle some
exceptions.

This change enables the LLILC exception handling when compiling
Roslyn.


richardlford:
I won't commit this until the dotnet_llilc_roslyn_test passes with this change.

@JosephTremoulet, PTAL.

JosephTremoulet:
LGTM :smiley:

PR #942, opened 2015-11-14
Remove dubious usage of getNextNode

JosephTremoulet:
LLVM change r252694 (git commit 6426aea5) fixed a problem with getNextNode
when called on the last item in a list (instr in a block / block in a
functino), so that it now returns nullptr as expected, whereas previously
it cast the end iterator sentinel to a bogus node pointer.

This caused some tests to fail in createTemporary where we had evidently
(perhaps unwittingly) worked around the previous behavior by comparing the
result against the end iterator rather than nullptr.

This change updates createTemporary to deal directly in iterators when
setting the insertion point, rather than translating them back to
instructinos.  I audited all the other calls to getNextNode() in LLILC
sources, and changed several of them to likewise deal in iterators, to
simplify reasoning about the code.  There remaining 6 calls all appear to
be correct in that the input is never null/end and null is the desired
output if the input was the last element.

### DIFF
--- a/lib/Reader/readerir.cpp
+++ b/lib/Reader/readerir.cpp
@@ -580,13 +580,16 @@ void GenIR::cloneFinallyBodies() {
       continue;
     }
 
-    Instruction *InsertBefore;
+    BasicBlock *InsertBlock;
+    BasicBlock::iterator InsertPoint;
     if (auto *Catchpad = dyn_cast<CatchPadInst>(First)) {
-      InsertBefore = &*Catchpad->getNormalDest()->getFirstInsertionPt();
+      InsertBlock = Catchpad->getNormalDest();
+      InsertPoint = InsertBlock->getFirstInsertionPt();
     } else {
-      InsertBefore = First->getNextNode();
+      InsertBlock = &BB;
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
@AndyAyersMS (or anybody) PTAL.  This fixes an issue that blocks automatic FIs from LLVM.

russellhadley:
LGTM

pgavlin:
LGTM as well.

AndyAyersMS:
@JosephTremoulet wonder if we can now undo the change we made in microsoft/llvm#75 ?

JosephTremoulet:
@AndyAyersMS looks like that change (Microsoft/llvm@10aaea39936349fc95cd13d924d1991c26f998d0) updated `splitBasicBlock` to avoid attempting an invalid cast when the block is empty; IIUC, that cast is still invalid, and what changed here (llvm-mirror/llvm@6426aea5) is that `getNextNode` no longer attempts the same invalid cast.  But `splitBasicBlock` isn't going through `getNextNode`.  So I think undoing that is still waiting on to stop using empty blocks (which I thought we'd decided to do, but I can't find an open issue).

PR #943, opened 2015-11-16
Make 13 more test exclusions conditional on EH

JosephTremoulet:
Support for generating/initializing the PSPSym and using it to find the
root frame in funclet prologs was added to LLVM in r252983 (git commit
a736fb542); update the exclusions list to reflect the tests that now pass
when ExecuteHandlers is set.


JosephTremoulet:
@pgavlin / @AndyAyersMS PTAL

pgavlin:
LGTM assuming a green build.

PR #944, opened 2015-11-16
Use llvm::Function for direct call targets

AndyAyersMS:
When we have a direct call target, lookup or create an llvm::Function for the value. Since we don't know the right function type at that time, use void() and patch up the function type later on during ABI processing. Handle direct recursion specially by using the llvm::Function for the current method.

The fixup process is a bit unsavory and uses the "dangerous" mutate method. We could avoid this by running part of the ABI signature unravelling when we first encounter the target method handle; I haven't looked into how feasible this is.

This gets us normal looking function references in direct calls. The global declarations still show the void() types however.


AndyAyersMS:
@pgavlin or @JosephTremoulet PTAL

JosephTremoulet:
Sure, that works

AndyAyersMS:
Looks like there are some test failures... will report back when I've debugged these.

AndyAyersMS:
One issue is handling the following IL. We're compiling `Test::Main`. So we know its type up front. We don't know the other function types, so they're typed as `void()`. We assert when typing the PHI at `MERGE1`. I can add logic to `getStackMergeType` for this I suppose, but I think there are still possibly going to be problems at the `calli`.

```
      ldsfld int32 JitTest.Test::n
      switch (CASE1, CASE2, CASE3)
      ldftn int32 JitTest.Test::Main()
      br.s MERGE1
CASE1:
      ldftn int32 JitTest.Test::Main1()
      br.s MERGE1
CASE2:
      ldftn int32 JitTest.Test::Main2()
      br.s MERGE1
CASE3:
      ldftn int32 JitTest.Test::Main3()
MERGE1:
      tail. calli int32()

**Comment by AndyAyersMS** [comment]
@erozenfeld I'm trying to understand better why we have code in `ABICallSignature::emitCall` where `IsJmp` impacts the imputed target function type -- seems like this could create type mismatches for the changes I'm proposing above, if a method has a mixture of call and jmp sites with the same target.

**Comment by erozenfeld** [comment]
We use musttail LLVM attribute for calls corresponding to jmp. LLVM restriction for this attribute is that the signatures of the caller and the callee must match. MSIL instructions always do match (that's a restriction of jmp opcode) but if the caller has a secret parameter, we need to pass one to the callee (even though it may be ignored) to satisfy that constraint.

**Comment by AndyAyersMS** [comment]
Thanks, I will skip retyping functions when we're jumping to them, then, and just cast the type instead.

**Comment by AndyAyersMS** [comment]
Updated to handle phis of function types and jmps.

**Comment by AndyAyersMS** [comment]
Ok, updated the type check to use `getValueType` instead of `getType`, and squashed...

**Comment by AndyAyersMS** [comment]
One more update where I relent on names matching up.

**Comment by AndyAyersMS** [comment]
Now hitting a few (~8) cases where we have "inconsistent theories" around function types because of the assert I added to detect type thrashing. Here's one example (from localloc large), compiling `System.Threading.Thread.SetStartHelper`...

```
  IL_002e:  ldftn      instance void System.Threading.ThreadHelper::ThreadStart(object)
  IL_0034:  newobj     instance void System.Threading.ParameterizedThreadStart::.ctor(object,                                                                                      native int)
```
with resulting type conflict:
```
void (%System.Threading.ThreadStart addrspace(1)*, %System.Object addrspace(1)*, i64)*
void (%System.Threading.ParameterizedThreadStart addrspace(1)*, %System.Object addrspace(1)*, i64)*
```
We'll likely have to back off from mutating the type in such cases, and just cast.

**Comment by AndyAyersMS** [comment]
Handle variant function types by mutating for the first and casting for the rest. This passes local testing so should be good.

File: lib/Reader/abisignature.cpp [hunk 1]
```diff
@@ -580,7 +581,38 @@ Value *ABICallSignature::emitCall(GenIR &Reader, Value *Target, bool MayThrow,
...
+      // of LLVM's musttail that would cause this signature to diverge
+      // from what we'd have come up with in a normal call to the target.
+      if (!IsJmp) {
+        // We shouldn't change our view of function types multiple times,
+        // so only modify the type if it's the placeholder type we installed
+        // in makeDirectCallTargetNode.
+        Type *VoidType = Type::getVoidTy(Context);
+        PointerType *PlaceholderTy =
+            FunctionType::get(VoidType, false)->getPointerTo();
+        MutateType = TargetFunc->getType() == PlaceholderTy;
```
JosephTremoulet: If we see two different signatures for the same function, and the first one happens to be void(), this will wind up mutating the type out from under IR that's already been generated with the previous type in mind.  Do you think that's worth worrying about, or that it can't/won't happen in practice?

JosephTremoulet:
LGTM, modulo one question.

lib/Reader/abisignature.cpp
AndyAyersMS: I don't think it can happen. 

Basically if we see multiple possible types for a given method they all need to be compatible in some way -- same #s and kinds of arguments (modulo the weird jump case we exclude). Otherwise we're going to generate bad code no matter what we do.

`void()` is only compatible with itself.

AndyAyersMS:
@dotnet-bot test this please

AndyAyersMS:
@mmitche I don't seem to be able to trigger CI tests either via push or manually. Any ideas?

JosephTremoulet:
> The global declarations still show the void() types however

Was curious about this, so took a look.  `Value` has instance field `Type *VTy`, which is what `mutateType` mutates.  `Function` introduces instance field `FunctionType *Ty` (which looks to be technically redundant but strongly typed, presumably to trade a pointer per `Function` and so avoid a lot of downcasts or virtualizing `Value::VTy`), which is not getting mutated, and is what `Function::dump` ultimately uses to determine what type to dump.  It's also what `Function::getFunctionType` returns, so indirectly gets used by `getReturnType` and `isVarArg`, as well as ` BuildLazyArguments`/`arg_size()`/`arg_empty()`.  I don't see a method for mutating the `Ty` field of a `Function`.

AndyAyersMS:
The fact that a function has two type fields seems somewhat troubling.  I'll bring this up on llvm-dev and see if it bothers anyone else. 

I suppose we'll just get chastised for using `mutateType`, but arguably, it should change both pointers.

AndyAyersMS:
Actually there are three cached type fields, one on `Function`, one on `GlobalValue`, and one on `Value`. When a function is created (with type X, say), the first two agree and the third is ptr-to-X. The one on `Value` can be mutated, but not the other two (accessible via `getValueType` and `getFunctionType` respectively).

JosephTremoulet:
Maybe we could avoid both `mutateType` and `WeakVH` by collecting the function typing decisions (say in a `Function`->`Type` map) and waiting to apply them to somewhere like the end of `readerPostVisit`? At that point we think the IR is complete (we're about to run the verifier and `removeUnreachable` on it), so maybe we could expect to be done with the global name map and operand stacks and etc (everything that RAUW would invalidate), and that all references would therefore be in the IR and be properly updated by `replaceAllUsesWith`?

AndyAyersMS:
Perhaps a post-pass could work, but I worry that these decisions are not local, since I had to modify the logic in the PHI type update as well.

JosephTremoulet:
Ah, yeah.  Isn't the type mutation what's making it nonlocal, though?  If you had to change `stackMergeType`, I think that means that on one path to a join we had a function on the stack but hadn't yet seen a direct call to it so didn't have a non-placeholder type, and on the other path had a function pointer whose type we did know... with that pattern, seeing the type at an eventual call isn't going to propagate back up through the phi to the function, so ISTM there's no benefit to retyping that function (as seen on the use in the PHI).
Maybe what we want is to just keep the void() typing during the reader, and insert bitcasts at places like `getDirectCallTarget` where we learn the true type (I believe this could be a `ConstantExpr` and so wouldn't need to insert an additional instruction).  That right there might be good enough to make the inliner happy?  Regardless, we could have a post-pass after the reader that looks through the bitcasts to find the best type, then replace it with a new properly-typed `Function`, inserting/removing bitcasts at uses as appropriate.
What would you think of that?

AndyAyersMS:
I think a post pass is probably the way to g. Just trying to think it through a bit while I'm off doing other things.

PR #945, opened 2015-11-16
Formalize object writer package names.

kyulee1:
Add support for redirection package using json.
Add individual nuspec for each platform.
Change local script to handle json file.

Note this doesn't affect any CIbuild but it is used for a post-process to create packages.


kyulee1:
@russellhadley  PTAL

russellhadley:
LGTM

PR #946, opened 2015-11-18
Exclude tests failing because of tail call.

richardlford:
Exclude tests that fail because LLILC does not yet
support tail call.


richardlford:
@AndyAyersMS, PTAL

AndyAyersMS:
LGTM.

Hopefully we can get this implemented soon....

richardlford:
The failures were unrelated to this change so I'm going to go ahead and merge.

PR #952, opened 2015-11-20
We don't need to chcp anymore

### LINKED ISSUE #12: Basic tests vs CoreCLR require catching an exception
In the current CoreCLR/CoreFX drops, tests that write to the console will require the jit to generate code that can handle exceptions, unless the current code page is 65001 (my cmd shell defaults to CP 437).

In particular we see a throw from a protected block in System.Text.Encoding.GetEncoding(int32 codePage) with a local catch.

Given that MSILC won't be in a position to generate correct code for catches for a while, we need a workaround -- either set the code page when running the tests, avoid tests that write to the console, get CoreFx to look at whether this behavior is reasonable, or blacklist that method in our jit and allow the production jit to handle it.

AndyAyersMS:
The issue in CoreFX that lead to us needing to run tests in code page 65001 is long since fixed. So remove the chcp from our test scripts and docs.

Closes #12.

### DIFF
--- a/Documentation/Debugging.md
+++ b/Documentation/Debugging.md
@@ -203,24 +203,6 @@ the following environment variables are consulted:
   is executing. It is a semi-colon-separated list of
   directories in which to look for the assemblies.
 
-
-### Code Page Bug
-There is (or was) a bug which caused an exception to be
-thrown if the current code page is not 65001. 
-Since LLILC is not currently able to produce code that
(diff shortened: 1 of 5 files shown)


AndyAyersMS:
@russellhadley PTAL

russellhadley:
LGTM

AndyAyersMS:
CI failure is unrelated (see #953). Merging.

PR #954, opened 2015-11-21
Don't sort includes

### LINKED ISSUE #953: formatter now wants to see sorted includes
Something changed recently and the formatter is now failing in the PR tests because of include order.

AndyAyersMS:
Closes #953.

We might want to support this someday, but right now sorting includes per this rule breaks lots of stuff.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,1 +1,2 @@
 BasedOnStyle: LLVM
+SortIncludes:    false


AndyAyersMS:
@JosephTremoulet PTAL

JosephTremoulet:
LGTM

AndyAyersMS:
Formatter passed, which is good enough for me.

PR #956, opened 2015-11-24
Implement support for tail calls

### LINKED ISSUE #191: Reader: implement tail calls

AndyAyersMS:
Closes #191.

Add the `tail` modifier to calls that satisfy correctness checks for tail calls (either explicit or implicit). This will enable LLVM to perform "sibling" call optimizations during lowering.

Do a very simplistic tracking of address-taken locals and args to screen out implicit tail call candidates. Do likewise with localloc. Both these need to be detected during the reader's first pass, so add suitable asserts.

Also start tracking unsafe locals, since they (and localloc) will inspire stack security protection checking (aka /GS) and will inhibit tail calls. We still don't actually do the checks (see #353).

Add logic to the ABI classifier so we can detect if we're introducing on-stack references for call lowerings and avoid tail calling in those cases too. This can also be made smarter (eg we might be able to copy-prop and use the values passed by the caller).

Have the jit options determine if the jit is optimizing and use the option setting in the code rather than checking the flags.

Remove existing NYI for explicit tail calls.

Verified by hand that the excluded tail call tests all now pass and seem to get most of the tail call cases the tests intended, so unexcluded them. Most "normal" bring up tests will create around 150 tail call sites so I think the new codegen gets pretty well tested.

Also verified all this works with EH enabled and GC enabled.


AndyAyersMS:
@JosephTremoulet PTAL

lib/Reader/abisignature.cpp
```diff
@@ -247,6 +247,20 @@ static Value *getFieldAddress(IRBuilder<> &Builder, Value *Base,
   return Address;
 }
 
+bool ABICallSignature::hasIndirectResultOrArg() const {
```
pgavlin: Nit: this should also check for by-value struct arguments. Such arguments will have kind `ABIArgInfo::Direct` and `Arg.getType()->isStructTy()` will return `true`.

AndyAyersMS: I believe there's some paranoia in lowering that should detect this case and avoid the optimization (it looks generally for by-val memory arguments from the caller's frame).

Does this case only happen with SysV? I wanted to test there but still have trouble doing any significant testing on Linux.

pgavlin: It does only happen with SysV, yes.

AndyAyersMS: Ok, let me see if I can do some targeted testing on Linux, just to make sure.

pgavlin: Note that you'll need to avoid testing using P/Invokes that return structs by value until [D14465](http://reviews.llvm.org/D14665) is merged.

lib/Reader/reader.cpp
```diff
@@ -2568,6 +2569,13 @@ void ReaderBase::fgBuildPhase1(FlowGraphNode *Block, uint8_t *ILInput,
       HasLocAlloc = true;
       break;
 
+    case ReaderBaseNS::CEE_LDLOCA:
+    case ReaderBaseNS::CEE_LDLOCA_S:
+    case ReaderBaseNS::CEE_LDARGA:
+    case ReaderBaseNS::CEE_LDARGA_S:
```
JosephTremoulet: The [Lang Ref for the tail annotation](http://llvm.org/docs/LangRef.html#call-instruction) says
> Both markers imply that the callee does not access allocas or varargs from the caller

so does this need to include `CEE_ARGLIST`?

JosephTremoulet:
LGTM aside from the `arglist` question.  Nice to see all those test exclusions disappearing.

AndyAyersMS:
Adding arglist makes sense, will do.

AndyAyersMS:
Updated. I'll let the CI run again before I merge, just because.

AndyAyersMS:
Latest CI failure is unrelated; merging.

PR #958, opened 2015-12-04
Always ask for the non-null entry point for a method

### LINKED ISSUE #951: Roslyn failing in recent CI tests
EG, [this one](http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_roslyn_test/90/).

@richardlford is this a known issue?

### LINKED ISSUE #957: Crossgen of mscorlib timing out with checked build.
@erozenfeld, crossgen of mscorlib is timing out. See the following jobs, 
for example:

- http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_check_win64_new/1/
- http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_ngen_check_win64/153/

AndyAyersMS:
We were seeing two different entry point addresses for some methods.

LLILC will insert an explicit null check for calls in the cases that require one, so always request the non-null entry point for a method. Otherwise we  need to (potentially) keep track of the two entry point addresses for each function, and make sure the right one is used at each call site.

This closes #957 and should also fix #951.

### DIFF
--- a/lib/Reader/reader.cpp
+++ b/lib/Reader/reader.cpp
@@ -4917,10 +4917,12 @@ IRNode *ReaderBase::rdrGetDirectCallTarget(CORINFO_METHOD_HANDLE Method,
     Address = CodePointerLookup.constLookup.addr;
     assert(Address != nullptr);
   } else {
+    // Ask for the generic "ANY" entry point. If NeedsNullCheck is true,
+    // we could instead ask for the CORINFO_ACCESS_NONNULL entry,
+    // but then we'd have to generalize the key used to look things
+    // up in the HandleToGlobalObjectMap.
     CORINFO_CONST_LOOKUP AddressInfo;
-    getFunctionEntryPoint(Method, &AddressInfo, NeedsNullCheck
-                                                    ? CORINFO_ACCESS_NONNULL
-                                                    : CORINFO_ACCESS_ANY);
+    getFunctionEntryPoint(Method, &AddressInfo, CORINFO_ACCESS_ANY);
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@JosephTremoulet PTAL

jkotas:
You should pass in CORINFO_ACCESS_ANY.

CORINFO_ACCESS_NONNULL means that the code is guarantees that the `this` pointer is non-null. If the VM knows that the instance of the given object was allocated, it can provide more direct/efficient entry point for the method.

AndyAyersMS:
Hmm, I'm confused by your comment.

LLILC is ensuring that `this` is non-null via an explicit null check before the call, so I believe we want the more direct entry point.

jkotas:
LLILC is ensuring that `this` is non-null via explicit null check before the call only if NeedsNullCheck is set.

If NeedsNullCheck is not set, the `this` pointer can be null. 

It is corner case that only matters for raw IL - you can call methods with null this pointer in IL. You cannot really do that in C# because C# compiler triggers nullchecks via callvirt close to everywhere.

AndyAyersMS:
Ok, thanks. Will update.

AndyAyersMS:
Updated.

jkotas:
LGTM

AndyAyersMS:
I am pretty sure the CI failures are unrelated (newly added CoreCLR tests that don't work with LLILC), and formatter gripe about header that I didn't touch.

JosephTremoulet:
LGTM too

PR #965, opened 2015-12-08
Add exclusions for new coreclr tests failing on LLILC

richardlford:
Add exclusions corresponding to tests added to
coreclr on core PR #2220.

Tests that did not fail when LLILC EH was enabled
were put into the corresponding section of the
exclusion file.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -2,7 +2,7 @@
     xmlns="http://schemas.microsoft.com/developer/msbuild/2003" >
     <ItemGroup Condition="'$(XunitTestBinBase)' != ''">
         <ExcludeList Include="$(XunitTestBinBase)\JIT\Methodical\tailcall_v4\hijacking\*" >
-             <Issue>13</Issue>
+             <Issue>964</Issue>
         </ExcludeList>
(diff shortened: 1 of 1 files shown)


richardlford:
@pgavlin, PTAL.

richardlford:
The tests being excluded are not on the coreclr master branch yet, but
are on the fork and branch specified on these tests:

- http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_check_win64_prtest/722/
- http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_release_win64_prtest/1256/

These changes should not be checked in unless these tests pass.

AndyAyersMS:
I've already added an exclusion for `tailcall_v4\hijacking` (first one in the list) so we should reconcile the two entries.

richardlford:
@AndyAyersMS, thanks for catching that. I've reconciled by deleting my entry, but copying
the issue number to the existing entry.

richardlford:
Both of the jobs that test this with the CoreClr branch that has the new tests
have passed now, so I'm going to merge in this change (after squashing).

pgavlin:
LGTM.

PR #966, opened 2015-12-09
Exclude SIMD tests that require EH

AndyAyersMS:
Anticipatory pending merge of dotnet/coreclr#2271. Note these fail even when ExecuteHandlers is set.


AndyAyersMS:
@richardlford PTAL

richardlford:
@AndyAyersMS, LGTM.

PR #968, opened 2015-12-09
Update LLILC to generate new funclet EH IR

JosephTremoulet:
LLVM has streamlined the representation for funclet-style EH somewhat, by
pulling the dispatch aspect of catchpad out into a separate catchswitch
instruction, and by explicitly annotating EH pads with their enclosing
parents, obviating the need for endpads.

Update LLILC to generate catchswitch, emit the enclosing pad linkage, and
stop generating endpads.

Also update finally cloning accordingly.

### DIFF
--- a/include/Reader/readerir.h
+++ b/include/Reader/readerir.h
@@ -533,18 +533,14 @@ public:
 
   void fgEnterRegion(EHRegion *Region) override;
 
-  /// Make an EH pad suitable as an unwind label for code in the try region
-  /// protected by the given handler, and any dispatch code needed after the
-  /// EH pad to transfer control to the handler code.
+  /// Make an EH pad suitable as the head of the given handler.
   ///
-  /// \param Handler           Catch/Finally/Fault/Filter region we need to
-  ///                          build an EH pad for.
-  /// \param EndPad [in/out]   CatchEndPad/CleanupEndPad that should be the
-  ///                          unwind target of code in the handler body.
-  /// \param NextPad           Next outer (or successor catch) EH region.
+  /// \param Handler       Catch/Finally/Fault/Filter region we need to
+  ///                      build an EH pad for.
+  /// \param EnclosingPad  EH pad representing the handler enclosing this one
+  ///                      if there is one, ConstantTokenNone otherwise.
   /// \returns A new CatchPad/CleanupPad in a populated EH pad block.
-  llvm::Instruction *createEHPad(EHRegion *Handler, llvm::Instruction *&EndPad,
-                                 llvm::Instruction *NextPad);
+  llvm::Instruction *createEHPad(EHRegion *Handler, llvm::Value *EnclosingPad);
 
   IRNode *fgNodeFindStartLabel(FlowGraphNode *Block) override;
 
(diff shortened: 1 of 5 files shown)


JosephTremoulet:
@AndyAyersMS / @pgavlin PTAL.  This depends on LLVM's [D15139](http://reviews.llvm.org/D15139) and [D15325](http://reviews.llvm.org/D15325), so I won't merge it until those go in (and will have to update it as they evolve), but wanted to go ahead and do the reviews in parallel since this will be needed to unblock FIs once those do go in.

include/Reader/readerir.h
```diff
@@ -533,18 +533,14 @@ class GenIR : public ReaderBase {
...
-  /// Make an EH pad suitable as an unwind label for code in the try region
-  /// protected by the given handler, and any dispatch code needed after the
-  /// EH pad to transfer control to the handler code.
-  ///
-  /// \param Handler           Catch/Finally/Fault/Filter region we need to
-  ///                          build an EH pad for.
-  /// \param EndPad [in/out]   CatchEndPad/CleanupEndPad that should be the
-  ///                          unwind target of code in the handler body.
-  /// \param NextPad           Next outer (or successor catch) EH region.
+  /// Make an EH pad suitable as the head of the hode for the given handler.
```
AndyAyersMS: Typo ... did you mean `node` or `code` ...

JosephTremoulet: Not sure ... don't think I need either.  Updated, thanks.

AndyAyersMS:
I looked through the rest and didn't spot anything obvious. 

I am not all that up to date on the new IR so I'm not sure I would catch any thing subtle.

JosephTremoulet:
I've updated this to also tag calls and invokes in funclets with the "funclet" operand bundle that they require as of [r255674](http://reviews.llvm.org/rL255674) / Microsoft/llvm@b46bb541f5f78520b08.  I pushed it to the PR branch as a separate commit so you can see the diff, but will squash before committing.
@AndyAyersMS PTAL.

JosephTremoulet:
This change and microsoft/llvm#123 will need to go in together; http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1344/ is running a test of the combination.

AndyAyersMS:
Updated part LGTM, just one comment that I think you might reword.

drbo:
Will this make more EH tests pass than before?

JosephTremoulet:
> Will this make more EH tests pass than before?

No, this will let us build against LLVM's ToT, and the EH test pass rate shouldn't really be affected.

JosephTremoulet:
rebased and updated the comment

JosephTremoulet:
Testing with --eh turned up four new failures.  I've updated this change to make their exclusions unconditional, since I think it makes sense to go ahead and unblock FIs and triage these failures along with the other standing eh-related failures.

AndyAyersMS:
Thanks for updating the comment. Let me know if there's anything I can do to help you land this and the LLVM parts....

LGTM

PR #971, opened 2015-12-13
Blanket exclude of new failing tests

AndyAyersMS:
Exclude all new failing tests with generic failure reason. We can sort out why later.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -1,6 +1,1044 @@
 <Project DefaultTargets = "GetListOfTestCmds"
     xmlns="http://schemas.microsoft.com/developer/msbuild/2003" >
     <ItemGroup Condition="'$(XunitTestBinBase)' != ''">
+        <ExcludeList Include="$(XunitTestBinBase)JIT\jit64\mcc\interop\mcc_i87\mcc_i87.cmd.cmd" >
+             <Issue>970</Issue>
+        </ExcludeList>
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
This will likely fail the first go round, but should be pretty close (342 new excludes).

AndyAyersMS:
Missed at least one in release, but will hold off pushing an update while the current test runs are in flight. 

Here's hoping the check build doesn't time out....

AndyAyersMS:
Release had just the one failure. Still waiting on check, no "in flight" failures so far.

AndyAyersMS:
Earlier commit's check run shows 3 more failures and is still going. Have provisionally added these to my not yet pushed update.

AndyAyersMS:
One last failure in check, so 347 excludes in all.

AndyAyersMS:
Am going to go ahead and merge this and hopefully the post-merge CI will be happy.

AndyAyersMS:
Odd, the baseline jit failed every test case. Wonder if there was some breaking change in the interface or something... let's see how the new jit fares.

AndyAyersMS:
In the new jit, all tests passed, but there were a few cases where the jit skipped methods. Mostly varargs brought in by the new tests, but also a few cases where we had IR validation errors. [Details here](http://dotnet-ci.cloudapp.net/job/dotnet_llilc_release_win64/593/console).

AndyAyersMS:
Verification error cases should have been excluded but I had `.cmd.cmd` at the end of the test name so they ran. Will fix this and exclude the varargs cases and put out a new update.

Issue #970, opened 2015-12-11
Need more exclusions


AndyAyersMS:
Based on my attempted merge of [latest LLVM bits](http://dotnet-ci.cloudapp.net/job/dotnet_llilc_llvm_prtest/194/) we need at least 158 more exclusions.

Some of these appear to be cases that require EH and work with ExecuteHandlers, but others don't.

@richardlford I know you are tracking some exclusion updates, so let's coordinate effort here.

AndyAyersMS:
For the "debug with release" LLVM update, I got 167 failures:
```
JIT.XUnitWrapper         Total: 3823, Errors: 0, Failed: 167, Skipped: 0, Time: 10710.669s
```

JosephTremoulet:
I ran into similar trying to test #968 with Microsoft/llvm@8cec2f281, so I ran [a pr job with no changes](http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1317/) to try to get a baaseline.  The "release with asserts" build timed out

```
  Build timed out (after 240 minutes). Marking the build as aborted.
  04:05:40.886 
  Build was aborted
```

and the release build had 343 failures:

```
=== TEST EXECUTION SUMMARY ===
   Interop.XUnitWrapper     Total:    1, Errors: 0, Failed:   0, Skipped: 0, Time:    4.430s
   JIT.XUnitWrapper         Total: 4096, Errors: 0, Failed: 343, Skipped: 0, Time: 2960.625s
   readytorun.XUnitWrapper  Total:    2, Errors: 0, Failed:   0, Skipped: 0, Time:    4.498s
                                   ----          -          ---           -        ---------
                      GRAND TOTAL: 4099          0          343           0        2969.554s (2982.336s)
```

which is a lot more than you're seeing, so maybe something funny is going on or maybe I somehow submitted that incorrectly.

JosephTremoulet:
I saw similar results [on a rerun](https://github.com/dotnet/llilc/issues/970) baseline job, except that only 342 tests failed in the release leg:

```
=== TEST EXECUTION SUMMARY ===
   Interop.XUnitWrapper     Total:    1, Errors: 0, Failed:   0, Skipped: 0, Time:    3.326s
   JIT.XUnitWrapper         Total: 4096, Errors: 0, Failed: 342, Skipped: 0, Time: 2907.833s
[...]

PR #974, opened 2015-12-15
Exclude failing HFA tests

AndyAyersMS:
Still using blanket reason #970, need triage.


AndyAyersMS:
@JosephTremoulet PTAL

AndyAyersMS:
Looks like there are some non-HFA failures from the chk build for #973 so this may need tweaking.

AndyAyersMS:
Added four more excludes (none from HFA tests).

JosephTremoulet:
LGTM

PR #978, opened 2015-12-16
Exclude some varargs tests, suppress popups in local testing

AndyAyersMS:
With this set of four tests excluded, LLILC no longer skips any methods in the 3874 tests that remain.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -1036,6 +1036,18 @@
         <ExcludeList Include="$(XunitTestBinBase)\JIT\jit64\mcc\interop\mcc_i07\mcc_i07.cmd" >
              <Issue>970</Issue>
         </ExcludeList>
+        <ExcludeList Include="$(XunitTestBinBase)\JIT\jit64\mcc\interop\mcc_i10\mcc_i10.cmd" >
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@JosephTremoulet or @richardlford PTAL

JosephTremoulet:
I think for llilc_run.py it would make sense to have a command line flag that lets popups happen, no?
Does the lab invoke tests using llilc_runtest.py?  If so, is the redundancy troubling, should we stop explicitly blocking popups in the lab env script?

AndyAyersMS:
I thought the lab used `LLILCENV.cmd` which is what I changed originally, but looking at it now, it uses llilc_runtest.py, which creates a new LLILCENV on the fly. So that part is not redundant.

As for llilc_run.py, `-x` could be fixed to allow popups, by not having it assume an AltJit prefix, and by processing it after everything else. How does that sound?

JosephTremoulet:
SGTM

AndyAyersMS:
Updated.

JosephTremoulet:
LGTM

AndyAyersMS:
Something wiped out the last batch of test runs. Pretty sure this change is good, but retesting just to be cautious.

PR #979, opened 2015-12-18
Add CoreDisTools Library

swaroop-sridhar:
This change adds CoreDisTools -- a library containing
LLVM disassembly tools for CoreClr to LLILC repo.

Currently, the library supports instruction-wise
disassembly API required for implementing GcStress.

coredistools.h -- the export/import header
coredistools.cpp -- the implementation as a wrapper around
    LLVM disassemblers to suite the required API.

This change also adds code to package the library into a
NuGet package.

### DIFF
--- a/include/CoreDisTools/coredistools.h
+++ b/include/CoreDisTools/coredistools.h
@@ -1,0 +1,64 @@
+//===--------- coredistools.h - Dissassembly tools for CoreClr
+//-------------===//
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// \brief Core Disassembly Tools API for CoreClr AOT/JIT
+///
+//===----------------------------------------------------------------------===//
+
+#if !defined(LLILC_TOOLS_COREDISTOOLS)
+#define LLILC_TOOLS_COREDISTOOLS
+
+#include <stdint.h>
+
+#if defined(__cplusplus)
+#define EXTERN_C extern "C"
+#else
+#define EXTERN_C
+#endif // defined(__cplusplus)
+
+#if defined(_MSC_VER)
+#if defined(DllInterfaceExporter)
+#define DllIface EXTERN_C __declspec(dllexport)
+#else
+#define DllIface EXTERN_C __declspec(dllimport)
+#endif // defined(DllInterfaceExporter)
+#else
+#define DllIface
+#endif // defined(_MSC_VER)
+
+enum TargetArch {
+  Target_Host, // Target is the same as host architecture
+  Target_X86,
+  Target_X64,
+  Target_Thumb,
+  Target_Arm64
+};
+
+struct CorDisasm;
+
+DllIface CorDisasm *InitDisasm(enum TargetArch Target);
+
+// DisasmInstruction -- Disassemble one instruction
+// Arguments:
+// Address -- The address at which the bytes of the instruction
+//            are intended to execute
+// Bytes -- Pointer to the actual bytes which need to be disassembled
+// MaxLength -- Number of bytes available in Bytes buffer
+// PrintAssembly -- Whether to dump the decoded instriction to stdout
+// Returns:
+//   -- The Size of the disassembled instruction
+//   -- Zero on failure
+DllIface size_t DisasmInstruction(const CorDisasm *Disasm, size_t Address,
+                                  const uint8_t *Bytes, size_t Maxlength,
+                                  bool PrintAssembly);
+
+DllIface void FinishDisasm(const CorDisasm *Disasm);
+
+#endif // !defined(LLILC_TOOLS_COREDISTOOLS)
--- a/lib/CoreDisTools/.nuget/runtime.json
+++ b/lib/CoreDisTools/.nuget/runtime.json
@@ -1,0 +1,19 @@
+{
+  "runtimes": {
+    "win7-x64": {
+      "Microsoft.DotNet.CoreDisTools": {
+        "toolchain.win7-x64.Microsoft.DotNet.CoreDisTools": "1.0.0-prerelease"
+      }
+    },
+    "ubuntu.14.04-x64": {
+      "Microsoft.DotNet.CoreDisTools": {
+        "toolchain.ubuntu.14.04-x64.Microsoft.DotNet.CoreDisTools": "1.0.0-prerelease"
+      }
+    },
+    "osx.10.10-x64": {
+      "Microsoft.DotNet.CoreDisTools": {
+        "toolchain.osx.10.10-x64.Microsoft.DotNet.CoreDisTools": "1.0.0-prerelease"
+      }
+    }
+  }
+}
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -1,0 +1,226 @@
+//===-------- coredistools.cpp - Dissassembly tools for CoreClr -----------===//
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// \brief Implementation of Disassembly Tools API for AOT/JIT
+///
+//===----------------------------------------------------------------------===//
+
+#include "llvm/ADT/Optional.h"
+#include "llvm/ADT/Triple.h"
+#include "llvm/MC/MCAsmInfo.h"
+#include "llvm/MC/MCContext.h"
+#include "llvm/MC/MCDisassembler.h"
+#include "llvm/MC/MCInst.h"
+#include "llvm/MC/MCInstPrinter.h"
+#include "llvm/MC/MCInstrInfo.h"
+#include "llvm/MC/MCObjectFileInfo.h"
+#include "llvm/MC/MCRegisterInfo.h"
+#include "llvm/MC/MCSubtargetInfo.h"
+#include "llvm/Support/Format.h"
+#include "llvm/Support/ManagedStatic.h"
+#include "llvm/Support/PrettyStackTrace.h"
+#include "llvm/Support/Signals.h"
+#include "llvm/Support/SourceMgr.h"
+#include "llvm/Support/TargetRegistry.h"
+#include "llvm/Support/TargetSelect.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/DataTypes.h"
+
+#define DllInterfaceExporter
+#include "coredistools.h"
+
+using namespace llvm;
+using namespace std;
+
+// Instruction-wise disassembler helper.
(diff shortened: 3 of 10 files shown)


swaroop-sridhar:
@RussKeldorph please review.

include/CoreDisTools/coredistools.h
```diff
@@ -0,0 +1,36 @@
...
+/// \file
+/// \brief Core Disassembly Tools API for CoreClr AOT/JIT
+///
+//===----------------------------------------------------------------------===//
+
+#if !defined(LLILC_TOOLS_COREDISTOOLS)
+#define LLILC_TOOLS_COREDISTOOLS
+
+#if !defined(DllIface)
+#define DllIface extern "C" __declspec(dllimport)
```
RussKeldorph: I would expect this header to be compatible with C unless there is some convention discouraging that that I'm not aware of (possible).  Constructs like `extern "C"` need a `defined(__cplusplus)` guard.  Also, is `__declspec(dllimport)` relevant/appropriate on non-Win?

include/CoreDisTools/coredistools.h
```diff
@@ -0,0 +1,36 @@
...
+#endif // defined(DllIface)
+
+enum TargetArch {
```
RussKeldorph: Related to C-compat comment above.  I think this should be a `typedef`.

include/CoreDisTools/coredistools.h
```diff
@@ -0,0 +1,36 @@
...
+  Target_Host,
```
RussKeldorph: I believe there's a subtle ambiguity with `Target_Host` on Linux/ARM since there both ARM and Thumb code can exist in the same process.  This may not matter on Windows since generally only Thumb is allowed, though the concept is not unheard of even on Windows since I believe even Windows mixes x86 and x64 under WoW.

One option would be to have a major/minor architecture parameter, similar to how LLVM classifies things, but I think it's probably simpler to just make Thumb a peer of ARM for our purposes.

NOTE that I'd also probably be ok with just ignoring this issue under the assumption that this library is purpose-built for CoreCLR, which doesn't acknowledge the possibility of more than one simultaneous target architecture.  In that case, what you've done makes perfect sense, though I would probably rename Target_Arm to Target_Thumb to allow for unambiguous future expansion.

include/CoreDisTools/coredistools.h
```diff
@@ -0,0 +1,36 @@
...
+  Target_X86,
+  Target_X64,
+  Target_Arm,
+  Target_Arm64
+};
+
+class CorDisasm;
```
RussKeldorph: `struct` instead of `class` for C compat?

kyulee1:
Can you fix the build break and squash the change? I will take a look then.

swaroop-sridhar:
@kyulee1: the build issue is now fixed. The remaining error is an unrelated llilc test timeout

include/CoreDisTools/coredistools.h
```diff
@@ -0,0 +1,52 @@
...
+enum TargetArch {
+  Target_Host, // Target is the same as host architecture
+  Target_X86,
+  Target_X64,
+  Target_Thumb,
+  Target_Arm64
+};
+
+struct CorDisasm;
+DllIface CorDisasm *InitDisasm(enum TargetArch Target);
```
kyulee1: Can you make Target argument default as Target_Host so that we can skip setting it on the same  Host/Target?

lib/CoreDisTools/.nuget/toolchain.ubuntu.14.04-x64.Microsoft.DotNet.Coredistools.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <projectUrl>https://github.com/dotnet/coreclr</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Machine Code Disassembly Tools</description>
+    <releaseNotes>Initial release</releaseNotes>
+    <copyright>Copyright &#169; Microsoft Corporation</copyright>
+  </metadata>
+  <files>
+    <file src="../libcoredistools.so" target="runtimes/ubuntu.14.04-x64/native/libcoredistools.so" />
+    <file src="../include/coredistools.h" target="runtimes/win7-x64/native/coredistools.h" />
```
kyulee1: Please fix the folder name "win7-x64" -> "ubuntu.14.04-x64".

lib/CoreDisTools/.nuget/toolchain.osx.10.10-x64.Microsoft.DotNet.Coredistools.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <projectUrl>https://github.com/dotnet/coreclr</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Machine Code Disassembly Tools</description>
+    <releaseNotes>Initial release</releaseNotes>
+    <copyright>Copyright &#169; Microsoft Corporation</copyright>
+  </metadata>
+  <files>
+    <file src="../libcoredistools.dylib" target="runtimes/osx.10.10-x64/native/libcoredistools.dylib" />
+    <file src="../include/coredistools.h" target="runtimes/win7-x64/native/coredistools.h" />
```
kyulee1: Please fix the folder name "win7-x64" -> "osx.10.10-x64".

lib/CoreDisTools/coredistools.exports
```diff
@@ -0,0 +1,3 @@
+InitDisasm
+FinishDisasm
+DisasmInstruction
```
kyulee1: It appears PrintInstruction is an export in the header file but you didn't specify it here.

lib/CoreDisTools/coredistools.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+DllIface void FinishDisasm(const CorDisasm *Disasm) { delete Disasm; }
+
+DllIface size_t DisasmInstruction(const CorDisasm *Disasm, size_t Address,
+                                  const uint8_t *Bytes, size_t Maxlength) {
+  assert((Disasm != nullptr) && "Disassembler object Expected ");
+  return Disasm->disasmInstruction(Address, Bytes, Maxlength);
+}
+
+DllIface void PrintInstruction(const CorDisasm *Disasm, size_t Address,
```
kyulee1: PrintInstruction seems to do disasm and print instruction. Can we just add an optional argument to DisasmInstruction for this purpose instead of exporting it?

lib/CoreDisTools/coredistools.cpp
```diff
@@ -0,0 +1,236 @@
...
+    TargetArch = Target;
+  }
+
+private:
+  bool setTarget();
+
+  TargetArch TargetArch;
+  string TargetTriple;
+  const Target *TheTarget;
+  bool ShowInstructionBytes;
```
kyulee1: Is this always true? than we don't need to set this here.

include/CoreDisTools/coredistools.h
```diff
@@ -0,0 +1,52 @@
...
+DllIface size_t DisasmInstruction(const CorDisasm *Disasm, size_t Address,
+                                  const uint8_t *Bytes, size_t Maxlength);
+DllIface void PrintInstruction(const CorDisasm *Disasm, size_t Address,
+                               const uint8_t *Bytes, size_t Maxlength);
+DllIface void FinishDisasm(const CorDisasm *Disasm);
+
+#endif // defined(LLILC_TOOLS_COREDISTOOLS)
```
kyulee1: Nit. The comment is about the negative definition -- !defined(LLILC_TOOLS_COREDISTOOLS).

include/CoreDisTools/coredistools.h
swaroop-sridhar: @kyulee1: We're not using the default parameters in the DLL interface in order to keep the header compatible with C. Please see @RussKeldorph's comment above.

kyulee1: Ok. That's fine without having a default value.

lib/CoreDisTools/coredistools.cpp
swaroop-sridhar: We cannot use Optional arguments because of C compatibility. But I can add a mandatory argument to DisasmInstruction like: 

```
  DllIface size_t DisasmInstruction(const CorDisasm *Disasm, size_t Address,
                                                          const uint8_t *Bytes, size_t Maxlength, bool PrintAsm)
```

However I left it as two functions, because `DisasmInstruction` is designed for GcStress which is the sole user currently (to get the instruction size) and `PrintInstruction` is mostly used for debugging.

include/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 add_subdirectory(clr)
 add_subdirectory(Driver)
+add_subdirectory(CoreDisTools)
```
kyulee1: I know you're the one that adds this. But all of this additions including this CmakeLists.txt are useless. If you could clean up these since you're here, it would be great. Or at least let's not add this anymore including the empty CmakeLists.txt under CoreDisTools.

lib/CoreDisTools/coredistools.cpp
kyulee1: PrintInstruction functionally does DisasmInstruction and prints it, so I don't think it's much useful to have a separate API. Can you add the mandatory argument to DisasmInsutrction as you suggested while deleting PrintInstruction?

swaroop-sridhar:
Thanks for the review @kyulee1. I've addressed your comments in https://github.com/swaroop-sridhar/llilc/commit/bea7195c2be408f524239d5696c849e390dff39b.

lib/CoreDisTools/.nuget/runtime.json
```diff
@@ -0,0 +1,19 @@
+{
+  "runtimes": {
+    "win7-x64": {
+      "Microsoft.DotNet.CoreDisTools": {
+        "toolchain.win7-x64.Microsoft.DotNet.Coredistools": "1.0.0-prerelease"
```
kyulee1: Case should be aligned for all filenames/ids: "Coredistools" -> "CoreDisTools".
Please make sure this in the other cases.

swaroop-sridhar: Fixed, thanks.

kyulee1:
@swaroop-sridhar The file names are still not updated --e.g,  lib/CoreDisTools/.nuget/toolchain.osx.10.10-x64.Microsoft.DotNet.Coredistools.nuspec ==> lib/CoreDisTools/.nuget/toolchain.osx.10.10-x64.Microsoft.DotNet.CoreDisTools.nuspec

swaroop-sridhar:
I've renamed the files in the .nuget directory. The .h, .cpp, .exports files are lower cased, following the convention in the other directories.

kyulee1:
LGTM. Please squash change and fix formatting before merge.

PR #986, opened 2015-12-23
Refine error classifications

AndyAyersMS:
Started in on reclassifying the #970 errors. Some tests now pass, some pass with EH, others now have more specific issues to track why they fail. Still a bunch more to work through as well.

Also excluded a few varargs tests, so that for the tests that do run, LLILC does not skip any methods, and normalized all the varargs exclusions to reference dotnet/microsoft#1440.

### DIFF
--- a/test/exclusion.targets
+++ b/test/exclusion.targets
@@ -1,256 +1,70 @@
 <Project DefaultTargets = "GetListOfTestCmds"
     xmlns="http://schemas.microsoft.com/developer/msbuild/2003" >
     <ItemGroup Condition="'$(XunitTestBinBase)' != ''">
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@richardlford PTAL

richardlford:
@AndyAyersMS, LGTM assuming things still pass. What method are you using to reclassify. Do you remove the exclusions and rerun so you can see the failures again, or are you running the tests individually to diagnose?

AndyAyersMS:
My method is nothing fancy -- run excluded tests individually and make suitable updates to the exclusion list, then rerun the entire suite with the updated excludes. If a bunch of tests fail the same way I'll open an issue.

AndyAyersMS:
I can't repro either of the two failures seen so far in the check CI run.

AndyAyersMS:
Updated to exclude those two failures and two more tests that only fail if EH is enabled.

PR #987, opened 2015-12-24
Fix bad exclude syntax

AndyAyersMS:
Was causing the rolling CI to fail with skipped varargs methods.


AndyAyersMS:
Am just going to commit this directly since it's trivial.

PR #988, opened 2015-12-30
CoreDisTools: Handle disassembly of X86 Prefix bytes.

swaroop-sridhar:
On X86, LLVM disassembler does not handle instruction prefixes
correctly -- please see LLVM bug 7709.
The disassembler reports instruction prefixes separate from the
actual instruction. In order to work-around this problem, we
continue decoding past the prefix bytes.

### DIFF
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -78,7 +78,26 @@ bool CorDisasm::setTarget() {
 
   switch (TargetArch) {
   case Target_Host:
+    switch (TheTriple.getArch()) {
+    case Triple::x86:
+      TargetArch = Target_X86;
+      break;
+    case Triple::x86_64:
+      TargetArch = Target_X64;
+      break;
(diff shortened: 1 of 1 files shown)


swaroop-sridhar:
@AndyAyersMS  please review.

AndyAyersMS:
Given the way LLVM models x86 machine IR, I'm fine with this approach.

Is there any way we can ensure that the opcode numbers remain correct over time? I don't know how likely they are to ever change but it seems like they could. For instance, include some inline test case or something so the code can self-verify? Given how often we take in new LLVM bits this seems prudent. Feel free to open an issue for this and address it separately.

swaroop-sridhar:
Thanks for the feedback @AndyAyersMS  -- Actually the LLVM opcode numbers have already changed because of a recent FI. 

So, instead of checking for LLVM Opcodes, here's an alternate implementation that checks Machine opcode: https://github.com/dotnet/llilc/pull/992

PR #991, opened 2016-01-02
Exclude four tests that are timing out

JosephTremoulet:
Timeouts seem to have started with a recent LLVM FI (Microsoft/llvm#127).


JosephTremoulet:
@AndyAyersMS PTAL.  The same tests that timed out in Microsoft/llvm#127 are still timing out in Microsoft/llvm#128, so I've opened #990 to track and this change to exclude them.

AndyAyersMS:
Interesting to see this happen with an FI. Excluding them for now seems ok, the first two are crazy stress tests so are probably beating up some part of LLVM pretty badly.

LGTM.

PR #998, opened 2016-01-13
Exclusion of b99969

kyulee1:
https://github.com/dotnet/coreclr/commit/8353857ecea0fb4410d72fcf69aeb55fb09feef5 enabled this test,
but it appears it fails with llilc due to EH.


AndyAyersMS:
LGTM

AndyAyersMS:
Looking back at the CHK leg history, I think you also need to exclude `JIT\RyuJIT\DoWhileBndChk\DoWhileBndChk.cmd`

kyulee1:
That's right. Added one more exclusion.

PR #999, opened 2016-01-14
Handle alignment in reserveAllocationSpace

JosephTremoulet:
Change r257294 (git hash cb4ab730d69) added alignment parameters to this
interface.  Update the signature of the override in LLILC's
EEMemoryManager, and update its body to take them into account when
computing the alignment to request from the CLR EE.

### DIFF
--- a/include/Jit/EEMemoryManager.h
+++ b/include/Jit/EEMemoryManager.h
@@ -93,13 +93,18 @@ public:
   /// allocate all sections to be loaded.
   ///
   /// \param CodeSize - the total size of all code sections.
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
@russellhadley PTAL.  This will need to go in at the same time as Microsoft/llvm#131 (which this is intended to unblock).  http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1383/ is running to test them together.

russellhadley:
LGTM - once the test exception goes in of course.

Issue #1000, opened 2016-01-14
Failure in JIT/Performance/CodeQuality/V8/DeltaBlue/DeltaBlue


JosephTremoulet:
The stack trace indicates a bounds-check failure:

```
00:04:57         0x00007FF9E269871C (0x000000E7F10128C0 0x00007FF900000008 0x0000000000000000 0x0000000000000000), RaiseException() + 0x6C bytes(s)
00:04:57         0x00007FF9CE4AD4E6 (0x000000E7F0E992B0 0x0000000000000000 0x0000000000000000 0x0000000000000000), LinkFrameAndThrow() + 0xA6 bytes(s), d:\j\workspace\dotnet_llilc_check_win64_prtest\coreclr\src\vm\excep.cpp, line 7471
00:04:57         0x00007FF9CEB36639 (0x0000000000000000 0x0000000000000000 0x0000000000000000 0x0000000000000000), NakedThrowHelper2() + 0x9 bytes(s), d:\j\workspace\dotnet_llilc_check_win64_prtest\coreclr\src\vm\amd64\RedirectedHandledJITCase.asm, line 232
00:04:57         0x00007FF9CEB3666B (0x00007FF96FAC0782 0x0000000000010020 0x00000001FFFFFFFF 0x000000E7F10128C0), NakedThrowHelper_RspAligned() + 0x1E bytes(s), d:\j\workspace\dotnet_llilc_check_win64_prtest\coreclr\src\vm\amd64\RedirectedHandledJITCase.asm, line 236 + 0x2B byte(s)
00:04:57         0x00007FF9CEB36675 (0x0000000000010020 0x00000001FFFFFFFF 0x000000E7F10128C0 0x000000E7F10128C0), NakedThrowHelper_FixRsp() + 0x5 bytes(s), d:\j\workspace\dotnet_llilc_check_win64_prtest\coreclr\src\vm\amd64\RedirectedHandledJITCase.asm, line 236 + 0x35 byte(s)
00:04:57         0x00007FF96FAC0782 (0x00000001FFFFFFFF 0x000000E7F10128C0 0x000000E7F10128C0 0x000000E7F0E9A250) <unknown module>
00:04:57         0x0000000000010020 (0x000000E7F10128C0 0x000000E7F10128C0 0x000000E7F0E9A250 0x00007FF9CE311E41) <unknown module>
[...]

PR #1001, opened 2016-01-14
Exclude V8\DeltaBlue

JosephTremoulet:
This is a new test that is failing with LLILC.  Issue #1000 opened to
track.


JosephTremoulet:
@AndyAyersMS PTAL

russellhadley:
LGTM

AndyAyersMS:
I left a goofy try/catch in there, so yes, sadly, it does rely on catching a bounds check exception.
I'll fix the test.
```C#
    public void inst_main(String[] args)
    {
        int iterations = 200; // read iterations from arguments, walter 7/97
        try
        {   // read iterations from arguments, walter 7/97
            iterations = Int32.Parse(args[0]);
        }
        catch (Exception)
        {
        }
        inst_inner(iterations, true);
    }
```

PR #1002, opened 2016-01-14
Fix ObjectWriter version and update package script to handle Python 3.

kyulee1:


russellhadley:
Why do we need to support python 3?  I thought we standardized on 2.7?  Do the VSO machines have an inconsistent python?  Can the machine maintainers fix it?  I don't have a problem with the code per se but the real problem seems to be the variation in machine set up.

kyulee1:
This is internal/local script that I used so far so it doesn't matter actually in either case. Pat is sending official one for VSO consumption -- https://github.com/dotnet/llilc/pull/1003.  I may delete this script later, but let's keep it for now.

PR #1010, opened 2016-01-22
Don't register LLVM signal handlers in CoreCLR tools

swaroop-sridhar:
Objectwriter and ObjectWriter and CoreDistools as tools loaded as DLLs.
They currently always register signal handlers for SIGABRT and SIGSEGV,
which interferes with CoreCLR's signal handlers. This caused unexpected
behavior and dumps in certain GCStress runs. So, this change removes
the use of LLVM's signal handlers.


swaroop-sridhar:
@kyulee1 please review

kyulee1:
LGTM

PR #1011, opened 2016-01-26
Support for CFI Unwind

kyulee1:
This adds the following CFI unwind emission.
EmitCFIStart
EmitCFIEnd
EmitCFIBlob

The first two sets the range of the code that the frame covers.
The last one should be invoked for each offset of the code, if any.
The blob data format is shared by the one RyuJit layouts.

### DIFF
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -19,6 +19,7 @@
 #include "llvm/MC/MCAsmBackend.h"
 #include "llvm/MC/MCAsmInfo.h"
 #include "llvm/MC/MCContext.h"
+#include "llvm/MC/MCDwarf.h"
 #include "llvm/MC/MCInstPrinter.h"
 #include "llvm/MC/MCInstrInfo.h"
 #include "llvm/MC/MCObjectFileInfo.h"
@@ -48,6 +49,7 @@
 #include "llvm/Target/TargetMachine.h"
 #include "llvm/Target/TargetOptions.h"
 #include "llvm/Target/TargetSubtargetInfo.h"
+#include "cfi.h"
 
 using namespace llvm;
 using namespace llvm::codeview;
@@ -145,6 +147,7 @@ public:
   std::unique_ptr<raw_fd_ostream> OS;
   MCTargetOptions MCOptions;
   std::vector<DebugLocInfo> DebugLocInfos;
+  bool frameOpened;
 
 public:
   bool init(StringRef FunctionName);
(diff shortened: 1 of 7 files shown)


lib/ObjWriter/objwriter.cpp
```diff
@@ -420,6 +425,48 @@ extern "C" void EmitFrameInfo(ObjectWriter *OW, const char *FunctionName,
                 4);
 }
 
+extern "C" void EmitCFIStart(ObjectWriter *OW, int Offset) {
```
kyulee1: Offset is not used, instead MC API will create/insert label to the code directly. But I'd like to keep this on purpose in case we need offset information later. The same thing applies to the following APIs.

kyulee1:
The corresponding RyuJit change is -- https://github.com/dotnet/coreclr/pull/2868
@AndyAyersMS @pgavlin PTAL

AndyAyersMS:
LGTM.

Is it the case that callers should always invoke these methods in a particular sequence eg: start, blob, [blob,] end? It would be nice to add checks for this but it doesn't look like there is any convenient place to record a bit of state.

kyulee1:
That's right although such assertion is done in CoreRT or LLVM backend.
I added checks.

kyulee1:
Updated the code to consume cfi.h from CoreCLR.

lib/ObjWriter/objwriter.cpp
```diff
@@ -420,6 +420,53 @@ extern "C" void EmitFrameInfo(ObjectWriter *OW, const char *FunctionName,
...
+  assert(OW && "ObjWriter is null");
+  assert(OW->frameOpened && "frame should be opened before CFIEnd");
+  auto *AsmPrinter = &OW->getAsmPrinter();
+  auto &OST = *AsmPrinter->OutStreamer;
+
+  OST.EmitCFIEndProc();
+  OW->frameOpened = false;
+}
+
+extern "C" void EmitCFIBlob(ObjectWriter *OW, int Offset, const char *Blob) {
```
pgavlin: So is the idea here that CoreRT will do something like:
```
EmitCFIStart(ow, ...);

uint8_t *cfiBlob = ...; uint32_t cfiBlobSize = ...;
for (int offset = 0; offset < cfiBlobSize; offset += ?) {
    EmitCFIBlock(ow, &cfiBlob[offset]);
}
EmitCFIEnd(ow);
```

kyulee1: Here is full code for it -- https://github.com/dotnet/corert/pull/752

pgavlin: I might call this "EmitCFICode" instead of "EmitCFIBlob", then. IMO "EmitCFIBlob" implies that this function may process multiple CFI instructions.

kyulee1: Makes sense. Will update it.

PR #1013, opened 2016-01-27
Adopted updated Orc API

JosephTremoulet:
The object set passed through the lower layers is now a set of
`OwningBinary<ObjectFile>`s rather than a set of `ObjectFile`s directly;
update accordingly.

Also, relocation resolution in the stock Orc layers is now deferred until
getSymbol finalization, which runs after our LoadListener gets a chance to
inspect the relocations and report them to the CLR EE, which means that we
can switch to the stock ObjectLinkingLayer instead of continuing to
propagate forward a modified copy of it in LLILC sources.  It would still
be good to refactor some of the OrcJit code so LLILC could properly take
over relocation processing, but since the order is that the dynamic loader
parses relocations and copies any addends from the fixup locations to side
tables, then LLILC calls the CLR's recordRelocation which applies the
relocation in place, then the dynamic loader's resolveRelocation runs
using the stashed-away addends, the double relocation application is
harmless and we can drop this ongoing source of merge pain.

With this change, EESymbolResolver::findSymbol gets called on more
symbols, including some symbols not in the map.  Rather than expanding its
assertion to have the same exclusions as the similar assertion in
recordRelocations, simply remove the assert and rely on the
recordRelocations checking to catch misuses.


JosephTremoulet:
This change and Microsoft/llvm#133 have to go in at the same time; http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1398/ tests them together except for a subsequent change to fix the formatting error reported there, which subsequently passed formatting checks at http://dotnet-ci.cloudapp.net/job/dotnet_llilc_code_formatting_prtest/1133/

JosephTremoulet:
@AndyAyersMS / @erozenfeld PTAL

erozenfeld:
LGTM

PR #1015, opened 2016-01-27
Use operand bundles for gc-transition arguments

JosephTremoulet:
The prior approach of eagerly generating gc.statepoint calls with
gc-transition arguments is being deprecated (see e.g. r258551), in favor
of attaching these as operand bundles, which should make the semantics
less opaque during optimization.

Since the lowering code that looks for the bundles is in
RewriteStatepointsForGC, we now need to run that pass even with
conservative GC, in methods that have gc transition bundles.  Add a
post-pass to mark all other calls "gc-leaf-function" in that configuration
so that RewriteStatepointsForGC will leave them alone.  Expose a getter on
GenIR to indicate whether we had to generate a managed frame so that we
can avoid this in the common case.

Regardless of GC flavor, inline asm calls (which we use for nops) now must
be explicitly annotated "gc-leaf-function" to avoid lowering trying to
process them as statepoints.


JosephTremoulet:
@pgavlin / @swaroop-sridhar PTAL

include/Reader/readerir.h
```diff
@@ -814,6 +814,12 @@ class GenIR : public ReaderBase {
                          bool IsRelocatable, bool IsCallTarget,
                          bool IsFrozenObject = false);
 
+  /// Indicate whether this method needed an unmanaged frame generated for it
+  /// (to make a call to unmanaged code).
+  ///
+  /// \returns True iff an unmanaged frame was created.
+  bool getHasUnmanagedFrame() { return UnmanagedCallFrame != nullptr; }
```
pgavlin: Nit: I might call this `containsUnmanagedCall()` for the sake of clarity. Similar feedback applies for the output parameter added above.

pgavlin:
LGTM aside from one nit. This is much cleaner than eagerly emitting statepoints; thanks for making the change.

swaroop-sridhar:
Green for me too, thanks.

JosephTremoulet:
Updated with the suggested rename.  Will merge if the original CI testing (http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1406/) finishes cleanly.

PR #1017, opened 2016-01-29
Remove references to deleted options

JosephTremoulet:
Remove the code that sets `spp-no-statepoints` and
`rs4gc-use-deopt-bundles` to true; the options have been removed in
r259096 and r259129 and the behavior is now always as though they were set
to true.

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -201,17 +201,6 @@ ICorJitCompiler *__stdcall getJit() {
       Opts["disable-cgp-gc-opts"]->addOccurrence(0, "disable-cgp-gc-opts",
                                                  "true");
     }
-    // Set flags indicating to statepoint lowering that call->statepoint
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
This will need to go in along with Microsoft/llvm#136; testing for the pair is running at http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1412/

JosephTremoulet:
@pgavlin / @swaroop-sridhar PTAL

swaroop-sridhar:
Green, thanks.

JosephTremoulet:
The CI job failed due to the FI also including llvm-mirror/llvm@1689efb/r259130 which requires the objwriter change in #1018.  New CI job with both fixes and the FI is running at http://dotnet-ci.cloudapp.net/view/dotnet_llilc/job/dotnet_llilc_prtest/1415/

PR #1019, opened 2016-01-29
Custom Data Section Support for ObjWriter

nattress:
Allow consumers of ObjWriter to define their own object file sections.
This commit adds support for simple data sections configured to be
writeable / read-only in each of the LLVM-supported output object file
formats.

### DIFF
--- a/lib/ObjWriter/.nuget/Microsoft.DotNet.ObjectWriter.nuspec
+++ b/lib/ObjWriter/.nuget/Microsoft.DotNet.ObjectWriter.nuspec
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>Microsoft.DotNet.ObjectWriter</id>
-    <version>1.0.4-prerelease-00001</version>
+    <version>1.0.5-prerelease-00002</version>
     <title>Microsoft .NET Object File Generator</title>
     <authors>Microsoft</authors>
     <owners>Microsoft</owners>
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -26,6 +26,8 @@
 #include "llvm/MC/MCObjectFileInfo.h"
 #include "llvm/MC/MCParser/AsmLexer.h"
 #include "llvm/MC/MCRegisterInfo.h"
+#include "llvm/MC/MCSectionCOFF.h"
+#include "llvm/MC/MCSectionELF.h"
 #include "llvm/MC/MCSectionMachO.h"
 #include "llvm/MC/MCStreamer.h"
 #include "llvm/MC/MCSubtargetInfo.h"
@@ -148,6 +150,7 @@ public:
   MCTargetOptions MCOptions;
   std::vector<DebugLocInfo> DebugLocInfos;
   bool frameOpened;
+  std::map<std::string, MCSection *> CustomSections;
 
 public:
   bool init(StringRef FunctionName);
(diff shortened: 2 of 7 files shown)


nattress:
@kyulee1 Please take a look

lib/ObjWriter/.nuget/Microsoft.DotNet.ObjectWriter.nuspec
```diff
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>Microsoft.DotNet.ObjectWriter</id>
-    <version>1.0.4-prerelease-00001</version>
+    <version>1.0.6-prerelease-00001</version>
```
kyulee1: Can you change it to "1.0.5-prerelease-00002" instead? This is minor change that doesn't break the existing API. By increasing the postfix number, we don't need to update CoreRT consumption.
The same applies to all below.

lib/ObjWriter/objwriter.cpp
```diff
@@ -249,6 +252,58 @@ extern "C" void FinishObjWriter(ObjectWriter *OW) {
   delete OW;
 }
 
+extern "C" void CreateDataSection(ObjectWriter *OW, const char *SectionName,
+                                  bool ReadOnly)
+{
```
kyulee1: { starts the end of function declaration per LLILC coding convention (which aligns with LLVM).
I think the following code does not follow this either.
Please run codeformatter tool to fix this -- https://github.com/dotnet/llilc/blob/master/Documentation/Code-Formatting.md
Or PR run will fail. cc, for details --  @swaroop-sridhar

lib/ObjWriter/objwriter.cpp
```diff
@@ -249,6 +252,58 @@ extern "C" void FinishObjWriter(ObjectWriter *OW) {
...
+          if (!ReadOnly)
+          {
+            Flags |= ELF::SHF_WRITE;
+          }
+          Section = OutContext.getELFSection(SectionName, 
+              ELF::SHT_PROGBITS, Flags);
+          break;
+        }
+    case Triple::UnknownObjectFormat:
+      assert(false && "Unknown output target");
```
kyulee1: Can you replace "false &&" by "!" per convention?

lib/ObjWriter/objwriter.cpp
```diff
@@ -249,6 +252,58 @@ extern "C" void FinishObjWriter(ObjectWriter *OW) {
...
+          unsigned Flags = ELF::SHF_ALLOC;
+          if (!ReadOnly)
+          {
+            Flags |= ELF::SHF_WRITE;
+          }
+          Section = OutContext.getELFSection(SectionName, 
+              ELF::SHT_PROGBITS, Flags);
+          break;
+        }
+    case Triple::UnknownObjectFormat:
```
kyulee1: I think it should be "default:" instead so that we can capture any unknown (future) section that we don't handle.

kyulee1:
Note currently ObjectWriter is not automatically published although @pgavlin is working on it.
Until then, I will push it manually when this is merged.

lib/ObjWriter/.nuget/Microsoft.DotNet.ObjectWriter.nuspec
nattress: @kyulee1 Will CoreRT get this updated package if I do that? I need this change in CoreRT as soon as it's merged basically.

kyulee1: As commented below in the main reply, once it's merged, I will push it manually. But we don't need to update CoreRT which will automatically pick the latest one in case this postfix increment.

nattress: Ah I understand now. That sound better; I'll do that :)

lib/ObjWriter/objwriter.cpp
nattress: I was following the LLVM convention (e.g: MCObjectFileInfo::InitMCObjectFileInfo) but I can switch this to default if you prefer.

kyulee1: Yep. I saw that too. But default would capture this as well as any other case that may be added later. In fact, we might want to change InitMCObjectFileInfo in LLVM too.

lib/ObjWriter/objwriter.cpp
```diff
@@ -249,6 +252,56 @@ extern "C" void FinishObjWriter(ObjectWriter *OW) {
...
+  MCSection *Section = nullptr;
+  SectionKind Kind = ReadOnly ? 
+      SectionKind::getReadOnly() : SectionKind::getData();
+
+  switch (TheTriple.getObjectFormat()) {
+    case Triple::MachO:
+      Section = OutContext.getMachOSection("__DATA", SectionName, 0, Kind);
+      break;
+    case Triple::COFF:
+      {
```
pgavlin: Braces are off here and in the if/else statements throughout. As per the LLVM coding standards, opening braces should be on the same line as the construct (e.g. `if (...) {`, `case Foo: {`, etc.).

lib/ObjWriter/objwriter.cpp
```diff
@@ -249,6 +252,56 @@ extern "C" void FinishObjWriter(ObjectWriter *OW) {
   delete OW;
 }
 
+extern "C" void CreateDataSection(ObjectWriter *OW, const char *SectionName, bool ReadOnly) {
```
pgavlin: Nit: s/`ReadOnly`/`IsReadOnly`/

lib/ObjWriter/objwriter.cpp
```diff
@@ -249,6 +252,56 @@ extern "C" void FinishObjWriter(ObjectWriter *OW) {
...
+          if (!ReadOnly)
+          {
+            Flags |= ELF::SHF_WRITE;
+          }
+          Section = OutContext.getELFSection(SectionName, 
+              ELF::SHT_PROGBITS, Flags);
+          break;
+        }
+    default:
+      assert(!"Unknown output target");
```
pgavlin: Nit: s/target/format/

pgavlin: Also, should we think about returning an error in this case instead of just asserting?

lib/ObjWriter/objwriter.cpp
nattress: ccformat.py will fix this up. Stay tuned.

lib/ObjWriter/objwriter.cpp
nattress: Done

lib/ObjWriter/objwriter.cpp
nattress: Yeah - building a Binder style world full of asserts at debug and AVs in retail is not ideal.  I'll make the function return bool and return an error if the target is unknown.  I'll leave the other APIs alone but error handling in those would be useful in the future.

PR #1020, opened 2016-01-30
Restore EEObjectLinkingLayer

JosephTremoulet:
Something was flawed in my analysis of relocation handling in 10a9856, and
the double processing is causing ngen with llilc to generate incorrect
code.

Revert to our previous strategy of dragging along a copy of orc's
ObjectLinkingLayer with the relocation resolution suppressed, pending
working out an upstream refactoring to better handle this use case.


JosephTremoulet:
@AndyAyersMS PTAL

AndyAyersMS:
LGTM

PR #1021, opened 2016-02-03
Emit PDB local var debug information in object writer

hoyMS:
This change enables displaying local var and parameter values during debugging coreRT apps on Windows.

The build script is also changed to allow the use of two Code View header files from Micorsoft pdb repo https://github.com/Microsoft/microsoft-pdb.git. To build LLILC without ObjectWriter, use the original command line :

cmake -G "Visual Studio 14 2015 Win64" %llilcSrc% -DWITH_CORECLR=%coreclrBuild% -DWITH_LLVM=%llvmBuild% -DLLVM_OPTIMIZED_TABLEGEN=ON

To build LLILC with ObjectWriter which in turn checks out the MS pdb git repo, use
cmake -G "Visual Studio 14 2015 Win64" %llilcSrc% -DWITH_CORECLR=%coreclrBuild% -DWITH_LLVM=%llvmBuild% -DLLVM_OPTIMIZED_TABLEGEN=ON -DOBJWRITER=1


hoyMS:
@kyulee1 PTAL

kyulee1:
Can you make build successful before I review? You should follow LLILC coding convention per wiki.
Can you split the file to accommodate debug info generation in a separate file? I couldn't even add comment since this caused many changes in a big file.

hoyMS:
@kyulee1 Is it allowed to change the public header files in coreclr? For example,  corhdr.h doesn’t compile when included by LLVM. One problem is the LPVOID type used in corhdr.h is from windows 10 SDK c:\Program Files (x86)\Windows Kits\10\Include\10.0.10586.0\shared\minwindef.h and it looks that LLVM compilation doesn’t involve that file.


#ifdef _MSC_VER
#pragma warning(disable:4200) // nonstandard extension used : zero-sized array in struct/union.
#endif
typedef LPVOID  mdScope;                // Obsolete; not used in the runtime.
typedef ULONG32 mdToken;                // Generic token

kyulee1:
As I send PR https://github.com/dotnet/coreclr/pull/3006, you could now include cordebuginfo.h once it's merged. There are only two things DWORD and PORTABILITY_WARNING which you can locally define within LLILC for the inclusion of this header.

AndyAyersMS:
Doesn't LLVM have the codeview headers we need? It seems odd to have to pull in an entire new repo for this.

kyulee1:
We don't want to bring the entire pdb repo in the build system like this. We are considering either change build setup as prerequisite for ObjectWriter build only in LLILC or copying such headers into LLILC. This will be short-term until codeview.h in LLVM has full information about CV.

hoyMS:
@dotnet-bot test his please

kyulee1:
@hoyMS Can you please run "python utils\ccformat.py --untidy" per https://github.com/dotnet/llilc/blob/master/Documentation/Code-Formatting.md. lab will fail unless format is fixed. And also please remove unnecessary changes for the easier review.

hoyMS:
@kyulee1 Thanks. Clang-format gave me this which looks weird. Is this expected?

const CVRegMapping cvRegMapAmd64[] = {
    CVREGDAT(REGNUM_RAX, CV_AMD64_RAX) CVREGDAT(REGNUM_RCX, CV_AMD64_RCX)
        CVREGDAT(REGNUM_RDX, CV_AMD64_RDX) CVREGDAT(REGNUM_RBX, CV_AMD64_RBX)
            CVREGDAT(REGNUM_RSP, CV_AMD64_RSP) CVREGDAT(
                REGNUM_RBP, CV_AMD64_RBP) CVREGDAT(REGNUM_RSI, CV_AMD64_RSI)
                CVREGDAT(REGNUM_RDI, CV_AMD64_RDI) CVREGDAT(
                    REGNUM_R8, CV_AMD64_R8) CVREGDAT(REGNUM_R9, CV_AMD64_R9)
                    CVREGDAT(REGNUM_R10, CV_AMD64_R10)
                        CVREGDAT(REGNUM_R11, CV_AMD64_R11)
                            CVREGDAT(REGNUM_R12, CV_AMD64_R12)
                                CVREGDAT(REGNUM_R13, CV_AMD64_R13)
                                    CVREGDAT(REGNUM_R14, CV_AMD64_R14)
                                        CVREGDAT(REGNUM_R15, CV_AMD64_R15)};

lib/ObjWriter/jitDebugInfo.h
```diff
@@ -0,0 +1,46 @@
+#ifndef JIT_DEBUG_INFO_H
+#define JIT_DEBUG_INFO_H
+
+typedef unsigned int DWORD;
+#define _TARGET_AMD64_ 1
+
```
kyulee1: You might just include "LLILCPal.h" to handle the above specialization.

lib/ObjWriter/jitDebugInfo.h
```diff
@@ -0,0 +1,46 @@
...
+struct DebugLocInfo {
+  int NativeOffset;
+  int FileId;
+  int LineNumber;
+  int ColNumber;
+};
+
+struct DebugVarInfo {
+  int varNumber;
```
kyulee1: I don't see the need of 'varNumber' other than asserting it which is doable locally.

lib/ObjWriter/objwriter.cpp
```diff
@@ -550,22 +547,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
   OST.EmitCVFileDirective(FileId, FileName);
 }
 
-// This should be invoked at the end of function code emission to flush
-// debug function info.
-extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
-                                      const char *FunctionName,
-                                      int FunctionSize) {
+static void EmitPDBDebugVarInfo(MCObjectStreamer &OST, const MCSymbol *Fn,
```
kyulee1: Can you change "EmitPDBDebugVarInfo" to "EmitCVDebugVarInfo"? Since we're not in PDB yet. 
Ideally I like it to be an instance method of ObjectWriter object instead of this static function, but that part we can follow up later all together as well.

lib/ObjWriter/objwriter.cpp
```diff
@@ -550,22 +547,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+  MCContext &OutContext = OST.getContext();
+  const MCObjectFileInfo *MOFI = OutContext.getObjectFileInfo();
+  assert(MOFI->getObjectFileType() == MOFI->IsCOFF);
+
+  for (int i = 0; i < NumVarInfos; i++) {
+    // Emit an S_LOCAL record
+    DebugVarInfo var = LocInfos[i];
+    LocalSym sym;
+
+    memset(&sym, 0, sizeof(sym));
```
kyulee1: You can just declare like this instead of memset:
```
LocalSym sym = {};
```

lib/ObjWriter/objwriter.cpp
```diff
@@ -550,22 +547,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+
+    int sizeofSym = sizeof(LocalSym);
+    short rectyp = SymbolRecordKind::S_LOCAL;
+    short reclen = sizeof(rectyp) + sizeofSym + var.name.length() + 1;
```
kyulee1: Well. We now have struct layout so we can compute the length of symbol upfront like this. Another way is to just emit two labels (start/end) and emit difference so that we don't make any mistake. But this is fine with me, too.
But should we use "SymRecord" struct for this? I'm seeing this pattern in other places that emits length and kind. We might add a function that takes this common pattern.

lib/ObjWriter/objwriter.cpp
```diff
@@ -550,22 +547,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+    TypeIndex typeIndex(var.typeIndex);
+    sym.Type = typeIndex;
 
+    if (var.isParam) {
+      sym.Flags |= LocalSym::IsParameter;
+    }
+
+    OST.EmitIntValue(reclen, 2);
```
kyulee1: OST.EmitInt16(reclen)

lib/ObjWriter/objwriter.cpp
```diff
@@ -550,22 +547,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+
+      // Emit a range record
+      LocalVariableAddrRange *prange = nullptr;
+
+      switch (range.loc.vlType) {
+      case ICorDebugInfo::VLT_REG:
+      case ICorDebugInfo::VLT_REG_FP: {
+        DefRangeRegisterSym rec;
+        int sizeofRec = sizeof(rec);
+        memset(&rec, 0, sizeofRec);
```
kyulee1: Same thing for Struct initialization.

lib/ObjWriter/objwriter.cpp
```diff
@@ -599,19 +693,34 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
     // This is the important bit that tells the debugger where the function
     // code is located and what's its size:
     OST.EmitIntValue(FunctionSize, 4);
-    OST.EmitFill(12, 0);
+    OST.EmitFill(4, 0);                // SS_DBGSTART
```
kyulee1: Well. You are now in here. Can you use "ProcSym" to emit function symbol instead of emitting individual bytes like this? Then we don't need to add bunch of comments like this.

lib/ObjWriter/objwriter.cpp
```diff
@@ -599,19 +693,34 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+    OST.EmitIntValue(FunctionSize, 4); // SS_DBGEND
+    OST.EmitFill(4, 0);                // SS_TINDEX
     OST.EmitCOFFSecRel32(Fn);
     OST.EmitCOFFSectionIndex(Fn);
-    OST.EmitIntValue(0, 1);
+
+    // Emit flags, optimize debugging
+    OST.EmitIntValue(0x80, 1);
```
kyulee1: Where  0x80 comes from? I guess "ProcFlags::HasOptimizeDebugInfo".

lib/ObjWriter/objwriter.cpp
```diff
@@ -622,6 +731,46 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+                                      const char *FunctionName,
+                                      int FunctionSize) {
+  assert(OW && "ObjWriter is null");
+  auto *AsmPrinter = &OW->getAsmPrinter();
+  auto &OST = static_cast<MCObjectStreamer &>(*AsmPrinter->OutStreamer);
+  MCContext &OutContext = OST.getContext();
+  const MCObjectFileInfo *MOFI = OutContext.getObjectFileInfo();
+
+  if (MOFI->getObjectFileType() == MOFI->IsCOFF) {
+    EmitPDBDebugFunctionInfo(OW, FunctionName, FunctionSize);
```
kyulee1: Likewise, EmitPDBDebugFunctionInfo -> EmitCVDebugFunctionInfo ?

lib/ObjWriter/objwriter.cpp
```diff
@@ -622,6 +731,46 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+    return;
+
+  // Interpret ranges as an array of NativeVarInfo structs.
+  ICorDebugInfo::NativeVarInfo *varInfos =
+      (ICorDebugInfo::NativeVarInfo *)ranges;
+
+  int varNumber = varInfos[0].varNumber;
+
+  DebugVarInfo var = {varNumber, name, typeIndex, isParm, NULL};
+  var.ranges = new std::vector<ICorDebugInfo::NativeVarInfo>();
```
kyulee1: I think this is memory leak. You could declare vector inside struct instead of vector*.

lib/ObjWriter/objwriter.cpp
```diff
@@ -622,6 +731,46 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+
+  for (int i = 0; i < rangeCount; i++) {
+    assert(var.varNumber == varInfos[i].varNumber);
```
kyulee1: You could assert varNumber == varInfos[i].varNumber. But let's not allocate "varNumber" inside DebugVarInfo as mentioned above.

lib/ObjWriter/objwriter.cpp
```diff
@@ -658,4 +807,4 @@ extern "C" void EmitDebugModuleInfo(ObjectWriter *OW) {
   OST.SwitchSection(Section);
   OST.EmitCVFileChecksumsDirective();
   OST.EmitCVStringTableDirective();
-}
+}
```
kyulee1: Nit: Add new line same as before.

lib/ObjWriter/objwriter.cpp
```diff
@@ -550,22 +547,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+        // TODO: for optimized debugging
+        break;
+
+      default:
+        _ASSERTE(!"Unknown varloc type!");
+        break;
+      }
+
+      // Emit range
+      if (prange != nullptr) {
```
kyulee1: This is about range emission which is part of symbol emission above. I like it to be a function that is invoked for each case above instead of falling thru like this.

kyulee1:
@hoyMS format tool is not perfect and some of things we should handle manually. Per coding convention, please make sure local/param/fields names should be capitalized, which also the tool appear to miss.

lib/ObjWriter/objwriter.cpp
```diff
@@ -550,22 +547,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+
+      case ICorDebugInfo::VLT_REG_BYREF:
+      case ICorDebugInfo::VLT_STK_BYREF:
+      case ICorDebugInfo::VLT_REG_REG:
+      case ICorDebugInfo::VLT_REG_STK:
+      case ICorDebugInfo::VLT_STK_REG:
+      case ICorDebugInfo::VLT_STK2:
+      case ICorDebugInfo::VLT_FPSTK:
+      case ICorDebugInfo::VLT_FIXED_VA:
+        // TODO: for optimized debugging
```
kyulee1: Is this right comment? I thought you're emitting a way of optimized function per function symbol setting.

lib/ObjWriter/jitDebugInfo.h
hoyMS: Tried but didn't work. Saw new build failures such as

 C:\Program Files (x86)\Windows Kits\10\Include\10.0.10586.0\shared\minwindef.h(156): error C2371: 'DWORD': redefinition; different basic types [d:\build\llilc-build\lib\ObjWriter\objwrit
er.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10586.0\um\winnt.h(2591): error C2733: '_BitScanForward': second C linkage of overloaded function not allowed [d:\build\llilc-build\li
b\ObjWriter\objwriter.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10586.0\um\winnt.h(2598): error C2733: '_BitScanReverse': second C linkage of overloaded function not allowed [d:\build\llilc-build\li
b\ObjWriter\objwriter.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10586.0\um\winnt.h(2605): error C2733: '_BitScanForward64': second C linkage of overloaded function not allowed [d:\build\llilc-build\
lib\ObjWriter\objwriter.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.10586.0\um\winnt.h(2612): error C2733: '_BitScanReverse64': second C linkage of overloaded function not allowed [d:\build\llilc-build\

lib/ObjWriter/jitDebugInfo.h
hoyMS: Removed.

lib/ObjWriter/objwriter.cpp
hoyMS: Done.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/jitDebugInfo.h
kyulee1: From the error message, didn't you remove "typedef unsigned int DWORD;" yet?

lib/ObjWriter/objwriter.cpp
hoyMS: Using SymRecord struct is fine, but still need to explicitly set the fields everywhere, like
SymRecord rec = {sizeof(rectyp) + sizeofSym + var.name.length() + 1, SymbolRecordKind::S_LOCAL}

lib/ObjWriter/objwriter.cpp
hoyMS: Got an error " error C2039: 'EmitInt16': is not a member of 'llvm::MCObjectStreamer'". Looks like MCObjectStreamer' doesn't have the member method 'EmitInt16'.

lib/ObjWriter/jitDebugInfo.h
hoyMS: Looks like LLILCPal.h doesn't contain the definition of DWORD. Removing "typedef unsigned int DWORD;" gives :
D:\git\llilc\lib\ObjWriter\objwriter.cpp(277): error C2589: 'constant': illegal token on right side of '::' [d:\build\llilc-build\lib\ObjWriter\objwriter.vcxproj]
D:\git\llilc\lib\ObjWriter\objwriter.cpp(277): error C2059: syntax error: '::' [d:\build\llilc-build\lib\ObjWriter\objwriter.vcxproj]
D:\git\llilc\lib\ObjWriter\objwriter.cpp(280): error C2589: 'constant': illegal token on right side of '::' [d:\build\llilc-build\lib\ObjWriter\objwriter.vcxproj]
D:\git\llilc\lib\ObjWriter\objwriter.cpp(280): error C2059: syntax error: '::' [d:\build\llilc-build\lib\ObjWriter\objwriter.vcxproj]

lib/ObjWriter/objwriter.cpp
kyulee1: It's okay to set it explicitly but it gives us a hint what we are emitting.
With regard of sizeof(rectyp) which is always 2 because we don't count the length field (2byte) but starting with kind field. Such subtle meaning is defined in the header file at least.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
kyulee1: Okay. Let's keep this. I saw such things is asm streamer recently..

lib/ObjWriter/objwriter.cpp
hoyMS: I meant to say that those cases does not exist when minopts is on.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Thanks for catching this. Forgot this is C++ not C#. Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Done

lib/ObjWriter/jitDebugInfo.h
kyulee1: This is due to conflict on COFF related constant in enumeration in LLVM and macro in winnt.h And also you need to include "globals.h" to make this work.

Let's keep your change as it is for now and we need some discussions.

lib/ObjWriter/objwriter.cpp
kyulee1: This part should be removed now.

hoyMS: Fixed

lib/ObjWriter/objwriter.cpp
```diff
@@ -596,31 +682,42 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
-    // Some bytes of this segment don't seem to be required for basic debugging,
-    // so just fill them with zeroes.
-    OST.EmitFill(12, 0);
-    // This is the important bit that tells the debugger where the function
-    // code is located and what's its size:
-    OST.EmitIntValue(FunctionSize, 4);
-    OST.EmitFill(12, 0);
+    SymRecord RecHeader = {ulittle16_t(sizeof(ulittle16_t) + sizeof(ProcSym) +
+                                       strlen(FunctionName) + 1),
+                           ulittle16_t(SymbolRecordKind::S_GPROC32_ID)};
```
kyulee1: I think this looks good to me.
Can you do similar things in other cases like DefRangeRegisterSym, etc?
Ideally I would like it to be function something like this.
```

EmitSymRecord(int size /* the size of the following symbol */, int symbolKind)
{
SymbolRecord SR = { size + sizeof(little16_t), symbolKind }; // include the size of kind.
OST.EmitBytes(StringRef((char *)&SR, sizeof(SR)));
}
```

lib/ObjWriter/objwriter.cpp
```diff
@@ -596,31 +682,42 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+      std::unique_ptr<DebugVarInfo[]> varInfos(new DebugVarInfo[NumVarInfos]);
+      std::copy(OW->DebugVarInfos.begin(), OW->DebugVarInfos.end(),
+                varInfos.get());
+      EmitCVDebugVarInfo(OST, Fn, varInfos.get(), NumVarInfos);
+      OW->DebugVarInfos.clear();
+    }
 
     // We're done with this function.
     OST.EmitIntValue(0x0002, 2);
     OST.EmitIntValue(unsigned(SymbolRecordKind::S_PROC_ID_END), 2);
```
kyulee1: If you have a function about SymbolRecord that I mentioned above, you can simple invoke it with 0 size since it doesn't need a following symbol emission.

lib/ObjWriter/objwriter.cpp
```diff
@@ -596,31 +682,42 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+    OST.EmitBytes(StringRef((char *)&RecHeader, sizeof(RecHeader)));
+    OST.EmitBytes(StringRef((char *)&ProRec, offsetof(ProcSym, CodeOffset)));
+
+    // Emit relocation
     OST.EmitCOFFSecRel32(Fn);
     OST.EmitCOFFSectionIndex(Fn);
-    OST.EmitIntValue(0, 1);
+
+    // Emit flags
+    OST.EmitIntValue(ProRec.Flags, sizeof(ProRec.Flags));
```
kyulee1: Didn't you set an optimized debug flag to it before? I don't see where the Flags are set.

lib/ObjWriter/objwriter.cpp
```diff
@@ -559,22 +556,111 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+      case ICorDebugInfo::VLT_REG:
+      case ICorDebugInfo::VLT_REG_FP: {
+        DefRangeRegisterSym rec = {};
+        int sizeofRec = sizeof(rec);
+        short rectyp = SymbolRecordKind::S_DEFRANGE_REGISTER;
+        short reclen = sizeofRec + sizeof(rectyp);
+        rec.Range.OffsetStart = range.startOffset;
+        rec.Range.Range = range.endOffset - range.startOffset;
+        rec.Range.ISectStart = 0;
+        rec.Register = cvRegMapAmd64[range.loc.vlReg.vlrReg];
```
kyulee1: Should we add assertion for the table access at minimum?

kyulee1:
@hoyMS I'm still seeing lower case for all locals/param/fields. Please use the capital letters per coding convention.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: That flag is not necessarily useful and the flag is defined in cvinfo.h.

kyulee1: Okay. I assume setting the Flags is unnecessary here so we emit 0 flag.

lib/ObjWriter/objwriter.cpp
```diff
@@ -596,31 +682,42 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
-    OST.EmitBytes(FunctionName);
-    OST.EmitIntValue(0, 1);
-    OST.EmitLabel(ProcSegmentEnd);
+    OST.EmitBytes(StringRef(FunctionName, strlen(FunctionName) + 1));
+
+    // Emit local var info
+    int NumVarInfos = OW->DebugVarInfos.size();
+    if (NumVarInfos > 0) {
+      std::unique_ptr<DebugVarInfo[]> varInfos(new DebugVarInfo[NumVarInfos]);
+      std::copy(OW->DebugVarInfos.begin(), OW->DebugVarInfos.end(),
```
kyulee1: Do you really need a copy here since you've already retained the data structure in vector?
You could simply pass &OW->DebugVarInfos[0] to EmitCVDebugVarInfo instead of the copied on varInfos.get().

lib/ObjWriter/objwriter.cpp
```diff
@@ -631,6 +728,44 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+extern "C" void EmitDebugVar(ObjectWriter *OW, char *name, int typeIndex,
+                             bool isParm, int rangeCount, char *ranges) {
+  assert(OW && "ObjWriter is null");
+
+  if (rangeCount == 0)
+    return;
+
+  // Interpret ranges as an array of NativeVarInfo structs.
+  ICorDebugInfo::NativeVarInfo *varInfos =
+      (ICorDebugInfo::NativeVarInfo *)ranges;
```
kyulee1: I think type cast is not needed. Instead, you could declare the type for ranges as "NativeVarInfo*" instead of "char *" in the function argument since you are marshalling the array of struct.

lib/ObjWriter/objwriter.cpp
hoyMS: Yes.

lib/ObjWriter/objwriter.cpp
hoyMS: Yes, this sounds better. Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed

lib/ObjWriter/jitDebugInfo.h
```diff
@@ -0,0 +1,40 @@
...
+
+struct DebugLocInfo {
+  int NativeOffset;
+  int FileId;
+  int LineNumber;
+  int ColNumber;
+};
+
+struct DebugVarInfo {
+  std::string name;
```
kyulee1: Use capital letter for fields.

lib/ObjWriter/objwriter.cpp
```diff
@@ -559,22 +556,122 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
   }
 }
 
-// This should be invoked at the end of function code emission to flush
-// debug function info.
-extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
-                                      const char *FunctionName,
-                                      int FunctionSize) {
+static void EmitSymRecord(MCObjectStreamer &OST, int size, int symbolKind) {
```
kyulee1: Use capital letter for parameters.

lib/ObjWriter/objwriter.cpp
```diff
@@ -559,22 +556,122 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+  SymRecord Rec = {ulittle16_t(size + sizeof(ulittle16_t)),
+                   ulittle16_t(symbolKind)};
+  OST.EmitBytes(StringRef((char *)&Rec, sizeof(Rec)));
+}
+
+static void EmitVarDefRange(MCObjectStreamer &OST, const MCSymbol *Fn,
+                            LocalVariableAddrRange &Range) {
```
kyulee1: Use capital letter for parameters.

lib/ObjWriter/objwriter.cpp
```diff
@@ -559,22 +556,122 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+  const MCSymbolRefExpr *baseSym =
```
kyulee1: Use capital letter for locals.

lib/ObjWriter/objwriter.cpp
```diff
@@ -559,22 +556,122 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+      case ICorDebugInfo::VLT_REG_STK:
+      case ICorDebugInfo::VLT_STK_REG:
+      case ICorDebugInfo::VLT_STK2:
+      case ICorDebugInfo::VLT_FPSTK:
+      case ICorDebugInfo::VLT_FIXED_VA:
+        // TODO: for optimized debugging
+        break;
+
+      default:
+        _ASSERTE(!"Unknown varloc type!");
```
kyulee1: Can you use just assert?

lib/ObjWriter/objwriter.cpp
```diff
@@ -631,6 +734,41 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+
+extern "C" void EmitDebugVar(ObjectWriter *OW, char *name, int typeIndex,
+                             bool isParm, int rangeCount,
+                             ICorDebugInfo::NativeVarInfo *ranges) {
+  assert(OW && "ObjWriter is null");
+
+  if (rangeCount == 0)
+    return;
+
+  int varNumber = ranges[0].varNumber;
```
kyulee1: varNumber is just used in assertion. This will fail with release build. Just inline it inside assert below.

lib/ObjWriter/objwriter.cpp
```diff
@@ -631,6 +734,41 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+
+  OW->DebugVarInfos.push_back({name, typeIndex, isParm});
+  DebugVarInfo &var = OW->DebugVarInfos.back();
```
kyulee1: Use capital letter for locals.

lib/ObjWriter/objwriter.cpp
```diff
@@ -631,6 +734,41 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+
+  for (int i = 0; i < rangeCount; i++) {
```
kyulee1: Use capital letter for locals.

lib/ObjWriter/objwriter.cpp
```diff
@@ -631,6 +734,41 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+    // TODO: Should convert this for non-Windows.
+  }
+}
+
+extern "C" void EmitDebugVar(ObjectWriter *OW, char *name, int typeIndex,
+                             bool isParm, int rangeCount,
+                             ICorDebugInfo::NativeVarInfo *ranges) {
+  assert(OW && "ObjWriter is null");
+
+  if (rangeCount == 0)
```
kyulee1: Do we ever have case with 0 count? Shouldn't we just assert this is not case?

lib/ObjWriter/objwriter.cpp
kyulee1: I don't see the update yet.

lib/ObjWriter/objwriter.cpp
```diff
@@ -559,22 +556,122 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+  OST.EmitIntValue(Range.Range, 2);
+}
+
+static void EmitCVDebugVarInfo(MCObjectStreamer &OST, const MCSymbol *Fn,
+                               DebugVarInfo LocInfos[], int NumVarInfos) {
+  MCContext &OutContext = OST.getContext();
+  const MCObjectFileInfo *MOFI = OutContext.getObjectFileInfo();
+  assert(MOFI->getObjectFileType() == MOFI->IsCOFF);
+
+  for (int i = 0; i < NumVarInfos; i++) {
```
kyulee1: Use capital letter for locals.

kyulee1:
@hoyMS I'm not sure why new push doesn't trigger the test. Please make sure it runs tests by sending a message to dotnet-bot.

lib/ObjWriter/jitDebugInfo.h
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Which one?

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

hoyMS:
It triggers.

lib/ObjWriter/objwriter.cpp
kyulee1: I meant EmitSymRecord function which now you've fixed them all. With regard to SymbolKind, it would be better to use precise enumeration type "SymbolRecordKind" instead of int for the better type safety.

hoyMS: Fixed.

kyulee1:
@hoyMS As shown in the lab result link above, linux builds are failing. Please update the build break.

lib/ObjWriter/objwriter.cpp
```diff
@@ -559,22 +556,119 @@ extern "C" void EmitDebugFileInfo(ObjectWriter *OW, int FileId,
...
+    // Emit an S_LOCAL record
+    DebugVarInfo Var = LocInfos[I];
+    LocalSym Sym = {};
+
+    int SizeofSym = sizeof(LocalSym);
+    EmitSymRecord(OST, SizeofSym + Var.Name.length() + 1,
+                  SymbolRecordKind::S_LOCAL);
+
+    TypeIndex TypeIndex(Var.TypeIndex);
+    Sym.Type = TypeIndex;
```
kyulee1: Nit: Using the same type/local name,  TypeIndex is odd. Can you just inline the above to the below.

lib/ObjWriter/objwriter.cpp
```diff
@@ -631,6 +730,37 @@ extern "C" void EmitDebugFunctionInfo(ObjectWriter *OW,
...
+}
+
+extern "C" void EmitDebugVar(ObjectWriter *OW, char *Name, int TypeIndex,
+                             bool IsParm, int RangeCount,
+                             ICorDebugInfo::NativeVarInfo *Ranges) {
+  assert(OW && "ObjWriter is null");
+  assert(RangeCount != 0);
+  DebugVarInfo NewVar(Name, TypeIndex, IsParm);
+  OW->DebugVarInfos.push_back(NewVar);
+  DebugVarInfo &Var = OW->DebugVarInfos.back();
```
kyulee1: Nit: You can push_back after the loop below, then you don't need this operation.

kyulee1:
@hoyMS LGTM other than a few nits. You can just check code-format locally and merge since you've already tested this.

lib/ObjWriter/objwriter.cpp
hoyMS: That may result in copying a large array from NewVar to &Var, due to how verctor.copy is implemented.

lib/ObjWriter/objwriter.cpp
hoyMS: Fixed.

hoyMS:
@kyulee1 I don't have write access. Could you please do the merge for me?

lib/ObjWriter/objwriter.cpp
kyulee1: I don't think ranges for a variable is not that many especially once we introduce a scope concept.
If performance is a concern, then we should change data structure.
Logically, constructing and initializing NewVar as a whole and then pushing/publishing it to container is more readable than splitting like this.. Ideally I would put range initialization insider the constructor of DebugVarInfo.

hoyMS: Ranges are produced basic block wise by Ryujit. So given a large CFG with dozens of local vars, the number of range may amount to thousands.

kyulee1: As I said above, the ranges (that Jit reports) will be filtered by the scope eventually in CoreRT. So at most we will have one or two per variable.

hoyMS: That sounds good. Just moved the push_back after the loop.

kyulee1:
Thanks for the work!

PR #1022, opened 2016-02-04
Expect `Indirect` pointers in stack map

JosephTremoulet:
LLVM change r256352 (llvm-mirror/llvm@da801219) had the effect of
switching the reporting at statepoints in LLILC from being annotated with
LocationKind::Direct (which would mean that the GC pointer being reported
is a pointer to the stack location indicated) to being annotated with
LocatoinKind::Indirect (which, correctly, means taht the stack location
being reported holds a GC pointer that the GC must track/update).  Update
LLILC's expectations when parsing the stack map accordingly.

### DIFF
--- a/lib/GcInfo/GcInfo.cpp
+++ b/lib/GcInfo/GcInfo.cpp
@@ -580,7 +580,7 @@ void GcInfoEmitter::encodeTrackedPointers(const GcFuncInfo *GcFuncInfo) {
         assert(false && "GC-Pointer Live in Register");
         break;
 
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
@swaroop-sridhar PTAL

swaroop-sridhar:
Green, thanks.

JosephTremoulet:
@dotnet-bot retest this please (first round of tests hit some infrastructure hiccup restoring packages)

PR #1024, opened 2016-02-09
Support Dwarf LineNumber

kyulee1:
This enables Dwarf line number emission for ObjectWriter.
Reused the code for Dwarf assembly debugging emission.
More fine grained logic like function/vars should be supported later.
With this, we can break into C# source using gdb/lldb on the binary
built from CoreRT in Linux/Osx.


kyulee1:
@AndyAyersMS PTAL.
@hoyMS PTAL. we should coordinate this with your pending work since it will conflict yours although the change is minimal.

hoyMS:
LGTM

Issue #1025, opened 2016-02-10
Investigate failure in JIT\opt\Inline\regression\mismatch32\mismatch32 and mismatch64


JosephTremoulet:
This causes an assertion in LLVM:

```
16:04:29 d:\j\workspace\dotnet_llilc_check_win64_prtest\coreclr\tests\src\TestWrappers_x64_Debug\JIT\JIT.XUnitWrapper.cs(146330): error : JIT._opt_Inline_regression_mismatch32_mismatch32_mismatch32_._opt_Inline_regression_mismatch32_mismatch32_mismatch32_cmd [FAIL] [d:\j\workspace\dotnet_llilc_check_win64_prtest\coreclr\tests\runtest.proj]
16:04:29         Assertion failed: isa<X>(Val) && "cast<Ty>() argument of incompatible type!", file D:\j\workspace\dotnet_llilc_check_win64_prtest\LLVM\include\llvm/Support/Casting.h, line 237
16:04:29         0x00007FF802D64095 (0x0000000000000016 0x00007FF802A9FEA6 0x00000000000000ED 0x00007FF80D949594), HandleAbort() + 0x5 bytes(s), d:\j\workspace\dotnet_llilc_check_win64_prtest\llvm\lib\support\windows\signals.inc, line 390
16:04:29         0x00007FF80D9A4A68 (0x0000000000000001 0x00007FF800000000 0x0000000000000000 0x00000061B9989550), raise() + 0x1E8 bytes(s)
16:04:29         0x00007FF80D9A5A31 (0x00007FF800000003 0x0000000000000003 0x00007FF803A2E720 0x00007FF803A2E650), abort() + 0x31 bytes(s)
16:04:29         0x00007FF80D9A7A81 (0x00000000000000ED 0x00007FF803A2E720 0x00000061B9C2F2E0 0x0000293130303030), _get_wide_winmain_command_line() + 0x1171 bytes(s)
16:04:29         0x00007FF80D9A7D8F (0x00000061B9C25A10 0x00000061B998A2E0 0x00000061B9C25A10 0x0000000000000000), _wassert() + 0x3F bytes(s)
[...]

PR #1026, opened 2016-02-10
Exclude test inline\regression\mismatch32

JosephTremoulet:
Opened #1025 to investigate.


JosephTremoulet:
@AndyAyersMS PTAL.  The failure shows up in the testing for Microsoft/llvm#140.

AndyAyersMS:
Hmm, I wonder who wrote that test.

LGTM.

JosephTremoulet:
Updated to also exclude mismatch64, which unsurprisingly fails the same way.

PR #1029, opened 2016-02-19
Add exclusions for new failing tests.

richardlford:
When exception handling is enabled only 4 of the new tests
fail. However these also are exception handling tests, so
I've given all of the tests the exception handling issue number.


richardlford:
@pgavlin, please take a look.

richardlford:
I'm going to go ahead and merge.

PR #1030, opened 2016-02-22
Fix build on NetBSD: Solve name clash of TargetArch

krytarowski:
Address problem triggered on NetBSD-7.99.26 with GCC 4.8.2

```
make[1]: stopped in /tmp/pkgsrc-tmp/wip/llilc-git/work/build
/tmp/pkgsrc-tmp/wip/llilc-git/work/llilc/lib/CoreDisTools/coredistools.cpp:58:19:
error: declaration of 'TargetArch CorDisasm::TargetArch' [-fpermissive]
   enum TargetArch TargetArch;
                   ^
In file included from /tmp/pkgsrc-tmp/wip/llilc-git/work/llilc/lib/CoreDisTools/coredistools.cpp:35:0:
/tmp/pkgsrc-tmp/wip/llilc-git/work/llilc/include/CoreDisTools/coredistools.h:36:6:
error: changes meaning of 'TargetArch' from 'enum TargetArch' [-fpermissive]
 enum TargetArch {
      ^
```

PR #1031, opened 2016-02-22
CoreDisTools: Implement Near-Differ

swaroop-sridhar:
This change implements the near-diff functionality necessary for
asm-diff analysis. The near-differ compares two code-sections
for equivalency using the LLVM disassembler .

The two code blocks are disassembled in parallel looking for similar
instructions. In the case of constant operands, some heuristics are
applied via the areOffsetsEquivalent() callback to determine if the
constants can be considered equivalent (ex: constant offsets from the
start of the respective code sections).

### DIFF
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -37,25 +37,101 @@
 using namespace llvm;
 using namespace std;
 
+class BlockIterator;
+
+// Represents a Code block
+class BlockInfo {
+public:
+  BlockInfo(const uint8_t *Pointer, uint64_t Size, uintptr_t Address,
+            const char *BlockName = "")
+      : Ptr(Pointer), BlockSize(Size), Addr(Address), Name(BlockName) {}
+
+  void printHeader() const;
+  void dump(const CorDisasm *Disasm) const;
+  // Determine if the current Block is equivalent to the input `Block`
+  // according to the settings in DiffControl.
+  bool nearDiffCodeBlock(const BlockInfo &Block,
+                         const DiffControl &Control) const;
+  bool isEmpty() const { return BlockSize == 0; }
+
+  // A pointer to the code block to diassemble.
+  const uint8_t *Ptr;
+  // The size of the code block to compare.
+  uint64_t BlockSize;
+  // The original base address of the code block.
+  uintptr_t Addr;
+  // An identifying string, debug output only
+  const char *Name;
+
+private:
+  bool nearDiff(const BlockInfo &Block, const DiffControl &Control) const;
+  static bool diffError(const DiffControl &Control, const char *Mesg,
+                        const BlockIterator &Left, const BlockIterator &Right);
+};
+
+// A block iterator reoresents a code-point within a code block.
+// It represents an instruction within the block, and can move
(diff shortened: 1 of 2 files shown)


swaroop-sridhar:
@AndyAyersMS can you please review this change? thanks.

lib/CoreDisTools/coredistools.cpp
```diff
@@ -37,25 +37,98 @@
...
+  void decodeInstruction(BlockIterator &BIter) const;
+  size_t disasmInstruction(BlockIterator &BIter, bool PrintAsm = false) const;
+  void printInstruction(const BlockIterator &BIter) const;
 
 private:
   bool setTarget();
   bool verifyPrefixDecoding();
 
-  TargetArch TargetArch;
+  enum TargetArch TargetArch;
```
AndyAyersMS: See #1030, where reusing the name `TargetArch` like this causes trouble

lib/CoreDisTools/coredistools.cpp
```diff
@@ -317,20 +390,212 @@ size_t CorDisasm::disasmInstruction(size_t Address, const uint8_t *Bytes,
...
-void CorDisasm::printInstruction(const MCInst *MI, size_t Address,
-                                 size_t InstSize,
-                                 ArrayRef<uint8_t> Bytes) const {
-  outs() << format("%8" PRIx64 ":", Address);
-  outs() << "\t";
-  dumpBytes(Bytes.slice(0, InstSize), outs());
+void CorDisasm::printInstruction(const BlockIterator &BIter) const {
+  assert(BIter.isDecoded() && "Cannot print before Decode");
+
+  outs() << format("%8" PRIx64 ":", BIter.Addr);
```
AndyAyersMS: Is `PRIx64` portable? I thought the portable equivalent was `LLd` or something similar.

lib/CoreDisTools/coredistools.cpp
```diff
@@ -317,20 +390,212 @@ size_t CorDisasm::disasmInstruction(size_t Address, const uint8_t *Bytes,
...
+
+          int64_t ImmL = OperandL.getImm();
+          int64_t ImmR = OperandR.getImm();
+
+          if (ImmL == ImmR) {
+            continue;
+          }
+
+          if (Control.areEquivalent(Control.userData, ImmL, ImmR)) {
+            // The client somehow things that these offsets are equivalent
```
AndyAyersMS: Typo `things`

AndyAyersMS:
Overall this looks good, just a few small nits.

lib/CoreDisTools/coredistools.cpp
swaroop-sridhar: OK, I'll rename the field as TheTargetArch -- so that it'll merge with #1030.
The rename would probably merge correctly even without this change.

lib/CoreDisTools/coredistools.cpp
swaroop-sridhar: I used PRIx64 because llvm-obump also uses it:
https://github.com/Microsoft/llvm/blob/master/tools/llvm-objdump/llvm-objdump.cpp#L317 

I'll rename it to "llx"

PR #1036, opened 2016-03-03
[objwriter] Allow emitting arbitrary relocs in Windows unwind blob

jkotas:


jkotas:
@kyulee1 PTLA

It goes together with https://github.com/dotnet/corert/pull/886

lib/ObjWriter/objwriter.cpp
```diff
@@ -387,14 +389,19 @@ GetSymbolRefExpr(ObjectWriter *OW, const char *SymbolName,
 }
 
 extern "C" void EmitSymbolRef(ObjectWriter *OW, const char *SymbolName,
-                              int Size, bool IsPCRelative, int Delta = 0) {
+                              int Size, bool IsPCRelative, int Delta,
+                              int Kind) {
```
russellhadley: The only question I've got is around whether it's a good idea to use the opaque kind and cast to the VariantKind.  It's not quite leaking LLVM types out of the ObjWriter but it's close since you need to keep sources in sync.  Is it worth adding separate entry points for the different kinds with appropriate names?   That has it's own downside in possible number of entrypoints but it would mean that LLVM changes to the relocs only need to fixed on one side.

jkotas: I do not think one entrypoint per reloc kind is a good idea. The reloc kind should be enum.

We have a stable reloc kind enum on the JIT-EE interface (it is stable because of it matches COFF reloc kind values where possible). I am going to use that instead.

jkotas: Made the change.

lib/ObjWriter/objwriter.cpp
```diff
@@ -386,30 +388,53 @@ GetSymbolRefExpr(ObjectWriter *OW, const char *SymbolName,
...
+  int Size = 0;
+  MCSymbolRefExpr::VariantKind Kind = MCSymbolRefExpr::VK_None;
 
-  switch (Size) {
-  case 8:
-    assert(!IsPCRelative && "NYI no support for 8 byte pc-relative");
+  // Convert RelocType to MCSymbolRefExpr
+  switch (RelocType) {
+  case RelocType::IMAGE_REL_BASED_ABSOLUTE:
+    Kind = MCSymbolRefExpr::VK_COFF_IMGREL32;
```
kyulee1: Once we start using this then it implies Windows/COFF only. Can you add an assertion like this?:
```
assert(OW->MOFI->getObjectFileType() == OW->MOFI->IsCOFF);
```
I guess we will minimize such use for Window specific layout/representation.

kyulee1:
Can you also update all the files under
https://github.com/dotnet/llilc/tree/master/lib/ObjWriter/.nuget
from 1.0.7 to 1.0.8?

lib/ObjWriter/objwriter.cpp
jkotas: Fixed

jkotas:
Updated the version to 1.0.8

kyulee1:
LGTM

russellhadley:
LGTM too

jkotas:
@dotnet-bot test this please

PR #1038, opened 2016-03-11
CoreDisTools: Use Custom print API

swaroop-sridhar:
This commit has the following changes:

1) Separate out Asm-diff functionality from BlockInfo to a separate class
   CorAsmDiff, which inherits from CorDisasm, to avoid passing around
   some disassebler information.

2) Use customer provided printing routines via PrintControl interface
   instead of printing ot stdout/stderr

3) Use Intel syntax for X86/X64 asm instead of the default GCC syntax

### DIFF
--- a/include/CoreDisTools/coredistools.h
+++ b/include/CoreDisTools/coredistools.h
@@ -1,5 +1,4 @@
-//===--------- coredistools.h - Dissassembly tools for CoreClr
-//-------------===//
+//===--------- coredistools.h - Dissassembly tools for CoreClr ------------===//
 //
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license.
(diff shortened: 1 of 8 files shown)


swaroop-sridhar:
@AndyAyersMS  please review

swaroop-sridhar:
CC: @BruceForstall

include/CoreDisTools/coredistools.h
```diff
@@ -42,8 +42,37 @@ enum TargetArch {
...
+  const Printer Error;
+  const Printer Warning;
+  const Printer Log;
+  const Printer Dump;
+};
+
+// The type of a custom function provided by the user to determine
+// if two offsets are considered equivalent wrt diffing code blocks.
+// Offset1 and OFfset2 are thw two offsets to be compared
+// BlockOffset is the offest of the current instructions being
```
AndyAyersMS: A few typos in here: `OFfset2` and `offest`

AndyAyersMS: And `thw`

AndyAyersMS:
LGTM other than the small issues noted above.

swaroop-sridhar:
@AndyAyersMS I also changed the version number for the package, since the API has changed to include new functionality.

AndyAyersMS:
Still LGTM....

PR #1043, opened 2016-03-17
Support Comdat sections on Windows

nattress:
Add a parameter to CreateCustomSection specifying a Comdat symbol to use
for the section being created. The selection policy is left as SelectAny
to allow the linker to pick.

### DIFF
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -270,6 +270,8 @@ extern "C" bool CreateCustomSection(ObjectWriter *OW, const char *SectionName,
   std::string SectionNameStr(SectionName);
   assert(OW->CustomSections.find(SectionNameStr) == OW->CustomSections.end() &&
          "Section with duplicate name already exists");
+  assert(ComdatName == nullptr ||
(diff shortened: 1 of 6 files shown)


nattress:
@kyulee1 Please take a look

lib/ObjWriter/objwriter.cpp
```diff
@@ -259,7 +259,8 @@ enum CustomSectionAttributes : int32_t {
 };
 
 extern "C" bool CreateCustomSection(ObjectWriter *OW, const char *SectionName,
-                                    CustomSectionAttributes attributes) {
+                                    CustomSectionAttributes attributes,
+                                    const char *ComdatName) {
```
kyulee1: ComdateName is for Windows/Coff only.
Can you add an assert at minimum if ComdatName is not null like?:
```
assert(OW->MOFI->getObjectFileType() == OW->MOFI->IsCOFF);
```

nattress: Done

kyulee1:
LGTM. Test failures appear from lab script bugs.

PR #1045, opened 2016-03-18
CoreDisTools: Add Newline to Default Printer

swaroop-sridhar:
CoreDistools uses the customer provided print controls
for printing the log/error messages.
Unfortunately, SuperPMI appends a newline to all messages
passed into its print control. Therefore, CoreDisTools'
messages don't have the trailing newline.

This change adds the newline to the default printer, so that
other customers (ex: GCStress debugging) see the newline
when dumping a code block.

Also, adjusts the code block pretty printer's padding of
code bytes -- three characters are printed for each byte.

Swaroop.

### DIFF
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -112,17 +112,29 @@ public:
 };
 
 // Default Print Controls
+//
+// The default controls simply print to stdout and stderr.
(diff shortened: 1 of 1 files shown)


swaroop-sridhar:
@AndyAyersMS please review

AndyAyersMS:
LGTM

PR #1050, opened 2016-04-08
Build ObjWriter/CoreDisTools Only

kyulee1:
To not track unnecessary build break, build objwriter/coredistools only.


kyulee1:
@swaroop-sridhar PTAL

swaroop-sridhar:
Green.

PR #1051, opened 2016-04-20
COMDAT sections for ELF

nattress:
Add support for COMDAT sections for ELF targets. When `ComdatName` is
non-null in CreateCustomSection, create a group whose symbol is
`ComdatName` and associate the created group with it. Add
`ELF::SHF_GROUP` flag to the created section so it will be correctly
coalesced with sections from the same group name when linked to other
object files.

Fix a couple small issues found with my Ubuntu VM's stricter compiler:
disambigurate a type and parameter name and add `const` where it was
being stripped unnecessarily.

### DIFF
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -272,7 +272,8 @@ extern "C" bool CreateCustomSection(ObjectWriter *OW, const char *SectionName,
   assert(OW->CustomSections.find(SectionNameStr) == OW->CustomSections.end() &&
          "Section with duplicate name already exists");
   assert(ComdatName == nullptr ||
-         OW->MOFI->getObjectFileType() == OW->MOFI->IsCOFF);
+         OW->MOFI->getObjectFileType() == OW->MOFI->IsCOFF ||
+         OW->MOFI->getObjectFileType() == OW->MOFI->IsELF);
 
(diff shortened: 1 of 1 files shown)


nattress:
@kyulee1 Please take a look

lib/ObjWriter/objwriter.cpp
```diff
@@ -315,12 +316,20 @@ extern "C" bool CreateCustomSection(ObjectWriter *OW, const char *SectionName,
...
+    }
+
     if (attributes & CustomSectionAttributes_Executable) {
       Flags |= ELF::SHF_EXECINSTR;
     } else if (attributes & CustomSectionAttributes_Writeable) {
       Flags |= ELF::SHF_WRITE;
     }
-    Section = OutContext.getELFSection(SectionName, ELF::SHT_PROGBITS, Flags);
+    Section = OutContext.getELFSection(SectionName, ELF::SHT_PROGBITS, Flags,
+        0, ComdatName != nullptr ? ComdatName : "");
```
kyulee1: Can you merge line 319-325 down to here like the case of COFF? Basically, 
```
if (ComdatName != nullptr) {
  // new comdat case
} else {
  // original logic
}
```

kyulee1:
Can you also update version number?

PR #1055, opened 2016-04-26
CoreDisTools: Remove Prefix Decode Verification.

swaroop-sridhar:
Remove VerifyPrefixDeocde() in the Core-Disassembler, because:
1) It doesn't serve any purpose -- CorDisasm::disasmInstruction()
   uses machine opcodes (and not LLVM opcodes) to decide when to
   continue disassembly past a prefix instruction
2) LLVM decodes different sets of prefixes as independent
   instructions on X86 and AMD64 -- so this leads to unnecessary
   overhead in maintaining the X86Prefix::IsLLVMInstruction state.

### DIFF
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -169,7 +169,6 @@ protected:
 
 private:
   bool setTarget();
-  bool verifyPrefixDecoding();
 
   string TargetTriple;
   const Target *TheTarget;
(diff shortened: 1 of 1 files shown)


swaroop-sridhar:
@AndyAyersMS please review

swaroop-sridhar:
CC: @LLITCHEV

AndyAyersMS:
LGTM

PR #1056, opened 2016-05-12
Ending Section to Ensure Section Symbols

kyulee1:
- Record sections that is used and endSection to ensure Section symbol creations.
- Register all executable sections to Dwarf sections so that debug info is created for them.

### DIFF
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -307,7 +307,8 @@ bool CorDisasm::init() {
 
   MRI.reset(TheTarget->createMCRegInfo(TargetTriple));
   if (!MRI) {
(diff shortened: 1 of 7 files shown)


kyulee1:
@janvorli @AndyAyersMS PTAL

kyulee1:
@mmitche It appears all lab tests are failing. Any idea?

janvorli:
LGTM, I've verified it fixes the problem with multiple executable sections on Linux that I was hitting in CoreRT.

AndyAyersMS:
LGTM too. 

Are we using clang-format on this code? Just curious about the reformatting....

kyulee1:
I used "ccformat --untidy". I think the format change is because the previous checkin didn't run it.

PR #1058, opened 2016-05-13
Add EmitCFILsda to ExportList

kyulee1:
I just realized we've missed exporting this.


kyulee1:
@janvorli PTAL

janvorli:
Ah, thanks! It works without the export on Linux, so that's why I've missed it.
LGTM

PR #1066, opened 2016-10-15
Fix possible unaligned pointer dereference at/lib/Jit/LLILCJit.cpp

j123123:
see https://spin.atomicobject.com/2014/05/19/c-undefined-behaviors/ 2. Upcasting Pointers

### DIFF
--- a/lib/Jit/LLILCJit.cpp
+++ b/lib/Jit/LLILCJit.cpp
@@ -774,14 +774,14 @@ uint64_t ObjectLoadListener::getRelocationAddend(uint64_t LLVMRelocationType,
   switch (LLVMRelocationType) {
   case IMAGE_REL_AMD64_ABSOLUTE:
     memcpy(&tmp, FixupAddress, sizeof(uint32_t));
-    Append = tmp;
+    Addend = tmp;
     break;
   case IMAGE_REL_AMD64_ADDR64:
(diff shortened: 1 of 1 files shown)


lib/Jit/LLILCJit.cpp
```diff
@@ -770,15 +770,18 @@ void ObjectLoadListener::recordRelocations(
...
+    Append = tmp;
     break;
   case IMAGE_REL_AMD64_ADDR64:
-    Addend = *(uint64_t *)FixupAddress;
+    memcpy(&Addend, FixupAddress, sizeof(uint64_t));
     break;
   case IMAGE_REL_AMD64_REL32:
-    Addend = *(uint32_t *)FixupAddress;
+    memcpy(&tmp, FixupAddress, sizeof(uint32_t));
+    Append = tmp;
```
ForNeVeR: Shouldn't it be `Addend` instead of `Append`?

russellhadley:
These look fine but I'm curious what prompted the change?  Where are you seeing errors from this?

j123123:
@russellhadley, I just studied the source code. Nothing more. This bug can occur if it compiled for ARM CPU. Like this: https://stackoverflow.com/questions/14032434/casting-pointers-on-embedded-devices

PR #1069, opened 2017-02-10
Sharedunwindinfo

nattress:
When a method is marked Comdat foldable (ie, generic methods, compiler
generated methods), ensure the related unwind info goes into a comdat
section associated with the same method. The linker will then atomically
take a method and its unwind info from the same object file. Previously
all unwind info was emitted into a single section and would cause
"Invalid .pdata contribution" errors at link time with comdat methods.


nattress:
@sandreenko @AndyAyersMS PTAL. Ignore the co-authored commits; they're needed here and part of a separate PR.

sandreenko:
LGTM

AndyAyersMS:
LGTM too. You will (eventually?) need to do the same thing for the local debug info.

PR #1070, opened 2017-02-14
Extend SwitchSection API 

nattress:
Provide more precise control over custom sections. Previously we would
create a custom section for Comdat data and refer to it by name when
switching to it. However we really want custom sections with the same
name as existing sections, differing only on attributes or comdat name.
To accomplish this, remove the name -> MCSection mapping and provide an
extended API for switching sections that allows Comdat / attributes to
be specified.

### DIFF
--- a/lib/ObjWriter/objwriter.exports
+++ b/lib/ObjWriter/objwriter.exports
@@ -16,4 +16,3 @@ EmitDebugLoc
 EmitDebugFunctionInfo
 EmitDebugModuleInfo
 EmitDebugVar
-CreateCustomSection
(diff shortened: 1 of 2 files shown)


nattress:
@sandreenko PTAL. This is the objectwriter piece of a related commit in CoreRT. See https://github.com/nattress/corert/commit/8590ef5f9bbea9b57a48f5e031c960e10a15477e.

lib/ObjWriter/objwriter.exports
```diff
@@ -1,6 +1,7 @@
 InitObjWriter
 FinishObjWriter
 SwitchSection
+SwitchSectionEx
```
sandreenko: Is it possible to change SwitchSection signature and don't add new method?

nattress: Yeah I can do that. When I wrote this last week I rationalized that we call SwitchSection often so shouldn't eat the extra interop overhead. But marshalling 0 and null is cheap so I'll just collapse them into one API.

nattress:
@sandreenko Updated. Please merge if this looks good. I will refresh the nuget packages as of these commits.

PR #1071, opened 2017-02-21
ObjectWriter refactoring

sandreenko:
export functions must only call ObjectWriter public methods.
Header file was creater.
Console options were deleted as unused.
Unclear code around TripleName was fixed.
It fixes warnings with OSX version.

### DIFF
--- a/lib/ObjWriter/CMakeLists.txt
+++ b/lib/ObjWriter/CMakeLists.txt
@@ -2,6 +2,7 @@ project(objwriter)
 
 include_directories(${LLVM_INCLUDE_DIRS})
 include_directories(${CORECLR_INCLUDE})
+include_directories(.)
 add_definitions(${LLVM_DEFINITIONS})
 
 if (WIN32)
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -13,7 +13,7 @@
 ///
 //===----------------------------------------------------------------------===//
 
-#include "llvm/CodeGen/AsmPrinter.h"
+#include "objwriter.h"
 #include "llvm/DebugInfo/CodeView/CodeView.h"
 #include "llvm/DebugInfo/CodeView/Line.h"
 #include "llvm/DebugInfo/CodeView/SymbolRecord.h"
@@ -23,20 +23,20 @@
 #include "llvm/MC/MCDwarf.h"
 #include "llvm/MC/MCInstPrinter.h"
 #include "llvm/MC/MCInstrInfo.h"
-#include "llvm/MC/MCObjectFileInfo.h"
 #include "llvm/MC/MCParser/AsmLexer.h"
+#include "llvm/MC/MCParser/MCTargetAsmParser.h"
 #include "llvm/MC/MCRegisterInfo.h"
 #include "llvm/MC/MCSectionCOFF.h"
 #include "llvm/MC/MCSectionELF.h"
 #include "llvm/MC/MCSectionMachO.h"
 #include "llvm/MC/MCStreamer.h"
 #include "llvm/MC/MCSubtargetInfo.h"
-#include "llvm/MC/MCParser/MCTargetAsmParser.h"
 #include "llvm/MC/MCTargetOptionsCommandFlags.h"
 #include "llvm/MC/MCWinCOFFStreamer.h"
 #include "llvm/Support/COFF.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Compression.h"
+#include "llvm/Support/ELF.h"
 #include "llvm/Support/FileUtilities.h"
 #include "llvm/Support/FormattedStream.h"
 #include "llvm/Support/Host.h"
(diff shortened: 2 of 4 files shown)


lib/ObjWriter/objwriter.cpp
```diff
@@ -13,838 +13,715 @@
 ///
 //===----------------------------------------------------------------------===//
 
-#include "llvm/CodeGen/AsmPrinter.h"
```
nattress: Generally it's better to include files where they're needed. I would imagine most of these are needed in the implementation c++ file, not the header.

lib/ObjWriter/objwriter.cpp
```diff
@@ -13,838 +13,715 @@
...
-  return true;
+Triple ObjectWriter::GetTriple()
+{
+	Triple TheTriple(TripleName);
+
+	if (TheTriple.getOS() == Triple::OSType::Darwin)
+	{
+		TheTriple = Triple(TheTriple.getArchName(), TheTriple.getVendorName(), "darwin",
+			TheTriple.getEnvironmentName()); // it is workaround for llvm bug
+											 // https://bugs.llvm.org//show_bug.cgi?id=24927.
```
nattress: Spacing is weird - I notice you switched to tabs. The code-gen team was following the [LLVM coding guidelines](http://llvm.org/docs/CodingStandards.html) for LLILC (including running clang-format and clang-tidy during CI) which is why this file was formatted weirdly. I would preserve the existing style personally, but that's up to your team.

lib/ObjWriter/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ endif()
 add_library(objwriter
    SHARED
    objwriter.cpp
+   objwriter.h
```
nattress: You shouldn't need to specify headers explicitly. The compiler will pick them up from its list of include directories. If that's not working, you can tell cmake about it with `include_directories` directive.

lib/ObjWriter/objwriter.cpp
```diff
@@ -13,838 +13,715 @@
...
+	llvm_shutdown_obj Y; // Call llvm_shutdown() on exit.
 
-// When object writer is created/initialized successfully, it is returned.
-// Or null object is returned. Client should check this.
-extern "C" ObjectWriter *InitObjWriter(const char *ObjectFilePath) {
-  ObjectWriter *OW = new ObjectWriter();
-  if (OW->init(ObjectFilePath)) {
-    return OW;
-  }
+						 // Initialize targets
```
nattress: Spacing

lib/ObjWriter/objwriter.h
```diff
@@ -0,0 +1,261 @@
...
+	IMAGE_REL_BASED_ABSOLUTE = 0x00,
+	IMAGE_REL_BASED_HIGHLOW = 0x03,
+	IMAGE_REL_BASED_DIR64 = 0x0A,
+	IMAGE_REL_BASED_REL32 = 0x10,
+};
+
+class ObjectWriter
+{
+public:
+	bool init(StringRef FunctionName);
```
nattress: We should be consistent with naming here. Either camel / pascal case for all class members.

lib/ObjWriter/objwriter.h
```diff
@@ -0,0 +1,261 @@
...
+}
+
+extern "C" void FinishObjWriter(ObjectWriter* OW)
+{
+	assert(OW && "ObjWriter is null");
+	OW->finish();
+	delete OW;
+}
+
+extern "C" void SwitchSection(ObjectWriter*           OW,
```
nattress: Spacing

AndyAyersMS:
Yeah, you should run clang-format on this; see [here](https://github.com/dotnet/llilc/blob/master/Documentation/Code-Formatting) for instructions.

sandreenko:
I used clang-format with CoreCLR settings.
I will fix comments about headers/names and update PR.

AndyAyersMS:
So you're going to reformat with LLILC settings.....?

sandreenko:
As you wish. For me it is better to use same settings in CoreCLR and ObjectWriter, but there is no problem in running clang-format before each PR.

lib/ObjWriter/CMakeLists.txt
sandreenko: Compiler doesn't add header file without that.

sandreenko:
@nattress please look again. 
tidy shows warnings, but I am not sure that it was checked in LLILC, so I want to ignore them for now. Maybe we will have time to fix it in future.

lib/ObjWriter/CMakeLists.txt
sandreenko: Is it why ObjectWriter was implemented as one cpp file?

nattress: ObjectWriter is only used by CoreRT's managed code P/Invoking to exported functions; there's no C++ code that needs to include it in compilation so no header is strictly necessary.

PR #1073, opened 2017-02-26
Debug function symbols for non windows

sandreenko:
Fix problem with lldb debugging.

So after this fix we emit correct attributes for function symbols and sections.
The corresponding [CoreRT PR ](https://github.com/sandreenko/corert/commit/064cee1f14d7a5958e4e98285f57a41baf7338fa).

Some examples:
```
(lldb) breakpoint set --file Hello.cs --line 14
Breakpoint 1: where = Hello`Hello_Hello_Program__Main, address = 0x00000000005366a4
(lldb) r
Process 9580 launched: '/home/sergey/git/corert/tests/src/Simple/Hello/bin/Debug/x64/native/Hello' (x86_64)
Process 9580 stopped
* thread #1, name = 'Hello', stop reason = breakpoint 1.1
    frame #0: 0x00000000005366a4 Hello`Hello_Hello_Program__Main at Hello.cs:14
   11  	    internal class Program
   12  	    {
   13  	        private static void Main(string[] args)
-> 14  	        {
   15  	            if (args.Length == 0)
   16  	            {
   17  	                Console.WriteLine("Usage: hello name");


(gdb) break Hello.cs:15
Breakpoint 1 at 0x5366bc: file /home/sergey/git/corert/tests/src/Simple/Hello/Hello.cs, line 15.
(gdb) r
Starting program: /home/sergey/git/corert/tests/src/Simple/Hello/bin/Debug/x64/native/Hello 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffdcbdd700 (LWP 9651)]

Breakpoint 1, Hello_Hello_Program__Main () at /home/sergey/git/corert/tests/src/Simple/Hello/Hello.cs:15
15	            if (args.Length == 0)
(gdb) 


(lldb) breakpoint set --name Hello_Hello_Program__Main
Breakpoint 1: where = Hello`Hello_Hello_Program__Main + 24, address = 0x00000000005366bc

```

There is problem with image --lookup output, it doesn't show source line for address, but it will be fixed in the other PR.

Also I want to rename all variables in camel case. gcc doesn't allow to name variables as types (TargetMachine TargetMachine; ) and it creates many problems. So I want to use the same naming rules as CoreCLR and CoreRT.

### DIFF
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -244,7 +244,9 @@ void ObjectWriter::SwitchSection(const char *SectionName,
     }
   }
 
-  Sections.push_back(Section);
+  if (Sections.count(Section) == 0) {
+    Sections.insert(Section);
+  }
   OST.SwitchSection(Section);
 
   if (!Section->getBeginSymbol()) {
@@ -675,7 +677,7 @@ void ObjectWriter::EmitDebugFunctionInfo(const char *FunctionName,
       OST.EmitSymbolAttribute(Sym, MCSA_ELF_TypeFunction);
       OST.emitELFSize(Sym, MCConstantExpr::create(FunctionSize, OutContext));
     }
-    // TODO: Should test it for MachO.
+    // TODO: Should test it for Macho.
   }
 }
 
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -174,8 +174,10 @@ void ObjectWriter::SwitchSection(const char *SectionName,
   Streamer->SwitchSection(Section);
   if (Sections.count(Section) == 0) {
     Sections.insert(Section);
-    if (ObjFileInfo->getObjectFileType() != ObjFileInfo->IsCOFF) {
+    if (ObjFileInfo->getObjectFileType() == ObjFileInfo->IsMachO) {
       assert(!Section->getBeginSymbol());
+      // Output a DWARF linker-local symbol.
+      // This symbol is used as a base for other symbols in a section.
       MCSymbol *SectionStartSym = OutContext->createTempSymbol();
       Streamer->EmitLabel(SectionStartSym);
       Section->setBeginSymbol(SectionStartSym);
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -670,7 +670,12 @@ void ObjectWriter::EmitDebugFunctionInfo(const char *FunctionName,
     OST.EmitCVFuncIdDirective(FuncId);
     EmitCVDebugFunctionInfo(FunctionName, FunctionSize);
   } else {
-    // TODO: Should convert this for non-Windows.
+    if (MOFI->getObjectFileType() == MOFI->IsELF) {
+      MCSymbol *Sym = OutContext.getOrCreateSymbol(Twine(FunctionName));
+      OST.EmitSymbolAttribute(Sym, MCSA_ELF_TypeFunction);
+      OST.emitELFSize(Sym, MCConstantExpr::create(FunctionSize, OutContext));
+    }
+    // TODO: Should test it for MachO.
   }
 }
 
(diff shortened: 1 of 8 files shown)


sandreenko:
PTAL @nattress @AndyAyersMS

lib/ObjWriter/objwriter.h
```diff
@@ -66,48 +70,55 @@ class ObjectWriter {
...
-  std::unique_ptr<TargetMachine> TM;
-  std::unique_ptr<AsmPrinter> Asm;
+  std::unique_ptr<MCRegisterInfo> RegisterInfo;
+  std::unique_ptr<MCAsmInfo> AsmInfo;
+  std::unique_ptr<MCObjectFileInfo> ObjFileInfo;
+  std::unique_ptr<MCContext> OutContext;
+  MCAsmBackend *AsmBackend; // Owned by MCStreamer
+  std::unique_ptr<MCInstrInfo> InstrInfo;
+  std::unique_ptr<MCSubtargetInfo> SubtargetInfo;
+  MCCodeEmitter *CodeEmmiter; // Owned by MCStreamer
```
AndyAyersMS: Typo: name should be `CodeEmitter`

lib/ObjWriter/objwriter.cpp
```diff
@@ -103,207 +103,221 @@ bool ObjectWriter::Init(llvm::StringRef ObjectFilePath) {
...
-  const MCObjectFileInfo *MOFI = OutContext.getObjectFileInfo();
-  Triple TheTriple(TripleName);
+  MCSection *Section = GetSection(SectionName, attributes, ComdatName);
+  Streamer->SwitchSection(Section);
+  if (Sections.count(Section) == 0) {
+    Sections.insert(Section);
+    if (!Section->getBeginSymbol()) { // Is it set for default sections?
+      MCSymbol *SectionStartSym = OutContext->createTempSymbol();
+      Streamer->EmitLabel(SectionStartSym);
+      Section->setBeginSymbol(SectionStartSym);
```
AndyAyersMS: I don't really understand what this does. 

Do we ever refer to this temp symbol anywhere else? If so, you should leave a comment indicating where.

If not, why do we need it create it at all?

JosephTremoulet:
> Also I want to rename all variables in camel case. gcc doesn't allow to name variables as types (TargetMachine TargetMachine; ) and it creates many problems. So I want to use the same naming rules as CoreCLR and CoreRT.

FYI, a number of places in the codebase (and in LLVM) dodge this by prefixing with "The" (TargetMachine TheTargetMachine;).  I'm not concerned whether you do it that way or camlCase as you suggest, just wanted to mention it for consideration.

sandreenko:
> FYI, a number of places in the codebase (and in LLVM) dodge this by prefixing with "The" (TargetMachine TheTargetMachine;). I'm not concerned whether you do it that way or camlCase as you suggest, just wanted to mention it for consideration.

Yes, it is possible variant, but it kills all advantages of intellisense because all variables start with the same prefix. But I will be happy with any decision.

lib/ObjWriter/objwriter.cpp
sandreenko: Begin and end symbols are used in AsmPrinter. They have to be set in every section. They produce base for debug symbols offsets. I will clear this commit and change it to assert(!Section->getBeginSymbol()).

lib/ObjWriter/objwriter.h
sandreenko: fixed

lib/ObjWriter/objwriter.cpp

sandreenko: Update: I checked it again. We need this symbols for DWARF only. They are used as base symbols when we emit debug locations.

AndyAyersMS: So are you adding it back? Fixing it some other way?

sandreenko: We need this symbols and they are used in dwarf for EmitLabelPlusOffset, where label is our beginSymbol.

AndyAyersMS: Sounds like that would make a good code comment.

sandreenko: Done.

AndyAyersMS:
Latest version LGTM.

PR #1074, opened 2017-03-20
User Defined types for windows

sandreenko:
GetEnumTypeIndex, GetClassTypeIndex and CompleteClassDescription were added to export.
llvm::codeview::TypeTableBuilder is used to create UDT declarations.
Current PR implements only simple classes and structures. 
More complex cases (Polymorphism, interfaces, virtual calls, static methods descriptions) will be implemented in a separate PR.


sandreenko:
PTAL @AndyAyersMS, @nattress.

lib/ObjWriter/CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ endif()
 add_library(objwriter
    SHARED
    objwriter.cpp
+   objwriter.h
```
nattress: Please remove headers from the set of C++ files to compile; they should not be needed if the include paths are configured properly.

lib/ObjWriter/typeBuilder.cpp
```diff
@@ -0,0 +1,32 @@
+#include "typeBuilder.h"
```
nattress: Missing copyright header

lib/ObjWriter/typeBuilder.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
nattress: Missing copyright header

nattress:
In future, would you mind squashing before sending the first PR? It's hard to review changes across 4 commits

sandreenko:
> In future, would you mind squashing before sending the first PR? It's hard to review changes across 4 commits

If you click on "files changed", it will show you all changes.

lib/ObjWriter/objwriter.cpp
```diff
@@ -713,3 +719,27 @@ void ObjectWriter::EmitDebugModuleInfo() {
     OutContext->setGenDwarfForAssembly(true);
   }
 }
+
+unsigned ObjectWriter::GetEnumTypeIndex(EnumTypeDescriptor TypeDescriptor,
+                                        EnumRecordTypeDescriptor *TypeRecords) {
+  if (ObjFileInfo->getObjectFileType() == ObjFileInfo->IsCOFF)
+    return TypeBuilder.GetEnumTypeIndex(TypeDescriptor, TypeRecords);
+  else
+    return 0;
```
nattress: We should just not call this at all on Unix and save the p/invoke overhead (ie, this should be `assert(ObjFileInfo->getObjectFileType() == ObjFileInfo->IsCOFF);`

lib/ObjWriter/typeBuilder.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+unsigned UserDefinedTypesBuilder::GetEnumFieldListType(
+    uint64 Count, EnumRecordTypeDescriptor *TypeRecords) {
+  FieldListRecordBuilder FLRB(TypeTable);
+  FLRB.begin();
+  uint64 MaxInt = (unsigned int)-1;
+  if (Count > MaxInt) {
+    // Do not expect more than int fields.
+    return 0;
```
nattress: This looks like it should be surfaced as an error instead of silent return of 0, at least in debug mode.

lib/ObjWriter/CMakeLists.txt
sandreenko: It doesn't work. It looks like it is a common problem for VS, [stack overflow](http://stackoverflow.com/questions/1167154/listing-header-files-in-visual-studio-c-project-generated-by-cmake)

lib/ObjWriter/typeBuilder.h
sandreenko: Fixed

lib/ObjWriter/objwriter.cpp

lib/ObjWriter/typeBuilder.cpp
sandreenko: Fixed

lib/ObjWriter/typeBuilder.cpp
sandreenko: Fixed

sandreenko:
>  Is there a reason you chose to go the native implementation route? If we ever change our object file emitter implementation, we will now have to re-implement the CV emission.

The main reason is that we want to have better debugging at least on windows as soon as possible to improve developers performance. This variant is the fastest one.

lib/ObjWriter/CMakeLists.txt
AndyAyersMS: You should add a comment, this is the second time people have wondered why you are doing this.

nattress:
LGTM - One other thing, the CI in CoreRT has a lot of ObjWriter output in its log. See, ie:

```
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         emit abs value 65
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         MCELf emit value impl kind 0  MCObjectValue impl
         emit abs value 362
```
Can you make sure any logging is disabled when you build the Nuget packages (and ensure it's the release version of objwriter we upload, too).

PR #1076, opened 2017-05-12
Array type implementation

sandreenko:
Array UDT.
Non-MD array has BaseClass, Count and ArrayElementType fields.
MD arrays also have each dimension length and lower bound values.

### DIFF
--- a/lib/ObjWriter/typeBuilder.cpp
+++ b/lib/ObjWriter/typeBuilder.cpp
@@ -13,7 +13,8 @@
 #include <sstream>
 
 UserDefinedTypesBuilder::UserDefinedTypesBuilder()
-    : Allocator(), TypeTable(Allocator), Streamer(nullptr), TargetPointerSize(0) {}
+    : Allocator(), TypeTable(Allocator), Streamer(nullptr),
+      TargetPointerSize(0) {}
 
 void UserDefinedTypesBuilder::SetStreamer(MCObjectStreamer *Streamer) {
   assert(this->Streamer == nullptr);
@@ -163,24 +164,17 @@ unsigned UserDefinedTypesBuilder::GetArrayTypeIndex(
 
   if (ArrayDescriptor.IsMultiDimensional == 1) {
     for (unsigned i = 0; i < ArrayDescriptor.Rank; ++i) {
-      std::stringstream ss;
-      ss << "length" << i;
-      char *LengthName = new char[ss.gcount() + 1];
-      ss >> LengthName;
+
       DataMemberRecord LengthDMR(Access, TypeIndex(SimpleTypeKind::Int32),
-                                 Offset, LengthName);
+                                 Offset, ArrayDimentions.GetLengthName(i));
       FLBR.writeMemberType(LengthDMR);
       FieldsCount++;
       Offset += 4;
     }
 
     for (unsigned i = 0; i < ArrayDescriptor.Rank; ++i) {
-      std::stringstream ss;
-      ss << "bounds" << i;
-      char *BoundsName = new char[ss.gcount() + 1];
-      ss >> BoundsName;
       DataMemberRecord BoundsDMR(Access, TypeIndex(SimpleTypeKind::Int32),
-                                 Offset, BoundsName);
+                                 Offset, ArrayDimentions.GetBoundsName(i));
       FLBR.writeMemberType(BoundsDMR);
       FieldsCount++;
       Offset += 4;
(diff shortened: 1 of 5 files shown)


sandreenko:
PTAL @nattress @AndyAyersMS

lib/ObjWriter/typeBuilder.cpp
```diff
@@ -130,10 +132,91 @@ unsigned UserDefinedTypesBuilder::GetCompleteClassTypeIndex(
...
+  DataMemberRecord CountDMR(Access, IndexType, Offset, "count");
+  FLBR.writeMemberType(CountDMR);
+  FieldsCount++;
+  Offset += TargetPointerSize;
+
+  if (ArrayDescriptor.IsMultiDimensional == 1) {
+    for (unsigned i = 0; i < ArrayDescriptor.Rank; ++i) {
+      std::stringstream ss;
+      ss << "length" << i;
+      char *LengthName = new char[ss.gcount() + 1];
```
nattress: This looks like a memory leak. You can probably just use ss.str().c_str() here to pass a char * through to the debug record serializer. Most string passing is done in LLVM with `llvm::StringRef` which might be more appropriate here, too.

lib/ObjWriter/typeBuilder.cpp
```diff
@@ -130,10 +132,91 @@ unsigned UserDefinedTypesBuilder::GetCompleteClassTypeIndex(
...
+                                 Offset, LengthName);
+      FLBR.writeMemberType(LengthDMR);
+      FieldsCount++;
+      Offset += 4;
+    }
+
+    for (unsigned i = 0; i < ArrayDescriptor.Rank; ++i) {
+      std::stringstream ss;
+      ss << "bounds" << i;
+      char *BoundsName = new char[ss.gcount() + 1];
```
nattress: More char* leaking

lib/ObjWriter/typeBuilder.h
```diff
@@ -67,14 +75,22 @@ class UserDefinedTypesBuilder {
...
                                 EnumRecordTypeDescriptor *TypeRecords);
+  unsigned GetPointerType(TypeIndex ClassIndex);
+
+  void AddBaseClass(FieldListRecordBuilder &FLBR, unsigned BaseClassId);
 
   MCObjectStreamer *Streamer;
   BumpPtrAllocator Allocator;
   TypeTableBuilder TypeTable;
+
+  unsigned TargetPointerSize;
```
nattress: This should get initialized to 0 in the constructor to be safe.

lib/ObjWriter/typeBuilder.cpp
```diff
@@ -130,10 +132,91 @@ unsigned UserDefinedTypesBuilder::GetCompleteClassTypeIndex(
...
+      ss >> BoundsName;
+      DataMemberRecord BoundsDMR(Access, TypeIndex(SimpleTypeKind::Int32),
+                                 Offset, BoundsName);
+      FLBR.writeMemberType(BoundsDMR);
+      FieldsCount++;
+      Offset += 4;
+    }
+  }
+
+  ArrayRecord AR(ElementTypeIndex, IndexType, ArrayDescriptor.Size, "");
```
nattress: Should we set the Name field to something useful here? I'm not very familiar with CodeView - does this feed some friendly array type name in the debugger?

sandreenko: No, it is the name of arrayElement + arrayIndex descriptor. ProjectN inserts empty string there too.

lib/ObjWriter/typeBuilder.h
sandreenko: Fixed

lib/ObjWriter/typeBuilder.cpp
sandreenko: Fixed

lib/ObjWriter/typeBuilder.cpp
sandreenko: Fixed

sandreenko:
@nattress could you please look again?

lib/ObjWriter/typeBuilder.cpp
```diff
@@ -130,10 +135,111 @@ unsigned UserDefinedTypesBuilder::GetCompleteClassTypeIndex(
...
+}
+
+const char *ArrayDimentionsDescriptor::GetBoundsName(unsigned index) {
+  if (Bounds.size() <= index) {
+    Resize(index);
+  }
+  return Bounds[index].c_str();
+}
+
+void ArrayDimentionsDescriptor::Resize(unsigned NewSize) {
```
nattress: Typo: `ArrayDimentionsDescriptor` should be `ArrayDimensionsDescriptor`

sandreenko: Thanks for your review.

PR #1078, opened 2017-05-16
Use references for args inside ObjectWriter 

sandreenko:
+ version update.

PTAL @nattress, this PR is very small.

### DIFF
--- a/lib/ObjWriter/.nuget/toolchain.osx.10.10-x64.Microsoft.DotNet.ObjectWriter.nuspec
+++ b/lib/ObjWriter/.nuget/toolchain.osx.10.10-x64.Microsoft.DotNet.ObjectWriter.nuspec
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>toolchain.osx.10.10-x64.Microsoft.DotNet.ObjectWriter</id>
-    <version>1.0.17-prerelease-00002</version>
+    <version>1.0.18-prerelease-00001</version>
     <title>Microsoft .NET Object File Generator</title>
     <authors>Microsoft</authors>
     <owners>Microsoft</owners>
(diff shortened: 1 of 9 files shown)


JosephTremoulet:
Should they be `const` references?

lib/ObjWriter/.nuget/toolchain.osx.10.10-x64.Microsoft.DotNet.ObjectWriter.nuspec
```diff
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>toolchain.osx.10.10-x64.Microsoft.DotNet.ObjectWriter</id>
-    <version>1.0.17-prerelease-00002</version>
+    <version>1.0.17-prerelease-00003</version>
```
nattress: We added an export altering the public surface area. We should bump to `1.0.18-rerelease-00001`

sandreenko:
> Should they be const references?

ObjectWriter gets them as values:
```
extern "C" unsigned GetArrayTypeIndex(ObjectWriter *OW,
                                      ClassTypeDescriptor ClassDescriptor,
                                      ArrayTypeDescriptor ArrayDescriptor)
```
So we own this values and can change them as we want. 
What is a value to do that?

lib/ObjWriter/.nuget/toolchain.osx.10.10-x64.Microsoft.DotNet.ObjectWriter.nuspec
sandreenko: Fixed

JosephTremoulet:
> What is a value to do that?

Just the standard ones for why to prefer `const` by default if the only reason you're passing things by reference is to reduce call overhead:
 - Intent of code is more clear from signature
 - Easier to reason about what state changes may occur where
 - Callers can pass rvalues

sandreenko:
The const attributes were added.

PR #1079, opened 2017-05-22
emit S_UDT symbols to keep all UDT alive after the linker.

sandreenko:
We need s_udt to guarantee that all UDT will be alive after the linker.

@nattress PTAL.

### DIFF
--- a/lib/ObjWriter/objwriter.cpp
+++ b/lib/ObjWriter/objwriter.cpp
@@ -702,9 +702,36 @@ void ObjectWriter::EmitDebugLoc(int NativeOffset, int FileId, int LineNumber,
   }
 }
 
+void ObjectWriter::EmitCVUserDefinedTypesSymbols() {
+  const auto &UDTs = TypeBuilder.GetUDTs();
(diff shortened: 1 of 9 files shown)


lib/ObjWriter/objwriter.cpp
```diff
@@ -702,9 +702,33 @@ void ObjectWriter::EmitDebugLoc(int NativeOffset, int FileId, int LineNumber,
   }
 }
 
+void ObjectWriter::EmitCVUserDefinedTypesSymbols() {
```
nattress: Out of interest, what do we do if there are no UDTs? Should we skip this altogether and save the bytes?

nattress:
What is the practical effect of this? What do we currently lose by not having this fix? It would be nice to have that rationale called out in the check-in comments so future spelunkers of the code know what the intent behind the fix is.

sandreenko:
@nattress it fixes dotnet/corert#3538.

lib/ObjWriter/objwriter.cpp

PR #1088, opened 2018-06-16
R2RDump - Add DumpInstruction which prints single instruction

acmyu:


lib/CoreDisTools/coredistools.cpp
```diff
@@ -684,6 +686,15 @@ DllIface size_t DisasmInstruction(const CorDisasm *Disasm,
   return DecodeLength;
 }
 
+DllIface size_t DumpInstruction(const CorDisasm *Disasm,
+	const uint8_t *Address, const uint8_t *Bytes,
+	size_t Maxlength) {
+	assert((Disasm != nullptr) && "Disassembler object Expected ");
```
nattress: Does this ensure we only dump at addresses at the beginning of an instruction? Ie, we want to make sure we're not disassembling from the middle of some instruction.

acmyu: If it gets passed an address in the middle of the instruction, there's decode failure. The existing DumpCodeBlock function also has decode failure if the address is in the middle of an instruction

acmyu:
@sandreenko Hi, is the PR okay to be merged?

PR #1092, opened 2018-06-28
Add fallthrough comment to prevent error on Linux build

acmyu:


sandreenko:
I think you need to add break statements there and a `default` case that prints `error` and returns `false`.

PR #1093, opened 2018-08-14
Update Microsoft.NETCore.CoreDisTools

zamont:
Modify Microsoft.NETCore.CoreDisTools to reference the runtime packages via direct NuGet dependencies instead of runtime.json. Update the version to 1.0.1-prerelease-00005 to update the Windows packages


zamont:
This change is required for the r2rdump app in dotnet/coreclr to work and this should remain functional with the GC stress dependencies in dotnet/coreclr

zamont:
@RussKeldorph @swaroop-sridhar @nattress

nattress:
I would really prefer we stick to runtime dependencies where possible so we don't download more than we need. We chatted in the hall about this earlier but could you summarize what wasn't working with the runtime.json method (this presumably worked before)? Then we can double check with the Nuget / dotnet CLI experts and see what has changed.

zamont:
After some more experimenting I've finally figured out the correct dotnet CLI commands to make this work with the runtime.json mechanism which I could never get working before. I will revert the runtime.json changes but still keep the versioning changes as we need to publish a version 1.0.1-prerelease-00005 of the Windows packages.

zamont:
@RussKeldorph does this look good? If so, can you perform the merge for me?

PR #1098, opened 2019-05-31
[WIP] Support Arm32 and Arm64 in nearDiff

echesakov:
Right now CorDisTools can not be used to disassemble JIT output during SuperPMI with ARM64 JIT due to the fact that on ARM64 the code block is appended with literal pool (https://github.com/dotnet/coreclr/blob/06f1634c3157dc279779f3dad7d2f327416e3f09/src/jit/emit.cpp#L4629-L4650) 

What happens is that the disassembler goes beyond JIT code and tries to decode data in literal pool where it obviously fails.

I couldn't find any information collected during JIT-EE interactions that can help to determine the size of  JIT code region. It seems that we are loosing such information during allocMem call by explicitly passing 0 in roDataSize (https://github.com/dotnet/coreclr/blob/06f1634c3157dc279779f3dad7d2f327416e3f09/src/jit/emit.cpp#L4645)

What I am doing here is trying to reconstruct boundary between JIT code and constant pool during instruction decoding by looking at LDR instructions that have PCRel immediate as a second operand and computing the corresponding label addresses using current value of PC. The minimum address is then assumed to be the beginning of a literal pool. 

I did experimentation with running SuperPMI with ARM64 JIT with CorDisTools with these changes and it seems that this approach works on the all framework libraries in Core_Root. 

@dotnet/jit-contrib 
Does anyone think about a better way to figure out the literal pool size?
@briansull @BruceForstall

### DIFF
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -281,7 +281,7 @@ bool CorDisasm::setTarget() {
     break;
 
   case Target_Thumb:
-    TheTriple.setArch(Triple::thumb);
+    TheTriple.setArchName("thumbv7");
     break;
   case Target_Arm64:
     TheTriple.setArch(Triple::aarch64);
--- a/lib/CoreDisTools/coredistools.cpp
+++ b/lib/CoreDisTools/coredistools.cpp
@@ -575,6 +575,15 @@ bool CorAsmDiff::nearDiff(const BlockInfo &LeftBlock,
   const uint8_t* ConsBlockL = nullptr;
   const uint8_t* ConsBlockR = nullptr;
 
+  constexpr int ARM_t2MOVTi16 = 2918;
+  constexpr int ARM_t2MOVi16  = 2920;
+
+  bool IsPrecededByMov  = false;
+  unsigned int Mov32Reg = 0;
+
+  int32_t Mov32ImmL = 0;
+  int32_t Mov32ImmR = 0;
+
   while (!Left.isEmpty() && !Right.isEmpty()) {
     const bool isAtLiteralPoolL = (ConsBlockL != nullptr) && (Left.Ptr == ConsBlockL);
     const bool isAtLiteralPoolR = (ConsBlockR != nullptr) && (Right.Ptr == ConsBlockR);
@@ -662,10 +671,22 @@ bool CorAsmDiff::nearDiff(const BlockInfo &LeftBlock,
             continue;
           }
 
-          if (Comparator(UserData, Left.BlockOffset(), Left.InstrSize, ImmL,
-                         ImmR)) {
+          if (Comparator(UserData, Left.BlockOffset(), Left.InstrSize, ImmL, ImmR)) {
             // The client somehow thinks that these offsets are equivalent
             continue;
+          }
+
+          // MOVW, MOVT instruction pair can be used to load a 32-bit value to a register. See if this is a case.
+          if (TheTargetArch == Target_Thumb) {
+            if ((InstL.getOpcode() == ARM_t2MOVTi16) && IsPrecededByMov && (InstL.getOperand(0).getReg() == Mov32Reg)) {
+              // Reconstruct 32-bit values that are being loaded to a register using MOV32.
+              Mov32ImmL = (ImmL << 16) | Mov32ImmL;
+              Mov32ImmR = (ImmR << 16) | Mov32ImmR;
+
+              if (Comparator(UserData, Left.BlockOffset(), Left.InstrSize, Mov32ImmL, Mov32ImmR)) {
+                continue;
+              }
+            }
           }
 
           return fail("Immediate Operand Value Mismatch", Left, Right);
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
We could put some invalid instruction padding in between the end of the actual code and the literal pool, and have the disassembly tools know what that means.

RussKeldorph:
Other ways that change what the JIT does and wastes a small amount of space or time:
 - If you don't want to increase the distance between the code and the literals, could you write the size or offset of the literals at the end of the allocated block?
 - Add a dummy API or parameter to the JIT-EE interface that conveys this info?

Hard to think of something simpler than what you're doing that doesn't add extra meta-data, though.  If you can discover all the code entrypoints from the JIT-EE interactions, you could do flow analysis to find live code and treat everything else as data.  That seems insane, though, unless we started interleaving data with code, and even then I think you're method would work most of the time.

I apologize for my JIT ignorance, but do we generate switch tables as inline data?  If so, that data may need to be discovered as well.

lib/CoreDisTools/coredistools.cpp
```diff
@@ -412,6 +412,37 @@ bool CorDisasm::decodeInstruction(BlockIterator &BIter, bool MayFail) const {
   return IsDecoded;
 }
 
+static bool tryDecodePCRelLabel(MCInst const& Inst, int64_t& Disp) {
+  constexpr unsigned int AArch64_LDRDl = 2359;
+  constexpr unsigned int AArch64_LDRSl = 2407;
+
+  bool HasPCRelLabel      = false;
+  unsigned int OperandNum = -1;
```
sandreenko: Who do you need this var?

lib/CoreDisTools/coredistools.cpp
```diff
@@ -412,6 +412,37 @@ bool CorDisasm::decodeInstruction(BlockIterator &BIter, bool MayFail) const {
...
+
+  switch (Inst.getOpcode()) {
+    case AArch64_LDRDl:
+    case AArch64_LDRSl:
+      HasPCRelLabel = true;
+	  OperandNum    = 1;
+	  break;
+  }
+
+  if (HasPCRelLabel) {
```
sandreenko: Can't it be replaced with `Inst.getOpcode() == AArch64_LDRDl || Inst.getOpcode() == AArch64_LDRSl`?

lib/CoreDisTools/coredistools.cpp
```diff
@@ -412,6 +412,37 @@ bool CorDisasm::decodeInstruction(BlockIterator &BIter, bool MayFail) const {
...
+  constexpr unsigned int AArch64_LDRDl = 2359;
+  constexpr unsigned int AArch64_LDRSl = 2407;
+
+  bool HasPCRelLabel      = false;
+  unsigned int OperandNum = -1;
+
+  switch (Inst.getOpcode()) {
+    case AArch64_LDRDl:
+    case AArch64_LDRSl:
+      HasPCRelLabel = true;
```
sandreenko: the formatting is broken here.

lib/CoreDisTools/coredistools.cpp
```diff
@@ -412,6 +412,37 @@ bool CorDisasm::decodeInstruction(BlockIterator &BIter, bool MayFail) const {
...
+  if (HasPCRelLabel) {
+    assert(Inst.getOperand(OperandNum).isImm() && "Operand must be immediate value");
+    int64_t ImmVal = Inst.getOperand(OperandNum).getImm();
+
+    // Adapted from AArch64Disassembler.cpp DecodePCRelLabel19
+    // Sign-extend 19-bit immediate.
+    if (ImmVal & (1 << (19 - 1)))
+      ImmVal |= ~((1LL << 19) - 1);
+
+    Disp = ImmVal * 4;
```
sandreenko: Is not it clear to return Disp as result, return 0 if there is no RelOffset?

lib/CoreDisTools/coredistools.cpp
```diff
@@ -541,7 +572,25 @@ bool CorAsmDiff::nearDiff(const BlockInfo &LeftBlock,
     return false;
   }
 
+  const uint8_t* ConsBlockL = nullptr;
```
sandreenko: Could you please add a comment or longer names for these two variables?

lib/CoreDisTools/coredistools.cpp
sandreenko: Can it be negative?

BruceForstall:
It seems like this is a bit of a hack to me.

On the one hand, if we want to support disassembling blobs of code that might also contain data, we should do something more sophisticated like construct a flow graph, and binary compare any bits between the "found" native code.

For this case, it feels like we should extend the "allocMem" JIT-EE interface method to take a "roCloseDataSize" that would be added by the VM to the code size, the way our call to `allocMem` does currently. (Actually, it seems like we should create `roCloseHotDataSize` and `roCloseColdDataSize`.) With this, SuperPMI would get the right info.

echesakovMSFT:
Closing in favor of https://github.com/dotnet/jitutils/pull/291

